PR #1, opened 2019-02-27
adds ambassador and argo components to manifests

swiftdiaries:
/cc @kkasravi @jlewi 

I'm starting with ambassador and argo because I already have the converted components. I'll add the remaining components pretty soon.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/1)


swiftdiaries:
Issue for reference/paper trail: https://github.com/kubeflow/kubeflow/issues/2566

argo/overlays/docker-for-desktop/kustomization.yaml
```diff
@@ -0,0 +1,2 @@
+bases:
+- ../../base
```
jlewi: Do we need to define an overlay if the overlay isn't doing any customization?

kustomization.yaml
```diff
@@ -0,0 +1,5 @@
+bases:
+  - ambassador/base
```
jlewi: Shouldn't there be a kustomization file for each package so that each application can be installed independently?

argo/overlays/docker-for-desktop/kustomization.yaml
swiftdiaries: I thought I'd get an initial commit in and add the overlays specific to each platform later

jlewi:
How did you generate these manifests? Did you just do ks show?

kustomization.yaml
swiftdiaries: Yes, there is one for each component.
This is to combine everything and deploy the set of components. 
I've explained the reasoning here: https://github.com/kubeflow/manifests/issues/4

swiftdiaries:
Yes, ks show

argo/overlays/docker-for-desktop/kustomization.yaml
jlewi: I wouldn't expect most packages to require platform specific customization.
So I would only expect us to define overlays for those packages and platforms requiring customization.

But this PR seems to be implying that we will define overlays for every platform for every package because you are creating stubs to be filled out later.

kustomization.yaml
jlewi: For each platform we will have different packages that we want to install; e.g. there are GCP specific packages. 
We also have different configurations for each platform;
e.g.
GCP + IAP
GCP + Basic Auth
Minikube

For each of the above there is a different set of applications to install and possibly different overlays to be used for each application.

Is the intent for the root kustomization.yaml to support all of that? Or is the root kustomization.yaml only going to provide a vanilla K8s cconfig.

swiftdiaries: Yeah, the root customization can choose what components to install from the different components. 
So for minikube/gke, we'd have an overlay based on the root customization.

argo/overlays/docker-for-desktop/kustomization.yaml
swiftdiaries: Oh yeah, that makes sense. I'll remove stubs for components that don't require customization.

kustomization.yaml
jlewi: Cool.

jlewi:
I think this is good to go once you remove the empty overlays.

swiftdiaries:
Oh sorry I didn't see your comment, will update the PR. Thanks for the review !

jlewi:
/lgtm
/approve

This is great thanks for the quick work.

PR #5, opened 2019-02-27
[wip] Adds tf-job and pytorch operators

swiftdiaries:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/5)


jlewi:
@swiftdiaries Shall we pick one package/PR and do that one first before doing other packages?

I have the same questions I had about your other PR? e.g. Do we need platform specific overlays if there is no platform specific customization?

swiftdiaries:
Yeah, we'll nail down the pattern for the ambassador, argo PR. Then modify the rest of the components accordingly.

swiftdiaries:
I assumed that every component would need to be customized. Not necessarily platform specific but some kind of customization. The stubs were meant for that. We could replace them with probable or common customizations that specific components need. For example, pytorch-job or TFJob would need the training image to be customized...

jlewi:
Marking this as WIP since its blocked on the other PRs being submitted first.

Per discussion in the other PR. I think we should avoid defining overlays until we have actual customization  to add. 

The reason is people will undoubtedly add new platforms to kfctl without adding new overlays to all existing packages. 

Its highly desirable if most packages don't require per platform customization as that reduces the maintenance burden significantly.

jlewi:
Do we need kustomize packages for different examples of TFJob and PyTorch jobs? 
It might be easier to just publish them as examples.

We can probably figure out what to do about example resources last since they aren't shared components.

That said if you want to include them I think its fine.

Issue #7, opened 2019-03-14
Profiles controller kustomize manifest depends on https://github.com/kubeflow/kubeflow/pull/2692


swiftdiaries:
Once this PR gets merged we should update the base in the kustomization at the root of the repo.

PR #8, opened 2019-03-15
Adds the Jupyter web app manifest

swiftdiaries:
Link to prototype: https://github.com/kubeflow/kubeflow/blob/master/kubeflow/jupyter/prototypes/jupyter-web-app.jsonnet

In the case where we have multiple prototypes within the same ksonnet package, I'm splitting the prototypes into separate bases.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/8)


swiftdiaries:
___
*[kustomization.yaml, line 5 at r1](https://reviewable.io/reviews/kubeflow/manifests/8#-La8B-HC0LuW-Oa4geSY:-La8DJLk1lr_0h4ApVFR:b4w71ga) ([raw file](https://github.com/kubeflow/manifests/blob/ce0b92cb5f3269ed678d79475381ee4b93cc3907/kustomization.yaml#L5)):*
<details><summary><i>Previously, jlewi (Jeremy Lewi) wrote…</i></summary><blockquote>

Do we have an issue tracking fixing the reference to the profiles controller?
</blockquote></details>

This is the issue for the profiles controller: https://github.com/kubeflow/manifests/issues/7

swiftdiaries:
___
*[jupyter-web-app/base/deployment.yaml, line 19 at r1](https://reviewable.io/reviews/kubeflow/manifests/8#-La8BMVo-Pz29852k2bo:-La8DSEI7R0MXqT0UQBv:b26jfx) ([raw file](https://github.com/kubeflow/manifests/blob/ce0b92cb5f3269ed678d79475381ee4b93cc3907/jupyter-web-app/base/deployment.yaml#L19)):*
<details><summary><i>Previously, jlewi (Jeremy Lewi) wrote…</i></summary><blockquote>

Should we avoid setting the image here and only set it in the kustomization file so that its easy to override and there is a single location where its being set?
</blockquote></details>

Do we want to do this for all deployments? We have a default image that we can modify using a kustomization file.

swiftdiaries:
___
*[jupyter-web-app/base/service.yaml, line 5 at r1](https://reviewable.io/reviews/kubeflow/manifests/8#-La8BjirBRCkT5-RYrBW:-La8EZPi6u2u4Ghnawjh:b-xy3pgw) ([raw file](https://github.com/kubeflow/manifests/blob/ce0b92cb5f3269ed678d79475381ee4b93cc3907/jupyter-web-app/base/service.yaml#L5)):*
<details><summary><i>Previously, jlewi (Jeremy Lewi) wrote…</i></summary><blockquote>

Is there anyway in YAML to do a multi-line literal?
</blockquote></details>

Yes, we should be able to change this to something more readable. Let me get that in

kustomization.yaml
```diff
@@ -1,7 +1,8 @@
 bases:
 - ambassador/base
 - argo/base
-- github.com/swiftdiaries/kubeflow//components/profile-controller/config?ref=profiles-controller
+- jupyter-web-app/base
+#- github.com/swiftdiaries/kubeflow//components/profile-controller/config/default?ref=profiles-controller
```
lluunn: should we remove this comment?

jupyter-web-app/base/service.yaml
```diff
@@ -0,0 +1,26 @@
...
+  annotations:
+    getambassador.io/config: |-
+      ---
+      apiVersion: ambassador/v0
+      kind:  Mapping
+      name: webapp_mapping
+      prefix: /jupyter/
+      service: jupyter-web-app.kubeflow
+      add_request_headers:
+       x-forwarded-prefix: /jupyter
```
lluunn: indent one more space?

kustomization.yaml
swiftdiaries: It's currently failing because remote builds don't work well with relative paths like `../`
We need pull the config/ folder of the profiles-controller from kfctl to enable this so I thought it'd be commented till then

jupyter-web-app/base/service.yaml
swiftdiaries: Thanks :) I'll commit that

PR #9, opened 2019-03-18
Adds notebook-controller manifest

swiftdiaries:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/9)


notebook-controller/base/kustomization.yaml
```diff
@@ -0,0 +1,18 @@
+namePrefix: notebooks-controller-
+
+resources:
```
lluunn: is it common pattern to have one resource per file?

swiftdiaries: Yeah, one set of resources per component

swiftdiaries:
/cc @lluunn @jlewi 
I think this is good to be merged

PR #12, opened 2019-04-02
add katib

hougangliu:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/12)

### DIFF
--- a/katib/v1alpha1/base/katib-ui-deployment.yaml
+++ b/katib/v1alpha1/base/katib-ui-deployment.yaml
@@ -1,0 +1,25 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: katib-ui
+  namespace: kubeflow
+  labels:
+    app: vizier
+    component: ui
+spec:
+  replicas: 1
+  template:
+    metadata:
+      name: katib-ui
+      labels:
+        app: vizier
+        component: ui
+    spec:
+      containers:
+      - name: katib-ui
+        image: gcr.io/kubeflow-images-public/katib/katib-ui:v0.1.2-alpha-156-g4ab3dbd
+        command:
+          - './katib-ui'
+        ports:
+        - name: ui
+          containerPort: 80
--- a/OWNERS
+++ b/OWNERS
@@ -1,5 +1,6 @@
 approvers:
   - abhi-g
+  - hougangliu
   - jlewi
   - kkasravi
   - kunmingg
--- a/katib/v1alpha1/base/katib-db-pvc.yaml
+++ b/katib/v1alpha1/base/katib-db-pvc.yaml
@@ -1,0 +1,10 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: katib-mysql
+spec:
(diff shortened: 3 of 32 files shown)


katib/v1alpha1/base/katib-ui-deployment.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: katib-ui
+  namespace: kubeflow
```
kkasravi: can we pass this as a var?

katib/v1alpha1/base/katib-ui-deployment.yaml
```diff
@@ -0,0 +1,25 @@
...
+  template:
+    metadata:
+      name: katib-ui
+      labels:
+        app: vizier
+        component: ui
+    spec:
+      containers:
+      - name: katib-ui
+        image: gcr.io/kubeflow-images-public/katib/katib-ui:v0.1.2-alpha-156-g4ab3dbd
```
kkasravi: should this be a parameter?

jlewi:
/assign @richardsliu 
/assign @johnugeorge 

@hougangliu Could you update the PR description to call out the differences between the customizations enabled between the current ksonnet package and the kustomize package?

katib/v1alpha1/base/katib-ui-deployment.yaml
swiftdiaries: We can do `kustomize edit set namespace <new namespace>` inside the `katib/v1alpha1/base` folder to set it to a different namespace.

swiftdiaries:
@hougangliu small nit - can we change the directory structure to have `katib/base` and the v1alpha1 or any other version can be an overlay of that?
Then, if the user so wishes to change vresions, they can simple pick the right overlay from the root kustomization file.

katib/v1alpha1/base/katib-ui-deployment.yaml
hougangliu: it can be overwritten by `namespace` in kustomization.yaml

katib/v1alpha1/base/katib-ui-deployment.yaml
hougangliu: it can be overwritten by same image name in images list of kustomization.yaml

hougangliu:
> /assign @richardsliu
> /assign @johnugeorge
> 
> @hougangliu Could you update the PR description to call out the differences between the customizations enabled between the current ksonnet package and the kustomize package?

sure

hougangliu:
> @hougangliu small nit - can we change the directory structure to have `katib/base` and the v1alpha1 or any other version can be an overlay of that?
> Then, if the user so wishes to change versions, they can simply pick the right overlay from the root kustomization file.

@swiftdiaries thanks for your comment.
now only katib v1alpha1 can work. And we're working on v1alpha2, which has big change compared with v1alpha1. It's hard to extract a common katib base. And I think after v1alpha2 works solidly, we will drop v1alpha1.

PR #13, opened 2019-04-06
port to kustomize, initial checkin

kkasravi:
see [First Class Support for Kustomize Packages In Kubeflow](https://docs.google.com/document/d/156H79GiQ-gCAputHzcCdmD6gZZaMBLqC-kzwKqc2MIA/edit#heading=h.c4khqss4mjzp)

Components have been initially generated using `ks show default -c <component>`. 
This output is redirected to a `<package>/<component>/resources.yaml`.
Params from the respective `<component>.jsonnet` are converted to vars or json patches.
See the above document for details.

Kfctl will traverse these directories to find and build kustomize targets based on it's configuration file `app.yaml`. Each target processed by kfctl will result in an output yaml file. The output file is generated by calling kustomize's API.  The kustomize package manager in kfctl will read app.yaml and apply the packages, components and componentParams to kustomize in the following way:

- **packages**
  - are always top-level directories under the manifests repo
- **components**
  - are also directories but may be a subdirectory in a package.
  - components may also be a top-level directory if there is a base or overlay in that directory in which case the component name is equal to the package name.
  - otherwise a component is a sub-directory
  - in all cases a component's name in app.yaml must match the directory name.
  - components are output as `<component>.yaml` under the kustomize subdirectory during `kfctl generate...`.
  - in order to output a component a synthetic kustomization.yaml is created under manifests that sets common parameters and labels that looks like
```
kind: Kustomization
bases:
  - <component>/{base|overlay/<platform>}
commonLabels:
  app.kubernetes.io/name: <appName>
namespace:
  <namespace>
```
- **component parameters**
  - are applied to a component's params.env file. There must be an entry whose key matches the component parameter. The params.env file is used to generate a ConfigMap. Entries in params.env are resolved as kustomize vars or referenced in a deployment or statefulset env section in which case no var definition is needed.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/13)


jlewi:
Are these being auto-generated in some way? Is it just the result of doing ks show?

jlewi:
Looks like you're out of date.

Can you update the PR description to explain how you are generating them? I think that might be useful from the perspective of figuring out how we might redo the process later on if needed to pull in latest changes.

Also if we're doing a sort of bulk conversion then I think it makes sense to not submit individual PRs for the packages because we aren't doing a lot of specific package customization.

jlewi:
@kkasravi Is this ready? On my end I think I'm just looking for an updated PR description.

kkasravi:
other than generating the Application CR, it's ready to review. 
I have done ~20 components. Let me update the PR with what's been done

swiftdiaries:
Great work @kkasravi ! 
Could we break out the `resources.yaml` into their respective components? So that users can get a picture of what they're installing.
Commented on the design doc for the same

argo/base/argo-wfc-configmap.yaml
```diff
@@ -1,27 +1,27 @@
 apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: workflow-controller-configmap
+  namespace: kubeflow
 data:
   config: |
     {
-    executorImage: argoproj/argoexec:v2.2.0,
+    executorImage: $(executorImage),
```
jlewi: Is $(executorImage) built in kustomize functionality or is this part of the KFP you are writing?

What's the point of using substitution here? The entire file is basically a substitution. If the user needs to customize this config map why don't they just customize it directly rather than doing a layer of indirection and configure parameters?

argo/kustomization.yaml
```diff
@@ -0,0 +1,94 @@
...
+    newName: argoproj/workflow-controller
+    newTag: v2.2.0
+configMapGenerator:
+- name: workflow-controller-parameters
+  env: params.env
+generatorOptions:
+  disableNameSuffixHash: true
+vars:
+- name: executorImage
+  objref:
```
jlewi: Why doesn't the user just modify the ConfigMap rather than setting all the parameters here?

common/centraldashboard/kustomization.yaml
```diff
@@ -0,0 +1,30 @@
...
+- base/role.yaml
+- base/deployment.yaml
+- base/rolebinding.yaml
+- base/service.yaml
+commonLabels:
+  kustomize.component: centraldashboard
+images:
+  - name: gcr.io/kubeflow-images-public/centraldashboard
+    newName: gcr.io/kubeflow-images-public/centraldashboard
+    newTag: latest
```
jlewi: Any particular reason you are defaulting to latest here? I think we always to pin to specific images so that the behavior is deterministic.

gcp/cert-manager/kustomization.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
jlewi: cert-manager isn't really GCP specific 
Filed #19 to fix this later

gcp/gcp-credentials-admission-webhook/kustomization.yaml
```diff
@@ -0,0 +1,28 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
jlewi: Filed #20 to move this out of the gcp package.

jupyter/jupyter-web-app/base/service.yaml
```diff
@@ -7,20 +7,17 @@ metadata:
       apiVersion: ambassador/v0
       kind:  Mapping
       name: webapp_mapping
-      prefix: /jupyter/
-      service: jupyter-web-app.kubeflow
+      prefix: /$(prefix)/
```
jlewi: Are we sure we want to make this configurable? It might be assumed in other places in Kubeflow.
I don't think that needs to block this PR though.

jupyter/notebook-controller/base/cluster-role-binding.yaml
```diff
@@ -1,12 +1,11 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name: jupyter-web-app-binding
+  name: role-binding
```
jlewi: Is a common prefix going to be added by kustomize?

kubebench/kustomization.yaml
```diff
@@ -0,0 +1,38 @@
...
+    newName: gcr.io/kubeflow-images-public/kubebench/kubebench-operator
+    newTag: v0.4.0-13-g262c593
+  - name: gcr.io/kubeflow-images-public/kubebench/kubebench-controller
+    newName: gcr.io/kubeflow-images-public/kubebench/kubebench-controller
+    newTag: v0.4.0-13-g262c593
+  - name: gcr.io/kubeflow-images-public/kubebench/kubebench-example-tf-cnn-post-processor
+    newName: gcr.io/kubeflow-images-public/kubebench/kubebench-example-tf-cnn-post-processor
+    newTag: v0.4.0-13-g262c593
+
+# @optionalParam serviceAccount string null The service account used to run the job
```
jlewi: Why is this commented out?

metacontroller/resources.yaml
```diff
@@ -0,0 +1,335 @@
+---
```
jlewi: Should we move this out of the metacontroller package? Filed #21

tensorboard/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
jlewi: Per https://github.com/kubeflow/manifests/issues/18

I  might suggest not porting over manifests for tensorboard, TFServing, TFJob, Katib, Seldon Deployment, until we figure out what we want to do here.

jlewi:
@kkasravi This PR is pretty impressive in its comprehensiveness! What's the path to getting this merged? i.e what's the goal for this PR? I assume we won't be able to get everything working in this PR; can we define a good end state for this PR and then use follow on issues/PRs to tweak the individual packages and shard the work?

Could you update the PR description with more information about the updates you are applying across a large number of manifests. I think I see the following

  * Stripping out namespace from resources; presumably because kustomize will apply it?
  * Removing application specific prefixes from resource names; presumably because kustomize will add a common prefix?
   * Might be worth documenting these in the README

Did you intend to delete notebook-controller?

kkasravi:
@jlewi I think this is a good first cut and has been rebased with master.
Subsequent ksonnet components can go in new issues/prs.
Thanks Kam

argo/kustomization.yaml
jlewi: Filed #23

jlewi:
Agreed.

I filed additional issues to follow up on.

/lgtm
/approve

PR #17, opened 2019-04-11
add profiles

hougangliu:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/17)


jlewi:
/assign @swiftdiaries 
/assign @kkasravi 
/unassign @jlewi 

@hougangliu Could you update the PR description to explain how the manifests were created? e.g. were they hand crafted or did you follow some process to generate them?

profiles/overlay/kustomization.yaml
```diff
@@ -0,0 +1,5 @@
+bases:
+- ../base
```
swiftdiaries: Could we avoid relative paths?
This breaks remote build for kustomize 2.X
Related kubebuilder issue: https://github.com/kubernetes-sigs/kubebuilder/issues/595

hougangliu:
> @hougangliu Could you update the PR description to explain how the manifests were created? e.g. were they hand crafted or did you follow some process to generate them?
porting kubeflow ksonnet counterpart, for example as below, list manifest of profiles:
```
 ks show  default -c profiles
```

profiles/overlay/kustomization.yaml
hougangliu: kubernetes-sigs/kubebuilder#595

> Could we avoid relative paths?
[kubernetes-sigs/kubebuilder#595](https://github.com/kubernetes-sigs/kubebuilder/issues/595) says that it does't support relative paths of resource in `resources` section of kustomization.yaml. For `bases` section, relative path works well.

swiftdiaries: The kustomization.yaml at the root of the repo doesn't seem interpret relative paths very well. 

For example:
If I do `kustomize build https://github.com/kubeflow/manifests/tree/1787892d71bff56287ba4a31d701a8458c06ddc5` on your PR. 
I get a `Error: evalsymlink failure` and the build fails

hougangliu: `evalsymlink failure` is not caused by relative path, but [an issue](https://github.com/kubernetes-sigs/kustomize/issues/913) of kustomize.
When kustomize build remote, it first downloads remote package into local and then build the local one. Somehow the local file tree doesn't match what kustomize build $the_local_path.

swiftdiaries: Thanks for the clarification.

swiftdiaries: Correct me if I'm wrong, is the overlay using the same image as the base?
If so, do we need the customization?

hougangliu: it's same now.
The reason why I list here is that when related images upgrade, in most case user only need update related image tag here.

swiftdiaries: If it's just the image we're customizing, then doing a `kustomize edit set image <image name>` inside the base should change the image. 
Maybe we could add this to the README, instead of having an overlay?

hougangliu: OK, I think both keeping it in overlay and README is ok.

swiftdiaries: I'm sorry for dragging this but having an overlay breaks remote builds. 
Could we have a base with the image being set via kustomize CLI for customization? 
Is there a specific reason why we need an overlay here?

swiftdiaries:
Thanks ! 
/lgtm
/approve

profiles/overlay/kustomization.yaml
hougangliu: For this case, I have no reason for overlay here. Delete it now

PR #22, opened 2019-04-17
add metacontroller

hougangliu:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/22)

### DIFF
--- a/metacontroller/base/clusterrolebinding.yaml
+++ b/metacontroller/base/clusterrolebinding.yaml
@@ -1,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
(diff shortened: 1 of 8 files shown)


metacontroller/base/clusterrolebinding.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: meta-controller-cluster-role-binding
```
kkasravi: you could probably use namePrefix: meta-controller- in the kustomization.yaml and then this would just be cluster-role-binding

hougangliu:
/close
close it since #13 including meta controller

PR #29, opened 2019-04-22
fixes to profiles

kkasravi:
- [x] use namePrefix
- [x] remove duplicate resource files
- [x] use commonLabels
- [x] add an Application Manifest

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/29)


profiles/base/deployment.yaml
```diff
@@ -1,20 +1,14 @@
...
-        app: profiles
     spec:
       containers:
       - command:
         - /manager
         image: gcr.io/kubeflow-images-public/profile-controller:v20190228-v0.4.0-rc.1-192-g1a802656-dirty-f95773
         imagePullPolicy: Always
         name: manager
-      serviceAccountName: profiles
+      serviceAccountName: profiles-service-account
```
hougangliu: s/profiles-service-account/service-account

profiles/base/crd.yaml
```diff
@@ -13,7 +11,7 @@ spec:
     shortNames:
     - prf
     singular: profile
-  scope: Cluster
+  scope: Namespaced
```
hougangliu: Profile's scope is Cluster

kkasravi: The profile-controller runs in the kubeflow namespace with the serviceAccountName of profiles-service-account. This serviceAccount is listed in the ClusterRoleBinding with the Role of cluster-admin so the profile-controller runs with cluster privileges. Submitting a Profile CR should be done to the kubeflow namespace. The profiler-controller will [reconcile](https://github.com/kubeflow/kubeflow/blob/master/components/profile-controller/pkg/controller/profile/profile_controller.go#L111) Profile CRs in the kubeflow namespace. The user submitting the Profile CR will be listed as a subject in the RoleBinding in the kubeflow namespace - either as a User or Group in GCP IAM. This way the user submitting the ProfileCR doesn't need cluster level RBAC.

Note - I need to add the rolebinding, role, and serviceaccount back in for 'user|group' level operations.

profiles/base/service-account.yaml
```diff
@@ -1,6 +1,10 @@
+---
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  labels:
-    app: profiles
-  name: profiles
+  name: controller-service-account
```
hougangliu: I cannot find where controller-service-account is used

profiles/base/crd.yaml
hougangliu: I still persist Profile's scope is Cluster. in its controller, Profile's name is used to create new ns, role and role binding. there is no any reference to its namespace.

lluunn: @kkasravi Kam if the cluster level RBAC we are giving is just creating Profile (which is cluster resource), is that a concern? Also, if it's namespaced, we will limit it to only exist in kubeflow namespace right?
I think both cluster or namespaced should work, but cluster is more intuitive to me.
WDYT @jlewi @kunmingg

kkasravi: Hi Lunkai - if the user submits a Profile with no namespace they will need a clusterrolebinding, the clusterrole can be very weak but it still exists at the cluster level. If they submit a Profile with the kubeflow <deployment> namespace then they use a rolebinding in kubeflow's <deployment> namespace. The role doesn't allow much more than create for Profiles. The profile-controller will then create a RoleBinding in the new profile namespace that grants that user 'edit' rights in their own namespace. So the user has two RoleBindings - 1 in the kubeflow deployment namespace and 1 in their own namespace. The Role in the kubeflow <deployment> prevents any type of damage or leakage across 'tenants', the role in their own namespace also prevents 'tenant' leakage since they have no cluster rights. Should they try and escalate their rights, eg change their role, they're still bound by a namespace. But under cluster scope, if they try and escalate their clusterrole, now they can do much more damage and tenant leakage is global. So the threat model is quite different with clusterroles.

kkasravi: This was the original design writeup - https://github.com/kubeflow/kubeflow/tree/master/kubeflow/profiles/prototypes

kkasravi:
I'm going to release the hold. If there is a strong preference for Cluster on the Profile I can change it back. We could revisit the issue with pros/cons and design impacts in a subsequent issue/pr. Let me know.

PR #30, opened 2019-04-23
service namespace var in getambassador.io/config annotations

hougangliu:
Now, if a service need to be routed by ambassador, $(service-name).$(service-namespace) is in need in its annotations. in kustonmize, service namespace is configured during build. so we'd better set service namespace in its annotations by a var.
before this PR, the namespace is set by configmap, so we must modify the configmap manually each time when we set a new namespace for the application. but if the namespace comes from its service metadata.namespace, we can never modify manually for it any more.

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/30)

### DIFF
--- a/kubebench/base/kustomization.yaml
+++ b/kubebench/base/kustomization.yaml
@@ -25,3 +25,13 @@ images:
   - name: gcr.io/kubeflow-images-public/kubebench/kubebench-example-tf-cnn-post-processor
     newName: gcr.io/kubeflow-images-public/kubebench/kubebench-example-tf-cnn-post-processor
     newTag: v0.4.0-13-g262c593
+vars:
+- name: namespace
+  objref:
+    kind: Service
+    name: kubebench-dashboard
+    apiVersion: v1
+  fieldref:
+    fieldpath: metadata.namespace
+configurations:
+- params.yaml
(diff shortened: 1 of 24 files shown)


kubebench/base/kustomization.yaml
```diff
@@ -25,3 +25,13 @@ images:
   - name: gcr.io/kubeflow-images-public/kubebench/kubebench-example-tf-cnn-post-processor
     newName: gcr.io/kubeflow-images-public/kubebench/kubebench-example-tf-cnn-post-processor
     newTag: v0.4.0-13-g262c593
+vars:
```
kkasravi: Nice, could you also update jupyter/jupyter/base/kustomization.yaml, jupyter-web-app/base/kustomization.yaml? They reference the namespace in the configmap, we might as well be consistent.

```
- name: namespace
  objref:
    kind: ConfigMap
    name: parameters
    apiVersion: v1
  fieldref:
    fieldpath: data.namespace
```

jlewi:
nit: Could you update the PR description please to explain why this change is needed? Imagine someone not familiar with kustomize and/or Ambassador.

kubebench/base/kustomization.yaml
hougangliu: good catch! thanks

hougangliu:
> nit: Could you update the PR description please to explain why this change is needed? Imagine someone not familiar with kustomize and/or Ambassador.

Thanks! Done

PR #31, opened 2019-04-25
Support for multiple overlays

kkasravi:
Support for multiple overlays. Paired with kubeflow/kubeflow/pull/3108

Multiple overlays -

This PR adds multiple overlays to the Profiles manifest
```
profiles
├── base
│   └── kustomization.yaml
└── overlays
    ├── debug
    │   └── kustomization.yaml
    └── devices
        └── kustomization.yaml
```

Normally kustomize provides the ability to overlay a 'base' set of resources with changes that are merged into the base from resources that are located under an overlays directory
if the target is named foo there will be a foo/base/ and a foo/overlays/bar.
both foo/base and foo/overlays/bar have kustomization.yaml files
running kustomize build in foo/base will generate resources listed in kustomization.yaml
running kustomize build in foo/overlays/bar will generate resources - some of which will overlay the resources in foo/base.
kustomize doesn't provide for an easy way to combine more than one overlay for example foo/overlays/bar, foo/overlays/baz
the ability to combine more than one overlay is key to handling components like tf-job-operator and types of overlays that 'mix-in' debug, device info, etc.

Merging multiple overlays
Since each overlay references foo/base as its base - including several overlays will result in kustomize complaining that it recursed on foo/base. The approach is to create a foo/kustomization.yaml that includes foo/base under its bases and has the attributes of foo/overlays/bar/kustomization.yaml and  foo/overlays/baz/kustomization.yaml moved to foo/kustomization.yaml. This requires some path corrections and some awareness of the behavior of configMapGenerator, secretMapGenerator thats copied from each overlay kustomization.yaml.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/31)


hougangliu:
@kkasravi can you rebase the PR and remove #29 diff in this PR so that we can better review it? Thanks

hougangliu:
I think maybe we can drop this PR since #34 can cover use case of this one

kkasravi:
I think we need unit tests, the profile use case differs from the tf-job-operator use case.
In the tf-job-operator I used one overlay which references as it's base another overlay.
In profile I used 2 overlays which needed to be merged.

jlewi:
Can you update the PR description please

* I'm not sure which README.md you are refering to 

* Please explain what you mean by support for multiple overlays

* Please explain what is changing in this PR to support multiple overlays

jlewi:
/lgtm
/approve
/hold
In case someone with more kustomize expertise should review.

We should plan on documenting how we use multiple overlays somewhere.

jlewi:
/lgtm
/approve

@hougangliu Can you please review as the kustomize expert and remove the hold if this looks good?

PR #32, opened 2019-04-25
Add kustomize config for kubeflow/metadata

zhenghuiwang:
https://github.com/kubeflow/metadata/issues/11

/cc @kkasravi @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/32)


metadata/base/deployment.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: metadata-deployment
+  labels:
+      deployment: metadata
```
hougangliu: please use `commonLabels` in kustomization.yaml  to replace all labels/matchLabels

metadata/base/kustomization.yaml
```diff
@@ -0,0 +1,4 @@
+resources:
+- deployment.yaml
+- service.yaml
+
```
hougangliu: use `namePrefix: metadata-`

metadata/base/kustomization.yaml
```diff
@@ -0,0 +1,4 @@
+resources:
```
hougangliu: we'd better also add below two lines in this file:
```
kind: Kustomization
```

metadata/base/service.yaml
```diff
@@ -0,0 +1,12 @@
+kind: Service
+apiVersion: v1
+metadata:
+  name: metadata-service
+spec:
+  selector:
+    deployment: metadata
+  type: LoadBalancer
```
swiftdiaries: Is it possible to make this NodePort? It might be better for non-cloud clusters.

metadata/base/kustomization.yaml
zhenghuiwang: This is not necessary since is it is under "metadata" dir.

metadata/base/deployment.yaml
zhenghuiwang: Added it. Thanks for the review!

metadata/base/service.yaml
zhenghuiwang: This is an initial pr to set up the kustomize files. I set it as "LoadBalancer" as other components. We can add overlay to customize it later if it is needed.

metadata/base/kustomization.yaml
hougangliu: yes, it is not necessary, but as [best practice](https://github.com/kubeflow/manifests#Best-practices-for-kustomize-targets), we'd better follow this rule.

zhenghuiwang: To follow the rule, I will make the change.

but I think the rule means to be used for a directory containing multiple targets. Grouping files by targets using the same prefix makes sense, e.g. /katib/base. 

For a directory with a single target, the prefix adds overhead but no value. e.g. /tensorboard/base/ gets it right.

zhenghuiwang:
@hougangliu @swiftdiaries thanks for the comments. PTAL.

swiftdiaries:
@zhenghuiwang I don't see the resources files. I only see the kustomization.yaml for metadata.

swiftdiaries:
Also, could you add a namePrefix and rename the deployment and service like it's done [here](https://github.com/kubeflow/manifests/blob/master/modeldb/base/kustomization.yaml#L1)

metadata/base/metadata-service.yaml
```diff
@@ -0,0 +1,14 @@
+kind: Service
+apiVersion: v1
+metadata:
+  labels:
+    app: metadata
+  name: metadata-service
```
hougangliu: owning to `namePrefix: metadata-`, this service will be named `metadata-metadata-service`, so I think name it `service` is ok here, similar as above Deployment

zhenghuiwang: Done!

zhenghuiwang:
@hougangliu @swiftdiaries Any changes need to be done for approving it?

swiftdiaries:
No lgtm, thanks :) 
/approve

PR #40, opened 2019-05-02
Katib UI ISTIO

gabrielwen:
Fixes kubeflow/kubeflow#2963

/assign @kkasravi @hougangliu 
/cc @richardsliu

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/40)

### DIFF
--- a/katib/base/katib-ui-virtual-service.yaml
+++ b/katib/base/katib-ui-virtual-service.yaml
@@ -2,7 +2,6 @@ apiVersion: networking.istio.io/v1alpha3
 kind: VirtualService
 metadata:
   name: katib-ui
-  namespace: kubeflow
 spec:
   gateways:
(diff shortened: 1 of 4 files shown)


katib/base/katib-ui-virtual-service.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: networking.istio.io/v1alpha3
+kind: VirtualService
+metadata:
+  name: katib-ui
+  namespace: kubeflow
```
hougangliu: we'd better remove `namespace` for [Best-practice](https://github.com/kubeflow/manifests#Best-practice-for-kustomize-targets)

gabrielwen: removed.

PR #42, opened 2019-05-03
Pipeline fixes

kkasravi:
Rework pipeline so that parameters mysqlPd and minioPd are used as vars.

Use suggested best practices for nameprefix, commonLabels per pipeline target.
Remove the pipeline bundling since this prevented mysqlPd, minioPd and namespace resolution.
All pipeline targets are now individually named within the generated config when --packagemanager=kustomize

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/42)


pipeline/mysql/base/kustomization.yaml
```diff
@@ -1,7 +1,9 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+nameprefix: mysql-
```
hougangliu: I think we can rename it as `pipeline-mysql-` since other kubeflow components (saying kabit) has mysql, too

swiftdiaries: +1 on the rename. At least until https://github.com/kubeflow/kubeflow/issues/2422 is solved

pipeline/minio/overlays/minioPd/kustomization.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+nameprefix: minio-
+bases:
+- ../../base
```
swiftdiaries: This might break remote deployments  like this: `kustomize build github.com/kubeflow/manifests/pipeline/...`
Maybe open an issue for this to track?

swiftdiaries:
Minor nit, otherwise looks good

kkasravi:
could i get a lgtm @hougangliu, @swiftdiaries, @ashahba ?

ashahba:
That sounds good @kkasravi

PR #59, opened 2019-05-09
Adding test harness for Kustomize targets

ashahba:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/59)

PR #63, opened 2019-05-10
Set up prow for manifests

hougangliu:
For now, make test cannot work, so ignore its error in test, after #59 fix it, please uncomment corresponding part in run-tests.sh @ashahba 

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/63)


tests/workflows/environments/releasing/params.libsonnet
```diff
@@ -0,0 +1,21 @@
+local params = import '../../components/params.libsonnet';
+
+params + {
+  components+: {
```
kkasravi: may want to pass in params here

tests/scripts/run-tests.sh
```diff
@@ -0,0 +1,27 @@
...
+# This shell script is used to build an image from our argo workflow
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+cd tests
+
+set +o errexit # when "make all" can work, remove this line
+make all
```
kkasravi: this is where we may want to get params passed in

PR #67, opened 2019-05-13
add application-controller and application overlays (disabled by default)

kkasravi:
Note: This feature is disabled by default.

One can enable this feature by editing the app.yaml before `kfctl generate...` and 
adding `application` to components, packages. This will deploy the application-controller.

Application CR's can be emitted for each kustomize target as an overlay.
Each target is able to deploy an Application CR by adding an application 
overlay to their componentParams in app.yaml.
(example jupyter-web-app)
```
     jupyter-web-app:
    - name: overlay
      value: application
```

The dependency mapping among applications will be a follow-on PR.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/67)


jlewi:
@kkasravi Any particular reason not to emit Application CRs by default?

kkasravi:
@jlewi we can enable by default in the future, for now i thought we would add piecemeal as each one stabilizes

PR #70, opened 2019-05-14
Fix ClusterRole in Jupyter Notebook controller

andreyvelich:
I have tried to deploy Jupyter using Kustomize manifests and got the error inside Notebook controller about not enough permissions, when I try to create new Notebook from the UI. 
I think, we missed this roles in ClusterRole, since we have them here:
https://github.com/kubeflow/kubeflow/blob/master/kubeflow/jupyter/notebook_controller.libsonnet#L138.
After I add these roles UI works correctly. 
/cc @kkasravi @ashahba

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/70)


stpabhi:
@andreyvelich Could you please update the verbs to just get, list and watch? I don't think * is necessary here.
```
- apiGroups:
  - ""
  resources:
  - pods
  verbs:
  - get
  - list
  - watch
```
/cc @kkasravi

kkasravi:
Could you rebase from upstream/master please?

andreyvelich:
@stpabhi 
It doesn't work for me, controller can't create new notebook in that case. 
I think, we should also add `create`, `update`, `delete`. 
Notebook-controller makes this changes. 
Or we can leave `*` as it is. 
What do you think?

andreyvelich:
@kkasravi Done.

stpabhi:
> @stpabhi
> It doesn't work for me, controller can't create new notebook in that case.
> I think, we should also add `create`, `update`, `delete`.
> Notebook-controller makes this changes.
> Or we can leave `*` as it is.
> What do you think?

@andreyvelich I think it's already covered as part of statefulsets and deployments rule. As the controller creates StatefulSet and not a pod directly it should work.

andreyvelich:
@stpabhi Got this error from Notebook-controller, if I leave only these roles:
```
{"error":"services is forbidden: User \"system:serviceaccount:kubeflow:notebook-controller-service-account\" cannot create services in the namespace \"kubeflow\""
```

stpabhi:
@andreyvelich it is missing services rule. Could you check if your clusterrole has services rule? [here](https://github.com/kubeflow/manifests/blob/master/jupyter/notebook-controller/base/cluster-role.yaml)

andreyvelich:
@stpabhi Got it, will make changes.

andreyvelich:
@stpabhi @kkasravi Should we merge this PR?

stpabhi:
@andreyvelich lgtm

PR #76, opened 2019-05-15
Change the Role verbs of a Profile user to prevent the user from getting a Profile

### LINKED ISSUE #52: Change the Role verbs of a Profile user to prevent the user from getting a Profile
The Role should be

```
kind: Role
metadata:
  creationTimestamp: "2019-05-05T15:25:10Z"
  labels:
    app.kubernetes.io/name: experiments
    kustomize.component: profiles
  name: profiles-role
  namespace: experiments
  resourceVersion: "3645969"
  selfLink: /apis/rbac.authorization.k8s.io/v1/namespaces/experiments/roles/profiles-role
  uid: f8a2dd93-6f49-11e9-9bef-42010a8a01ef
rules:
- apiGroups:
  - kubeflow.org
  resources:
  - profiles
  verbs:
  - create
  - watch
  - list
```

kkasravi:
fixes #52

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/76)

### DIFF
--- a/profiles/base/crd.yaml
+++ b/profiles/base/crd.yaml
@@ -10,7 +10,7 @@ spec:
   names:
     kind: Profile
     plural: profiles
-  scope: Cluster
+  scope: Namespaced
   validation:
     openAPIV3Schema:
       properties:
--- a/profiles/base/role-binding.yaml
+++ b/profiles/base/role-binding.yaml
@@ -1,11 +1,11 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
-  name: role-binding
+  name: default-role-binding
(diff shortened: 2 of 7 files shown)


jlewi:
Do we want users to be able to see all profiles?
I would normally do

```
kubectl create -f profile.yaml
kubectl get profiles
```

If users can't list profiles will that lead to a frustrating UX?

i.e. if a user forgets what they called their profile how would they do a list to see what it is?

jlewi:
Why not get as well?

kunmingg:
Shall we make it a ClusterRole since Profile is cluster-scope CRD?
And we can use a name like `profile-default` to indicate what it does?

PR #69, opened 2019-05-13
JWT policy for IAP ingress

gabrielwen:
Relates kubeflow/kubeflow#3122

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/69)

PR #77, opened 2019-05-15
iap-ingress is missing parameters used in ksonnet

### LINKED ISSUE #53: iap-ingress is missing parameters used in ksonnet
/manifests/gcp/iap-ingress/overlays/gcp is missing the parameters from ksonnet

ipName=
secretName=envoy-ingress-tls
hostname=
issuer=letsencrypt-prod
oauthSecretName=kubeflow-oauth
istioNamespace=istio-system
ingressName=envoy-ingress

kkasravi:
fixes #53

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/77)


gabrielwen:
could you add some description?  Seems like this PR is not only adding back parameters, you are also fixing istio namespace issue we discussed before?

gcp/iap-ingress/base/config-map.yaml
```diff
@@ -1,349 +1,74 @@
...
+      selector:
+        istio: ingressgateway
+      servers:
+      - port:
+          number: 80
+          name: http
+          protocol: HTTP
+        hosts:
+        - "*"
+  jwt-policy-template.yaml: |
```
gabrielwen: this is recently changed:
- added to ingress/base: https://github.com/kubeflow/manifests/pull/69
- removed from setup-backend and patch audience info later: https://github.com/kubeflow/kubeflow/pull/3298

gcp/iap-ingress/base/config-map.yaml
```diff
@@ -1,349 +1,74 @@
...
+          audiences:
+          - {{JWT_AUDIENCE}}
+          jwtHeaders:
+          - x-goog-iap-jwt-assertion
+          trigger_rules:
+          - excluded_paths:
+            - exact: /healthz
+            - prefix: /.well-known/acme-challenge
+      principalBinding: USE_ORIGIN
   setup_backend.sh: |
```
gabrielwen: original file for this is here: https://github.com/kubeflow/kubeflow/blob/master/kubeflow/gcp/setup_backend.sh

I'd like to use patch for this file.  in the meantime, if you wanna update the logic, could you also update the original file?

also, kubeflow/kubeflow#3298 changes the script, could you also update them here?

kkasravi: not sure I followed your suggestions. 

The [original file](https://raw.githubusercontent.com/kubeflow/kubeflow/master/kubeflow/gcp/setup_backend.sh) is using [patch](https://github.com/kubeflow/kubeflow/blob/master/kubeflow/gcp/setup_backend.sh#L58)

gabrielwen: sorry for the confusion - I meant patch of the whole `setup_backend.sh`.  Currently we hardcoded these scripts for config map, right?  Previously we maintain these scripts in separate files and import them when deployment.  I guess the analogy would be patching these scripts in kustomize?  But in any case this is not something we should include into this PR.

PR #79, opened 2019-05-16
Extend the test harness to incorporate parameters and negative testing

### LINKED ISSUE #57: extend the test harness to incorporate parameters and negative testing
currently the test harness is a pending PR #50 that generates golang test cases which compare the concatenation of all resources with the output of `kustomize build`.  This has a few limitations:

- [ ] the parameters most targets take should be passed into the target
- [ ] the other parts of the kustomization file: vars, configMapGenerator, patchesStrategicMerge, patchesJson6902 shouldn't be ignored as be the basis of additional tests
- [ ] negative testing should be done

kkasravi:
fixes #57

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/79)


ashahba:
@kkasravi I think `prow` prevents PRs from merging even after it's labeled as `lgtm` if they have been previously marked as `WIP`, so that needs to be removed 🙂

PR #90, opened 2019-05-20
Add go fmt to Makefile

ashahba:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/90)


tests/Makefile
```diff
@@ -16,17 +16,20 @@ GOLANG_VERSION ?= 1.12.4
...
 
 # Comma seperated items within {} for more than one file
-EXCLUDE ?= 
+EXCLUDE ?=
 
 export GO111MODULE = on
 export GO = go
 
-all: test
+fmt:
```
kkasravi: i think we want it after generate but not in test

PR #102, opened 2019-05-22
Istio namespace overlay

kkasravi:
- [x] Define a kustomize manifest to install ISTIO on GCP gcp/{istio,istio-crds,istio-install}
- [x] Rearrange some kustomize manifests as a work around for the fact that kustomize does not preserve install order (i.e. kubernetes-sigs/kustomize#821) ?
- [x] Change some kustomize manifests to properly set the namespace due to ingress and istio needing to be installed in the ISTIO namespace
- [x] Add virtual service resources to a bunch of the kustomize packages for webapps.
- [x] remove useIstio: false componentParams

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/102)


jlewi:
@hougangliu Can you review this so I'm not the blocker?

@kkasravi how come this PR is modifying 120 files?

kkasravi:
@jlewi this has been in the queue for a while. Most changes are minor - moving virtual-service.yaml to overlays/istio for example. @gabrielwen is the best choice to review given his familiarity with what the kustomize targets should look like

jlewi:
Is the PR description accurate? 

> if useIstio == true, adds bootstrap/config/overlays/istio which will
add components {istio,istio-install} to app.yaml
add istio as an overlay in componentParams for all components which have a virtual-service.yaml

Is that referring to your other PR to modify kfctl?

It looks to me like the changes in this PR are

* Define a kustomize manifest to install ISTIO on GCP
* Rearrange some kustomize manifests as a work around for the fact that kustomize does not preserve install order (i.e. kubernetes-sigs/kustomize#821) ?
* Change some kustomize manifests to properly set the namespace due to ingress and istio needing to be installed in the ISTIO namespace
* Add vitual service resources to a bunch of the kustomize packages for webapps.

jlewi:
Actually is this PR in sync with master?

It looks like most kustomize packages already have virtual-service.yaml file defined inside base. It looks like this PR is defining a virtual service inside the ISTIO overlay. 

But the diff doesn't show deleting any deletion of files in base?

What were your reasons for defining virtual service inside ISTIO overlays? I think that makes sense because with the goal of making each application like TFJob/pipelines installed standalone. In standalone mode we don't have a reverse proxy and may not need ISTIO so no need to add virtual services for it.

jlewi:
Looks like you are pushing commits to this PR. Let me know when its ready for another look.

PR #105, opened 2019-05-24
added Jupyter application resources

zabbasi:
needs testing.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/105)


jupyter/jupyter-web-app/base/kustomization.yaml
```diff
@@ -10,6 +10,7 @@ resources:
 - service-account.yaml
 - service.yaml
 - virtual-service.yaml
+- application.yaml
```
kkasravi: remove trailing black spaces on all kustomization.yamls and then do `make generate; make test` then tests will pass.

This has overlap with #67 but I can resolve when that goes in

kkasravi:
this is a good first pass on getting Application CRs into place. Thank you @zabbasi 
/lgtm
/approve

PR #112, opened 2019-06-05
Katib v1alpha2 kustomize package

richardsliu:
Split the katib manifests into v1alpha1 and v1alpha2, since we made a number of changes in v1alpha2.

Also removed the Istio virtual service in anticipation of #102.

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/112)


katib/v1alpha2/base/katib-db-pv.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
```
hougangliu: we don't need this file

richardsliu: Removed

PR #121, opened 2019-06-11
application target, jupyter-web-app, notebook-controller application overlays

kkasravi:
adds:
- [x] application/base
- [x] application/overlays/debug
- [x] jupyter/jupyter-web-app/overlays/application
- [x] jupyter/notebook-controller/overlays/application

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/121)


jupyter/jupyter-web-app/overlays/application/application.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: app.k8s.io/v1beta1
 kind: Application
 metadata:
-  name: jupyter-web-app-application
+  name: application
```
jlewi: Why is this an overlay rather than just part of the base package?

jupyter/jupyter-web-app/overlays/application/application.yaml
```diff
@@ -17,6 +17,22 @@ spec:
   - group: networking.istio.io
     kind: VirtualService
   descriptor:
-    type: kubeflow
-    version: ""
-    assemblyPhase: "Pending"
+    type: jupyter-web-app
+    version: v1alpha1
+    description: Replaces JupyterHub Spawner UI with a new Jupyter UI whcih enables to create/conect/delete jupyter notebooks.
```
jlewi: No need to reference JupyterHub here. Lets just describe what it does.

jupyter/notebook-controller/overlays/application/application.yaml
```diff
@@ -1,28 +1,31 @@
 apiVersion: app.k8s.io/v1beta1
 kind: Application
 metadata:
-  name: "application"
+  name: application
```
jlewi: Why is this becoming an overlay?

jlewi:
You're probably right that it might be good to be a little defensive and give a way to disable things in case it causes problems.

> it requires a number of labels for each kustomize target
the instance label may need some kfctl suffix generation

Can you point me at the labels you are referring to?

PR #124, opened 2019-06-11
add pytorch-job application CR

### LINKED ISSUE #120: PyTorch Job needs an Application CR
Need an Application CR in the pytorch package.
https://github.com/kubeflow/manifests/tree/master/pytorch-job

kkasravi:
fixes #120

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/124)


pytorch-job/pytorch-operator/overlays/application/application.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: app.k8s.io/v1beta1
```
jlewi: Does it make sense for the Application CR to be an overlay? Why not just create it by default?

kkasravi: here are some of my thoughts of why we want the application overlay to be loosely bound.

1. It may vary by platform (I know we only offer a flavor of an application right now)
2. Different users may feel different 'applications' involve several manifests
3. They may be application-controllers that provide different behaviors
4. It's not required for the component to function (and could be viewed similarly as logging or istio)

pytorch-job/pytorch-operator/overlays/application/application.yaml
```diff
@@ -0,0 +1,18 @@
...
+  name: pytorch-operator-application
+spec:
+  type: pytorch-operator
+  componentKinds:
+  - group: core
+    kind: ConfigMap
+  - group: apps
+    kind: Deployment
+  - group: core
+    kind: ServiceAccount
```
hougangliu: just wonder why we ignore other k8s resource here, such as clusterrole, clusterrolebinding, service and so on?

pytorch-job/pytorch-operator/overlays/application/kustomization.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+bases:
+- ../../base
+resources:
+- application.yaml
+commonLabels:
```
hougangliu: will spec.selector of `Application` be filled with these commonLabels? I know deployment spec.selector can be filled automatically, but now sure `Application`

jlewi:
/lgtm
/approve
/hold
Adding a hold because @hougangliu has some comments.

pytorch-job/pytorch-operator/overlays/application/application.yaml
richardsliu: I think it is because those resource types are not namespaced.

pytorch-job/pytorch-operator/overlays/application/kustomization.yaml
kkasravi: yes Application will have these labels that match the kustomize target the application.yaml is located in

pytorch-job/pytorch-operator/overlays/application/application.yaml
kkasravi: yes correct
We would need to expand the application-controller to include cluster resources

kkasravi:
@hougangliu please review

kkasravi:
@jlewi rebased and regenerated tests - needs another /lgtm

PR #125, opened 2019-06-11
Adding monitoring config to manifests

johnugeorge:
This contains minor version fixes and monitoring config

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/125)

### DIFF
--- a/tests/pytorch-operator-base_test.go
+++ b/tests/pytorch-operator-base_test.go
@@ -107,9 +107,10 @@ spec:
     spec:
       containers:
       - command:
-        - /pytorch-operator.v1beta2
+        - /pytorch-operator.v1
         - --alsologtostderr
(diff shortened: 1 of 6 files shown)


tests/pytorch-operator-base_test.go
```diff
@@ -107,9 +107,10 @@ spec:
     spec:
       containers:
       - command:
-        - /pytorch-operator.v1beta2
+        - /pytorch-operator.v1
```
richardsliu: Can you also add the service yaml to the test? See https://github.com/kubeflow/manifests/pull/110/files#diff-5f390b526dd8deee13f3c4c19f68385dR317

johnugeorge: done

PR #128, opened 2019-06-12
Reorganize katib v1alpha1 packages

richardsliu:
Split katib v1alpha1 into the following components:
- studyjob
- suggestions
- katib-ui
- metrics-collector
- vizier-core
- vizier-db

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/128)


richardsliu:
@kkasravi Not yet - that will follow this PR. Would you like me to create that one now?

richardsliu:
@kkasravi Actually - after 0.6, kubeflow master should install katib v1alpha2 (to be added), which means the components will be different. What should we do here?

PR #129, opened 2019-06-12
Fix error in ingress component

lluunn:
- remove the name prefix because components are referencing others, e.g. certificate --> ingress, service --> backendconfig.

For https://github.com/kubeflow/kubeflow/issues/3456

Also fix health check for https://github.com/kubeflow/kubeflow/issues/3459

/cc @kkasravi @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/129)

### DIFF
--- a/gcp/iap-ingress/base/cloud-endpoint.yaml
+++ b/gcp/iap-ingress/base/cloud-endpoint.yaml
@@ -6,3 +6,4 @@ spec:
   project: $(project)
   targetIngress:
     name: $(ingressName)
+    namespace: $(istioNamespace)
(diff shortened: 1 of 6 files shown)


jlewi:
FYI @kkasravi  is going to be out for a couple of days. ; so I'll try to review it in his absence.

@lluunn Which components are named incorrectly if we use a name-prefix? I think kustomize should try to apply the prefix consistently.

That said I don't think there is much harm in dropping the nameprefix. We don't really intend for there to be multiple of these resources deployed and certainly not with in the same namespace.

So I think its fine to drop the nameprefix to get things working. I'll review this PR now and we can always ask @kkasravi to weigh in and we can make changes in the future if needed.

PR #131, opened 2019-06-12
Katib v1alpha2 kustomize package

richardsliu:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/131)


jlewi:
Should the hierarchy be

katib/
   v1alpha1/....
   v1alpha2/....

Rather than making v1alpha1 and v1alpha2 top level directories?

richardsliu:
Discussed this with @kkasravi, we decided on katib-<version> instead of katib/version because the top level directory is the package, and lower level directory is the component.

jlewi:
@richardsliu filed #136 to fix the manifests blocked by kubeflow/kubeflow#3478

/lgtm

It looks like E2E tests for kubeflow/kubeflow might be failing because of issues with the manifest kubeflow/kubeflow#3475

PR #140, opened 2019-06-17
pipelines: add PVC with default StorageClass in MySQL, Minio bases

yanniszark:
The MySQL, Minio components in the pipelines package need a PVC to work.
Currently, that PVC is created inside the minioPd, mySqlPd overlays.
This only works in GCP enviroments.

This PR refactors the components so that the PVC is created inside the base with a default StorageClass.
The overlays then edit the PVC with their custom values using a strategicMergePatch.

This means that:
* Base will work for clusters with a default storage class.
* The GCP overlay is still working as it was before.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/140)

### DIFF
--- a/pipeline/minio/base/deployment.yaml
+++ b/pipeline/minio/base/deployment.yaml
@@ -27,4 +27,4 @@ spec:
       volumes:
       - name: data
         persistentVolumeClaim:
-          claimName: minio-pv-claim
+          claimName: $(minioPvcName)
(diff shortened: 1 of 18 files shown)


yanniszark:
@hougangliu thanks for the review ! :smile:  
I fixed the tests + a mistake I had in the minio component.
I also tested deploying it and it worked.
Can you please take another look ?

hougangliu:
/lgtm
/approve
thanks

PR #156, opened 2019-06-19
Prof crd mismatch

kunmingg:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/156)

### DIFF
--- a/profiles/base/crd.yaml
+++ b/profiles/base/crd.yaml
@@ -10,7 +10,7 @@ spec:
   names:
     kind: Profile
     plural: profiles
-  scope: Namespaced
+  scope: Cluster
   validation:
     openAPIV3Schema:
       properties:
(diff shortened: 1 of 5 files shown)


profiles/base/crd.yaml
```diff
@@ -10,7 +10,7 @@ spec:
   names:
     kind: Profile
     plural: profiles
-  scope: Namespaced
+  scope: Cluster
```
hougangliu: about this change, we had discussed in https://github.com/kubeflow/manifests/pull/29#discussion_r277858777
@kkasravi WDYT?

kunmingg: Latest CRD on controller side is Cluster scope:
https://github.com/kubeflow/kubeflow/blob/master/components/profile-controller/config/crds/kubeflow_v1alpha1_profile.yaml

PR #188, opened 2019-06-28
Deploy metadata ui

rileyjbauer:
Deploy the metadata UI service and configure istio routing.

Work in progress, do not merge.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/188)

### DIFF
--- a/metadata/overlays/istio/virtual-service.yaml
+++ b/metadata/overlays/istio/virtual-service.yaml
@@ -10,9 +10,9 @@ spec:
   http:
   - match:
     - uri:
-        prefix: /metadata/
+        prefix: /metadata
     rewrite:
-      uri: /metadata/
+      uri: /metadata
     route:
     - destination:
         host: $(service).$(ui-namespace).svc.$(ui-clusterDomain)
--- a/metadata/overlays/istio/virtual-service.yaml
+++ b/metadata/overlays/istio/virtual-service.yaml
@@ -1,0 +1,21 @@
+apiVersion: networking.istio.io/v1alpha3
+kind: VirtualService
+metadata:
+  name: ui
+spec:
+  gateways:
+  - kubeflow-gateway
+  hosts:
+  - '*'
+  http:
+  - match:
+    - uri:
+        prefix: /metadata/
+    rewrite:
+      uri: /metadata/
+    route:
+    - destination:
+        host: $(service).$(ui-namespace).svc.$(ui-clusterDomain)
+        port:
+          number: 80
+    timeout: 300s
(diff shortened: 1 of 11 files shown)


metadata/overlays/istio/virtual-service.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: networking.istio.io/v1alpha3
+kind: VirtualService
+metadata:
+  name: ui
```
prodonjs: Can we rename this `metadata-ui` so it's consistent with some of the other VirtualServices (ie. `ml-pipeline-ui`, `katib-ui`, etc.)?

rileyjbauer: Yes, I would actually prefer to do as you suggested, but I wasn't sure if I should do to the presence of a name-prefix in the kustimzation.yaml in`base`
(https://kubectl.docs.kubernetes.io/pages/reference/kustomize.html#nameprefix)
Perhaps it'd be better to just do away with that as well since it makes things less clear when skimming the files

prodonjs: I may not have done it properly when I added it to my cluster. I just ran `kustomize build overlays/istio/ | kubectl apply -f -` and it showed up as `ui`. Perhaps when it's actually run with `kfctl` it respects the prefix set in `base`.

*full-disclosure I'm shooting from the hip on this stuff just trying to get things to work!*

metadata/overlays/istio/virtual-service.yaml
```diff
@@ -0,0 +1,21 @@
...
+  gateways:
+  - kubeflow-gateway
+  hosts:
+  - '*'
+  http:
+  - match:
+    - uri:
+        prefix: /metadata/
+    rewrite:
+      uri: /metadata/
```
kunmingg: Do we have common path prefix `/metadata/` on metadata UI? Or rewrite should be `/`?

zhenghuiwang:
The tests are failing. You may want to to to the `manifests/tests` folder to run `make generate` to generate the actual yaml.

jlewi:
@rileyjbauer Any update? What is the test failure?

zhenghuiwang:
@rileyjbauer is OOO this week and may not be able to revise this pr in time I guess.

To unblock the 0.6 cut, I created a copy of this PR here https://github.com/kubeflow/manifests/pull/191 with unit tests fixed.  /cc @prodonjs

kunmingg:
@zhenghuiwang Can we close here?

rileyjbauer:
Closed in favor of PR #191

PR #191, opened 2019-07-02
Metadata UI manifests

zhenghuiwang:
Copy of https://github.com/kubeflow/manifests/pull/188 by @rileyjbauer with unit tests fixed.

Deploy the metadata UI service and configure istio routing.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/191)

### DIFF
--- a/metadata/base/metadata-ui-deployment.yaml
+++ b/metadata/base/metadata-ui-deployment.yaml
@@ -15,8 +15,7 @@ spec:
         app: metadata-ui
     spec:
       containers:
-      # TODO(rjbauer): Change this to something like gcr.io/metadata/frontend:0.0.1
-      - image: gcr.io/kubeflow-images-public/metadata-frontend:v0.0.1
+      - image: gcr.io/kubeflow-images-public/metadata-frontend:v0.1.8
         imagePullPolicy: IfNotPresent
         name: metadata-ui
         ports:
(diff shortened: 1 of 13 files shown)


metadata/base/metadata-ui-deployment.yaml
```diff
@@ -0,0 +1,25 @@
...
+      app: metadata-ui
+  template:
+    metadata:
+      name: ui
+      labels:
+        app: metadata-ui
+    spec:
+      containers:
+      # TODO(rjbauer): Change this to something like gcr.io/metadata/frontend:0.0.1
+      - image: gcr.io/kubeflow-images-public/metadata-frontend:v0.0.1
```
prodonjs: `v0.1.8` is the most recent image with the filtering and sorting support. Can we update this to point to that image and remove the TODO. Alternatively, I can remove the other versions and make the latest image `v0.1.7` to match the backend service although I don't anticipate they will still in sync for long.

metadata/overlays/istio/virtual-service.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: networking.istio.io/v1alpha3
+kind: VirtualService
+metadata:
+  name: ui
```
prodonjs: As I was chatting with @rileyjbauer the other day, I think we need to rename this to `metadata-ui` because it does not appear that the `namePrefix: metadata-` from the base  kustomization.yaml file gets applied here and the prefix definitely helps to disambiguate this from other UI services.

PR #195, opened 2019-07-03
Kubeflow On-Premise Authentication and Authorization Prototype

krishnadurai:
Implemented Scope:

- Sets up dex, ldap, dex client application with the help of kustomize in auth namespace
- Documents Kubernetes OIDC Authentication setup
- Applies Kubernetes Authorization through RBAC
- Applies Istio configurations for Authenticating and Authorization on ml-pipeline service in kubeflow namespace

Please follow README.md for instuctions to setup and run.

TODOs:
- [x] Add instructions for self signed certificate work-around
- [x] Add instructions for LDAP users setup
- [x] Usage walk-through guide (Going to be a part of Kubeflow blog)
- [x] Parameterize Kustomize overlay edits
- [x] Runnable through kfctl

Design Doc: http://bit.ly/2WPs1MS
Demo presentation link: http://bit.ly/31Td2V3

**Deployment Setup**

*Note*: This implementation's target platforms are Kubernetes clusters with access to modify Kubernetes' API config file, which is generally possible with on Premise installations of Kubernetes.

- Setup a Kubernetes cluster on a VM:
  - This can be done by deploying a VM in any cloud vendor like GCP or AWS etc.
  - A reference link to setup Kubernetes on this VM is mentioned [here](https://linuxconfig.org/how-to-install-kubernetes-on-ubuntu-18-04-bionic-beaver-linux).
  - You can set this up within a few minutes.

- Once we have the Kubernetes cluster ready, follow these steps in the README:
  - [Server Setup Instructions](https://github.com/kubeflow/manifests/blob/51b0e3f8f35053181ca5a357c4273cacd67c1df9/common/dex-auth/README.md#server-setup-instructions)

cc: @jlewi @yanniszark @animeshsingh @johnugeorge 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/195)


ryandawsonuk:
In the implementation in kfctl presently there's code which waits for the LoadBalancer IP to become available - https://github.com/kubeflow/kubeflow/blob/master/bootstrap/v2/pkg/kfapp/existing_arrikto/existing.go#L392

Presumably that's not present in this PR because here there's a domain so you know in advance what you're securing. But presumably that means you have to have a domain. And you need to put that domain in as per[ 'Replace *example.org* with your own domain.'](https://github.com/kubeflow/manifests/pull/195/files#diff-46b10a690795758a8db6b0f86c9afddcR16). Wouldn't it be better to make that a parameter to kfctl?

I'm also curious whether this could be parameterised to allow further redirectUris such as in https://github.com/kubeflow/kubeflow/pull/3607 (which is also relevant for being able to make calls from outside the cluster such as with https://github.com/kubeflow/kfserving/issues/234)

krishnadurai:
@ryandawsonuk agree with the domain being parametrised from kfctl. Would be quite intuitive for a deployer to specify this from the CLI rather than editing the overlay files. The kfctl integration is something which I'm working on presently and I will open a separate PR to integrate this effort with kfctl.
I'll shortly get back to you regarding kubeflow/kubeflow#3607

common/dex-auth/base/dex/ca_configmap.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: ca
+data:
+  ca.pem: |
```
krishnadurai: To be removed by parameterizing

common/dex-auth/base/dex/dex_configmap.yaml
```diff
@@ -0,0 +1,98 @@
...
+        inCluster: true
+    web:
+      https: 0.0.0.0:5556
+      tlsCert: /etc/dex/tls/tls.crt
+      tlsKey: /etc/dex/tls/tls.key
+    logger:
+      level: "debug"
+      format: text
+
+    connectors:
```
krishnadurai: Should have a facility to be overlayed. This ensures different connectors can be able to connect to dex. LDAP is a reference implementation.

common/dex-auth/base/dex_params.env
```diff
@@ -0,0 +1,21 @@
...
+dex_client_redirect_uri=http://login.example.org:5555/callback/onprem-cluster
+k8s_master_uri=https://k8s.example.com:443
+dex_client_listen_addr=http://127.0.0.1:5555 # Set to HTTPS if TLS is configured
+
+#TODO(krishnadurai): Paramerize certificates:
+# 1. data/ca.pem in ca configMap
+# 2. k8s_ca_pem in dex-k8s-authenticator-cm:cluster
+# 3. idp_ca_pem in dex-k8s-authenticator-cm:cluster
+# Not implemented since multiline values are not getting parsed in kustomize
+# environment file
```
krishnadurai: Certs need to be parameterized.

@kkasravi any suggestions here? Multi-line values in this file fail.
(Refer this link for details)[https://stackoverflow.com/questions/57142291/multiline-values-in-kustomize-environment-source-file]
This is required since these services can not mount these certs as a Kubernetes secret.

common/dex-auth/base/istio-auth/authentication_policy.yaml
```diff
@@ -0,0 +1,18 @@
...
+  - mtls: {}
+  origins:
+  - jwt:
+      audiences:
+        - $(dex_client_id)
+      issuer: $(issuer)
+      jwksUri: $(jwks_uri)
+  principalBinding: USE_ORIGIN
+  targets:
+  - name: ml-pipeline
```
krishnadurai: Needs to be merged with `istio` folder.

common/dex-auth/overlays/README.md
```diff
@@ -0,0 +1,44 @@
+# Editing Overlay files
```
krishnadurai: This file needs cleaning up along with overlays folder.

After parameterization, the only overlay applicable should be for specifying [connector configuration](https://github.com/kubeflow/manifests/pull/195/files#diff-237d2c060bba4c6361e74fc87f56bfc7R20) for Dex auth server.

common/dex-auth/overlays/authentication/authentication_policy.yaml
```diff
@@ -0,0 +1,12 @@
+---
```
krishnadurai: To be removed

common/dex-auth/overlays/authentication/dex.yaml
```diff
@@ -0,0 +1,126 @@
+---
```
krishnadurai: Only retain connector configuration in Dex ConfigMap.

common/dex-auth/overlays/authentication/dex_k8s_authenticator.yaml
```diff
@@ -0,0 +1,97 @@
+---
```
krishnadurai: To be removed

istio/istio/base/params.yaml
```diff
@@ -1,3 +1,3 @@
 varReference:
-- path: spec/mode
-  kind: ClusterRbacConfig
+- path: data/ca.pem
```
krishnadurai: To be reverted

krishnadurai:
@kkasravi I have restructured the folders and have created parameters for Kustomize.

yanniszark:
Thanks for the great PR @krishnadurai.
I aim to review it asap.

common/dex-auth/base/k8s_ca_contents.pem
```diff
@@ -0,0 +1,3 @@
+  -----BEGIN CERTIFICATE-----
```
krishnadurai: Indentation intentional because this content is replaced within a string-based configuration file [here](https://github.com/kubeflow/manifests/pull/195/files#diff-8d6c50839cd5e66a2c31e2b7c18d090bR51).

common/dex-auth/base/dex_params.env
krishnadurai: Found the solution to this problem using `files` directive for kustomize.

common/dex-auth/base/dex/dex_configmap.yaml
kkasravi: Would a 'facility' be a tool?

Given your current layout
```
common/dex-auth/base
├── dex
│   ├── ca_configmap.yaml
│   ├── crd.yaml
│   ├── deployment.yaml
│   ├── dex_configmap.yaml
│   ├── kustomization.yaml
│   └── service.yaml
├── dex-k8s-authenticator
│   ├── configmap.yaml
│   ├── deployment.yaml
│   ├── kustomization.yaml
│   └── service.yaml
├── dex_ca_contents.pem
├── dex_params.env
├── dex_params.yaml
├── idp_ca_contents.pem
├── istio-auth
│   ├── authentication_policy.yaml
│   └── kustomization.yaml
├── k8s_ca_contents.pem
├── kustomization.yaml
└── ldap
    ├── README.md
    ├── admin_service.yaml
    ├── deployment.yaml
    ├── kustomization.yaml
    └── service.yaml
```

the subdirs: dex, dex-k8s-authenticator, istio-auth, ldap are bases referenced in dex/kustomization.yaml.
This is fine although bases go away in kustomize-v2.+ and just resources are used. The [BestPractices](https://github.com/kubeflow/manifests/blob/master/docs/KustomizeBestPractices.md) document suggests a directory structure along with how vars, overlays, configMaps etc should be used but if we abide by the document and current patterns in the other manifests your current layout would morph to:

```
.
├── dex-authn
│   └── base
│       ├── config-map.yaml
│       ├── deployment.yaml
│       ├── dex_ca_contents.pem
│       ├── idp_ca_contents.pem
│       ├── k8s_ca_contents.pem
│       ├── kustomization.yaml
│       ├── params.env
│       ├── params.yaml
│       └── service.yaml
├── dex-crds
│   └── base
│       ├── config-map.yaml
│       ├── crds.yaml
│       ├── deployment.yaml
│       ├── kustomization.yaml
│       └── service.yaml
└── dex-ldap
    └── base
        ├── deployment.yaml
        ├── kustomization.yaml
        └── service.yaml
```

What makes this PR challenging is that a number of new patterns are introduced 
1. tooling is embedded under dex-auth/certs. The admission-controller generates a cert using admission-webhook/bootstrap/base/config-map.yaml - you should do something similar. Or there is code in controller-runtime that creates a cert I think.
2. Istio related rules are inverted - rather than being under a particular manifest target like ./pipeline/pipelines-ui/overlays/istio, the manifest target is referenced out-of-context under ./dex-auth/authorization/Istio/.
3. common/dex-auth/authorization/Kubernetes looks like it holds ClusterRoles, ClusterRoleBindings where the subjects are groups: admin, validator, datascience
4. The ldap README.md references the groups mentioned in 3, but outlines a set of steps the admin needs to do to prior to using kfctl which defines a mapping of groups to RBAC. The problem is this isn't integrated with profiles or istio so i don't think it would work - at least not in gcp.  I'm guessing there isn't a kubeflow/bootstrap/config/*.yaml that kfctl uses where dex-auth is deployed and works with ldap. 

I think I need to go through the ldap use-case more carefully and deploy it via a config.file

krishnadurai:
> Hi @krishnadurai
> 
> For the ServiceRole and ServiceRoleBinding you provide for ml-pipeline, is the suggestion that all targets would need ServiceRole, ServiceRoleBindings and you've only provided an example? 

@kkasravi 'ml-pipeline' is just an example of how to apply these RBAC policies to a service. The idea here is to extend this to all services within Kubeflow where we aim to restrict access through Istio RBAC.

>Would this only be for dex authenticated users?

This should be generic for any authentication as long as we should be able to have [subjects](https://istio.io/docs/reference/config/policy-and-telemetry/templates/authorization/#Subject) and have our authorization rules refer to them.

>Does only providing ml-pipeline mean that other services would need to have something similar?

Yes other services will need to have something similar. We would need to define a ServiceRole and its ServiceRoleBinding.

krishnadurai:
@kkasravi Thanks for this detailed review and I'm looking to address these comments shortly. A few clarifications regarding the scope here:

- This PR intends to integrate Dex, Dex Login Client and LDAP as Kubeflow components.
- I will raise a separate PR for integrating this with kfctl with its own config file.
- The RBAC configs in `common/dex-auth/authorization/Kubernetes` and `common/dex-auth/authorization/Istio` are here to serve as an example for its usage. I can move them to an examples folder. Here is an [issue](https://github.com/kubeflow/kubeflow/issues/3614) tracking RBACs adoption for all services in Kubeflow.
- The tooling of deploying RBAC rules and the subjects it affects can be made a part of kustomize configs and linked to kfctl.

common/dex-auth/authorization/Kubernetes/secrets_write_all_cluster_role_binding.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1beta1
```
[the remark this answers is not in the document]
krishnadurai: This is an example configuration to verify the setup.

The intention here is to eventually provide a config file to kfctl to define roles like the ones mentioned in here: admin, datascience and validator. This file should be composable, though we should be supplying defaults for setup. kfctl should be the deployer for these rbac rules and roles.

kkasravi:
thanks @krishnadurai 

@jlewi @kunmingg @gabrielwen this PR is showing how istio {ServiceRole, ServiceRoleBinding} can be added to pipelines-viewer. Seems like we would want to add this in a user's profile namespace via a controller - like what notebook-controller does. It also seems like the ServiceRole and Role for the user should share similar rules.

common/dex-auth/authorization/Kubernetes/cluster_read_all_cluster_role.yaml
```diff
@@ -0,0 +1,57 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: cluster-read-all-role
```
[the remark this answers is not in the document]
krishnadurai: This is an example role to read all Kubernetes resources. We are planning to specify these through the effort [here](https://github.com/kubeflow/kubeflow/issues/3614).

Lets move it to an examples folder.

common/dex-auth/base/dex/dex_configmap.yaml
krishnadurai: > Would a 'facility' be a tool?

Is there a way of overlaying a config from kfctl? Otherwise, we might need to introduce tooling around it. 

> Given your current layout

I'll adopt this structure according to v2.

> 
> What makes this PR challenging is that a number of new patterns are introduced
> 
>     1. tooling is embedded under dex-auth/certs. The admission-controller generates a cert using admission-webhook/bootstrap/base/config-map.yaml - you should do something similar. Or there is code in controller-runtime that creates a cert I think.

Will include this as an optional part of kfctl instead.

>     2. Istio related rules are inverted - rather than being under a particular manifest target like ./pipeline/pipelines-ui/overlays/istio, the manifest target is referenced out-of-context under ./dex-auth/authorization/Istio/.

I'll make this change.
 
>     3. common/dex-auth/authorization/Kubernetes looks like it holds ClusterRoles, ClusterRoleBindings where the subjects are groups: admin, validator, datascience

I'll move these to examples folder.

>     4. The ldap README.md references the groups mentioned in 3, but outlines a set of steps the admin needs to do to prior to using kfctl which defines a mapping of groups to RBAC. The problem is this isn't integrated with profiles or istio so i don't think it would work - at least not in gcp.  I'm guessing there isn't a kubeflow/bootstrap/config/*.yaml that kfctl uses where dex-auth is deployed and works with ldap.
> 
> 
> I think I need to go through the ldap use-case more carefully and deploy it via a config.file

LDAP instructions are just a guideline here to setup. I'll remove them from here and possibly add these to docs.

common/dex-auth/authorization/Istio/cluster_rbac_config.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: "rbac.istio.io/v1alpha1"
+kind: ClusterRbacConfig
+metadata:
+  name: default
+spec:
+  mode: 'ON_WITH_INCLUSION'
```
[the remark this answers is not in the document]
krishnadurai: Any service or component specified under 'inclusion' would be subject to RBAC. Since this is an example for 'ml-pipeline', I haven't kept it cluster-wide.

common/dex-auth/authorization/Istio/cluster_rbac_config.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: "rbac.istio.io/v1alpha1"
```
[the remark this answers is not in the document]
krishnadurai: Answered here: https://github.com/kubeflow/manifests/pull/195#issuecomment-517153260

I'll move RBAC ServiceRole and ServiceRoleBinding to an examples folder whereas I'll integrate `cluster_rbac_config.yaml` and Istio Authentication policy under existing structure.

krishnadurai:
@kkasravi I've addressed the review comments and manually tested this feature.

I'll need some help understanding and fixing this build failure:
`kubeflow-manifests-presubmit-kfctl-go-195-efe896d-2432-e298-phase 	Failed`

As discussed earlier, this PR will be followed with another one for ingress integration for Istio and Profiles; likewise another one for kfctl.

For now I've moved the `ml-pipeline` Istio RBAC rules for reference to an examples folder.

/cc @swiftdiaries

kkasravi:
@krishnadurai it looks like the config_file is empty. Let me look at your branch

krishnadurai:
@kkasravi Do I need to pull master in my branch?

krishnadurai:
@kkasravi looks like the rebase with master fixed the build.

jlewi:
Is the design doc referenced in the PR description up to date?

The design doc referenced a Kubeflow API server. I'm not sure what that is referring to because Kubeflow today doesn't have a single APIServer. Is it a reference to the ISTIO gateway?

swiftdiaries:
It is a reference to the ISTIO gateway. 
The current PR gives a reference model to be replicated for other services within Kubeflow.

jlewi:
> Applies Istio configurations for Authenticating and Authorization on ml-pipeline service in kubeflow namespace

> @jlewi @kunmingg @gabrielwen this PR is showing how istio {ServiceRole, ServiceRoleBinding} can be added to pipelines-viewer. Seems like we would want to add this in a user's profile namespace via a controller - like what notebook-controller does. It also seems like the ServiceRole and Role for the user should share similar rules.

Could you explain what roles and bindings are created for the pipelines-viewer?

I believe right now what we are doing is granting access at the namespace level.

* Profiles resource will create a service role that grants access to all services in the namespace
  
  Here's a sample
  ```
apiVersion: rbac.istio.io/v1alpha1
kind: ServiceRole
metadata:
  annotations:
    role: admin
    user: jlewi@google.com
  creationTimestamp: 2019-07-17T13:55:51Z
  generation: 1
  name: ns-access-istio
  namespace: jlewi-at-google-com-3ddace
  ownerReferences:
  - apiVersion: kubeflow.org/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: Profile
    name: jlewi-at-google-com-3ddace
    uid: 96427576-a89a-11e9-9bbc-42010af000d6
  resourceVersion: "4723"
  selfLink: /apis/rbac.istio.io/v1alpha1/namespaces/jlewi-at-google-com-3ddace/serviceroles/ns-access-istio
  uid: 96455830-a89a-11e9-9bbc-42010af000d6
spec:
  rules:
  - services:
    - '*'
status: {}
  ```

* The profile will create a servicerolebinding to the user listed in the profile

Does the pipelines viewer CR require different roles and binding? If a user creates Viewer CR's in a namespace they have access to wouldn't they already be covered by the existing roles and bindings created by the profile controller?

swiftdiaries:
Yes and no.
This PR specifically focuses on on-premise deployments where an existing LDAP/AD server is used by Dex to infer user groups. These are then used to create the roles and rolebindings for each service.
The key idea here is to give users a self-serve platform using existing credentials.
Again, this is a reference model and one of the ways to go about doing this.

If an org has no such pre-defined LDAP/AD servers, then a model where the Profiles Controller would supply / create the set of roles and rolebindings makes a better fit.

kkasravi:
@krishnadurai @swiftdiaries 
There were some errors in the `make generate` which I've addressed in #260.
If you rebase or compare this PR with #260 you'll see the changes.
They are also noted below

- leaf directories expect a 'kustomization.yaml' so assets, examples were moved to docs/dex-auth. You may probably point to them from the existing README.mds.
- *.pem files weren't being processed
- a few other minor changes to address current limitations in the golang code generators.

My suggestion is to pull in the changes in #260 and repush

krishnadurai:
@kkasravi I'll make these changes right now

kkasravi:
thanks for working through these issues with me @krishnadurai - getting dex in is quite important, even if some of your work needs to be transfered into profile-controller or istio overlays. My apologies for not being able to give you more of my time to get this in quickly and not getting enough eyes from others on this work. You're a very patient person !

kkasravi:
let me know when you've pulled in changes from #260 and I can close it. 
thanks.

krishnadurai:
@kkasravi Thanks for your patience and inputs as well! It's always great to have maintainers as diligent as you in the community.
I'm eager to contribute to profiles-controller and istio-overlays as well. Please engage me wherever appropriate - simply approach me on slack.

I apologise for the time taken here - I was looking to avoid the force push - though I couldn't.

We can close #260.

krishnadurai:
@googlebot I consent.

krishnadurai:
@jlewi 

> Could you explain what roles and bindings are created for the pipelines-viewer?
In the context of this PR, we have set roles and role bindings as mentioned here:

```
apiVersion: "rbac.istio.io/v1alpha1"
kind: ServiceRole
metadata:
  name: ml-pipeline-viewer
  namespace: kubeflow
spec:
  rules:
  - services: ["ml-pipeline.kubeflow.svc.cluster.local"]

---
apiVersion: "rbac.istio.io/v1alpha1"
kind: ServiceRoleBinding
metadata:
  name: bind-ml-pipeline-viewer
  namespace: kubeflow
spec:
  subjects:
  - group: "datascience"

  roleRef:
    kind: ServiceRole
    name: "ml-pipeline-viewer"
```
Consider if there are multiple 'roles' required here:
- A role which can view pipelines
- A role which can add pipelines
- A role which can delete pipelines

All such roles can be specified in a ServiceRole and be applied to respective subjects: users or groups - even if they are in the same shared namespace.

> Does the pipelines viewer CR require different roles and binding?

The answer to this might come down to whether we intend for namespaces to be exclusive to one user (role) or to be shared by multiple roles in an organisation with different access levels.

>  If a user creates Viewer CR's in a namespace they have access to wouldn't they already be covered by the existing roles and bindings created by the profile controller?

They should be covered if we intend to give that user full access to all resources in that namespace.
Here, we could target specific resources within the namespace and apply access control to them, like the scenarios discussed above.

jlewi:
Thanks @swiftdiaries ;

If this PR is about showing how LDAP can be integrated into Kubeflow to support on prem users; would it be better to start with notebooks rather than pipelines?

* Multi-user support in pipelines is still an open issue kubeflow/pipelines#1223
* Pipelines uses its own API server (rather than custom resources) so pipelines access would be controlled via ISTIO as opposed to RBAC.
* Kubeflow's current multi-user model is users/team consume Kubeflow in their own kubeflow-${TEAM} namespace
  * This namespace is where notebooks and TFJobs etc... run
  * The controllers and UIs run in the Kubeflow namespace
  * I don't think pipelines supports this model yet; I think in pipelines the Argo workflows are still running in the kubeflow namespace  
* Currently profile controller create's 2 RBAC role bindings in kubeflow-${TEAM}
  * default-editor - Permission to create/delete any K8s resource in that namespace
  * default-viewer - Permission to view K8s resources

Would it make sense in this PR to focus on integrating that model with LDAP? Specifically creating the appropriate LDAP integrations?

jlewi:
Looking at the PR it looks like the ML Pipelines roles and rolebindings are treated as example manifests; that seems fine to me. 

As noted above I think we should try to align the on prem story with the overall kubeflow multi-user story; i.e. restricting access to namespaces kubeflow-${TEAM}; but that seems more like a documentation issue. IIUC user manually creates the appropriate roles and role-bindings for LDAP groups so we'd just need to document some examples and setup which I think we could do in follow on PRs.

I think in follow on PRs/issues we should try to align this work with the existing_arrikto work; that would solve issues like generating the self-signed certificates; existing_arrikto could also install DEX using this kustomize manifest.

So on my end I don't think major revisions on this PR are required; other than maybe moving dex out of commons/ and making dex a top level package.

krishnadurai:
Thanks @jlewi for your review. I agree that the on prem story should tie into the existing multi-user story first and I will make required changes to enable it.

Though should integration with multi-user be a subject of another PR which I can extend to include ingress support? It would be better to limit the scope of this PR to Dex LDAP integration.

kkasravi:
@jlewi @krishnadurai @swiftdiaries 
if we move /manifests/common/dex-auth to /manifests/dex-auth we should probably move the other components under common to top-level as well (in a separate issue/pr).
I had originally suggested to @krishnadurai that he put this under common to match basic-auth

krishnadurai:
> your PR should also include my fix to hack/gen-test-target.sh for reading *.pems. Nice work.

Thanks! This has been included.

> GKE and other cloud-providers would be pretty important to get working with dex - at least banzai cloud claims to work with GKE among others https://beta.banzaicloud.io

I'll clarify. I meant to say that the Kubernetes API (kubectl) based RBAC rules would not be applicable to GKE with the current implementation. Otherwise Dex Auth should work with Istio Origin Authentication.

krishnadurai:
> Should dex be a top level directory in kubeflow/manifests rather than a subdirectory of commons?

I'll move this to top level along with the other auth mechanisms in common as @kkasravi has suggested, though in a separate PR.

> Does this need to be checked in?
> Shouldn't there be really good support for dealing with secrets with kustomize that avoids checking them in.

The issue here is some of the certs need to be populated as strings in config.yaml files for Dex. Now, these files are placeholders to be relpaced with conents from cert files - especially if they are not self-signed.

> 
> _[docs/dex-auth/examples/gencert.sh, line 3 at r10](https://reviewable.io/reviews/kubeflow/manifests/195#-LlgRNuFEBs5XUdWTOtQ:-LlgRNuFEBs5XUdWTOtR:b-uc6vc7) ([raw file](https://github.com/kubeflow/manifests/blob/ddf29e6157c07cb766b7626d8834b4dea13b6b0d/docs/dex-auth/examples/gencert.sh#L3)):_
> 
> > ```shell
> > #!/bin/bash
> > 
> > # TODO(krishnadurai): Remove this file as soon as cert tooling is introduced in kfctl
> > ```
> 
> Can you file an issue tracking this and list the issue in the TODO?

https://github.com/kubeflow/kfctl/issues/6

I'll commit this upsteam.

> _[docs/dex-auth/examples/README.md, line 1 at r10](https://reviewable.io/reviews/kubeflow/manifests/195#-LlgRXbi-MWxLu78vs5u:-LlgRXbi-MWxLu78vs5v:b-r6yf1n) ([raw file](https://github.com/kubeflow/manifests/blob/ddf29e6157c07cb766b7626d8834b4dea13b6b0d/docs/dex-auth/examples/README.md#L1)):_
> 
> > ```gfm
> > # Using Examples to enable RBAC
> > ```
> 
> Should this documentation go in kubeflow/website?

I am thinking whether it is appropriate to add it in the website before it is tagged and released. Should we wait for the documentation until then?

> _[docs/dex-auth/examples/README.md, line 24 at r10](https://reviewable.io/reviews/kubeflow/manifests/195#-LlgRj2E2-Ou1K5y6F0y:-LlgRj2E2-Ou1K5y6F0z:b-kyv4jh) ([raw file](https://github.com/kubeflow/manifests/blob/ddf29e6157c07cb766b7626d8834b4dea13b6b0d/docs/dex-auth/examples/README.md#L24)):_
> 
> > ```gfm
> > ## Setup Istio RBAC
> > 
> > Currently, the only service authenticated and authorized supported is ml-pipeline service.
> > ```
> 
> See comment:
> [#195 (comment)](https://github.com/kubeflow/manifests/pull/195#issuecomment-519116600)
> 
> Kubeflow already has really good multi-users/team support for TFJob and notebooks.
> Should we start with that so the LDAP/on-prem story is well aligned with the rest of Kubeflow?

https://github.com/kubeflow/manifests/pull/195#issuecomment-519132196

krishnadurai:
@jlewi @kkasravi if you can have a look once more, we should be able to merge this PR soon.

kkasravi:
thanks @krishnadurai - i'll try and get to it this afternoon and hopefully do lgtm+approve

jlewi:
@krishnadurai there are a couple open comment threads in reviewable can you please take a look.

Overall, I think I'd be fine with this PR going in as is and filing issues to resolve issues going forward.

Here are the big issues

1. Is Kubeflow actually using OIDC to authorize Kubernetes APIServer RBAC requests?

    * Which Kubeflow applications support using OIDC as opposed to authorizing requests based on the service account the web server is using?

    * Should we open an issue to figure out whether OIDC is something we actually want to support and if so what needs to happen?

1. I think we are already creating viewer and edit roles; so I'm not sure we need the clusterrole's for viewer and editor that you are creating. We should open an issue to track resolving that and consolidating on a single set of roles



> The issue here is some of the certs need to be populated as strings in config.yaml files for Dex. Now, these files are placeholders to be relpaced with conents from cert files - especially if they are not self-signed.

Is PEM the public part or the secret part? If its the public part then I think it makes more sense.

krishnadurai:
Build failing due to:
```
fatal: could not create leading directories of '/mnt/test-data-volume/kubeflow-manifests-presubmit-kfctl-go-195-6b73549-4000-820c/src/kubeflow/manifests': No spac
e left on device
```

krishnadurai:
@jlewi  Can you please have a look again? I've addressed all comments regarding certs, scripting and OIDC usage.

> Here are the big issues
> 
>     1. Is Kubeflow actually using OIDC to authorize Kubernetes APIServer RBAC requests?
>        
>        * Which Kubeflow applications support using OIDC as opposed to authorizing requests based on the service account the web server is using?

This facility for authorization is for requests through kubectl. Services implemented through CRs like tfjobs are still operable through kubectl.

>        * Should we open an issue to figure out whether OIDC is something we actually want to support and if so what needs to happen?

We can create an issue for this and discuss. My thinking here is that until the time there will be a solution for kubectl requests, this can be a solution.

>     2. I think we are already creating viewer and edit roles; so I'm not sure we need the clusterrole's for viewer and editor that you are creating. We should open an issue to track resolving that and consolidating on a single set of roles

A tracking issue for this has been created: #267 

> Is PEM the public part or the secret part? If its the public part then I think it makes more sense.

This is the public part for the CA certificate. I've now moved this to config maps without the use of file parameter like before in kustomize. We'll automate this using kfctl shortly.

For now I've added docs to enable this: 
[Docs](https://github.com/kubeflow/manifests/pull/195/files#diff-ba6ae59126e45ea192c878f275d056c1R62-R76)

Note: Build has been failing due to test infra's disk being full.

/cc @kkasravi

tests/dex-crds-base_test.go
```diff
@@ -0,0 +1,322 @@
...
+data:
+  config.yaml: |
+    issuer: $(issuer)
+    storage:
+      type: kubernetes
+      config:
+        inCluster: true
+    web:
+      https: 0.0.0.0:5556
+      tlsCert: /etc/dex/tls/tls.crt
```
yanniszark: I believe the current practice is to do tls termination at the Istio Gateway / Ingress and not use certificates for each individual application.
Because of that, can we remove this setting or put it in an overlay?
If we need additional security in the future, we can use Istio for mTLS in the kubeflow namespace.

krishnadurai:
@yanniszark thanks for your review.

Are you suggesting that the base dex config should rather have a mockCallback connector in base and an `ldap` overlay can contain LDAP as a connector?

yanniszark:
@krishnadurai I was thinking no connectors at all.
Maybe the base config can have support for a static user (or we could also put that in an overlay).

jlewi:
@krishnadurai  Thanks

/lgtm on my end
/hold for other reviewers

tests/dex-crds-base_test.go
krishnadurai: I understand. I've suggested an alternative for the base to just contain a mock callback and the overlay to have LDAP. Though the challenge here is that we may need to provide tooling to overlay instead of adding them to manifests.

krishnadurai:
@yanniszark I agree with the base having static user. 
I have a suggestion - please correct my understanding here if I'm wrong. This would result in multiple overlays for each type of connector - ldap, gcp etc. I suggest instead of adding each connector type to manifests can we define tooling for users to define this configuration? This way we would be able to limit the number of overlays in manifests.

yanniszark:
> This would result in multiple overlays for each type of connector - ldap, gcp etc.

Correct.

> I suggest instead of adding each connector type to manifests can we define tooling for users to define this configuration?

What would this tooling do exactly?
If possible, I'd like to avoid custom templating logic and rely on kustomize.

krishnadurai:
@yanniszark 
I was looking at a way to provide a separate dex config file as an argument to kfctl config for specifying dex configuration. This file-path could be added to the kfctl config like the way we are doing for kustomize parameters.

yanniszark:
>  I was looking at a way to provide a separate dex config file as an argument to kfctl config for specifying dex configuration. This file-path could be added to the kfctl config like the way we are doing for kustomize parameters.

This sounds interesting. I think we could even support this right now, the whole Dex config could be one environment variable, defined in params.env.

However, for large changes, I feel that it's better to follow the kustomize philosophy and make a new overlay.
With the new kfctl config, users can make their own overlay in a remote repo and point to it in their config file.

krishnadurai:
Build failing due to kubeflow/manifests#271

krishnadurai:
@yanniszark then the remote repository solution should be fine for overlaying.

PTAL for the overlay change. I've introduced staticPassword config in base instead of connectors, as discussed before.

common/dex-auth/dex-crds/base/deployment.yaml
```diff
@@ -0,0 +1,36 @@
...
+        - name: tls
+          mountPath: /etc/dex/tls
+      volumes:
+      - name: config
+        configMap:
+          name: dex
+          items:
+          - key: config.yaml
+            path: config.yaml
+      - name: tls
```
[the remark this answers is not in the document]
krishnadurai: Sure I'll correct this config

krishnadurai:
@yanniszark can you please have a look at the complete changes at one pass? I aim to address all comments.

krishnadurai:
@yanniszark PTAL. I've addressed your comments and raised an issue for Istio gateway integration for Dex.

yanniszark:
> @yanniszark I agree that it would be better to expose Dex and related services through a separate Istio Gateway.

Thanks for raising an issue.
I actually meant exposing Dex with a VirtualService through the Kubeflow Gateway.
Similarly to other applications in manifests:
1. Define an Istio Overlay for Dex.
2. Add the VirtualService for Dex to that overlay. 
3. Make the Dex Service ClusterIP.

What do you think?

krishnadurai:
@yanniszark 
> I actually meant exposing Dex with a VirtualService through the Kubeflow Gateway.
> Similarly to other applications in manifests:
> 
>     1. Define an Istio Overlay for Dex.
> 
>     2. Add the VirtualService for Dex to that overlay.

Should Dex virtual service be linked to Kubeflow Gateway? Would it be better to have a different gateway for modularity?

>     3. Make the Dex Service ClusterIP.

I've retained this as NodePort as an alternative to the ingress gateway. This wouldn't prevent any functionality which ClusterIP provides since NodePort is built as an extension to ClusterIP.

By the way, we should implement the ingress gateway for Dex.

krishnadurai:
@kkasravi @jlewi @yanniszark If there are no more comments we should look to merge this PR. Waiting for your approvals.

ryandawsonuk:
Will this option be provided as an alternative to the 'existing_arrikto' installation option or will that option be removed in preference to this one? I'm guessing this isn't simply a modification to that setup (as I don't see any use here of the oidc_authentication_service that the existing_arrikto setup uses).

yanniszark:
@krishnadurai I'm ok with integrating Dex with Istio in a followup PR.
Today is a day off, so I'll take a final pass tomorrow and lgtm, but I think there isn't something else from my side.

> Will this option be provided as an alternative to the 'existing_arrikto' installation option or will that option be removed in preference to this one?

Hi @ryandawsonuk, great question!
We'd like for existing_arrikto to also use this package for Dex.
This should be possible once we implement a plugin mechanism for kfctl.

> Also, is there any aim to use letsencrypt or cert-manager?

Absolutely.

krishnadurai:
@ryandawsonuk 
Refer [comment](https://github.com/kubeflow/manifests/pull/195#issuecomment-519126587)
We are planning to combine the work. This PR has a different login application named dex-authenticator.

We have a tracking issue here for the certificate management: https://github.com/kubeflow/kfctl/issues/6

krishnadurai:
@yanniszark Sure, please take a final pass over the configuration - so that this is compatible for all users.

yanniszark:
@krishnadurai  thanks for all the work in this PR.
One final comment for something I hadn't noticed before:
It seems that Dex is installed in a separate namespace, called auth.
Wouldn't it be better to reuse the `kubeflow` namespace, instead of introducing another one?

Otherwise, lgtm.
@krishnadurai @jlewi @kkasravi what do you think?

kkasravi:
@krishnadurai I think you'll need to fix the CLA issue
/lgtm
/approve

krishnadurai:
@yanniszark I had used the 'auth' namespace with the view that dex could be used by other services as well in the kubernetes cluster - especially when one of the options is to use dex for Kubernetes OIDC auth. This is similar to the way Istio is installed in 'istio-system' namespace.

@kkasravi you will have to consent to the CLA bot by commenting @googlebot I consent.

jlewi:
@krishnadurai If you want to fix  the CLA you could just rebase

I will also cancel the hold but in this case you or @kkasravi could have canceled the hold rather than waiting for me to do it since the point of the hold was to give the other reviews time to review; I already LGTM'd

/hold cancel

krishnadurai:
@jlewi thanks for clearing this PR.

PR #204, opened 2019-07-09
fix basic auth

kunmingg:
Resolves https://github.com/kubeflow/kubeflow/issues/3562

Most changes are from https://github.com/kubeflow/kubeflow/pull/3404

Related: https://github.com/kubeflow/kubeflow/pull/3631

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/204)

### DIFF
--- a/common/ambassador/base/deployment.yaml
+++ b/common/ambassador/base/deployment.yaml
@@ -2,6 +2,7 @@ apiVersion: apps/v1beta1
 kind: Deployment
 metadata:
   name: ambassador
+  namespace: istio-system
 spec:
   replicas: 3
   template:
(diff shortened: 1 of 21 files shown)


kunmingg:
/assign @jlewi 
/assign @lluunn 
Ready for review

common/ambassador/base/deployment.yaml
```diff
@@ -2,6 +2,7 @@ apiVersion: apps/v1beta1
 kind: Deployment
 metadata:
   name: ambassador
+  namespace: istio-system
```
lluunn: If we didn't specify namespace, how did it end up in namespace kubeflow?
I mean, shouldn't there be a way to set namespace for all components

kunmingg: Currently kfctl code use kubeflow ns by default

PR #211, opened 2019-07-16
switch argo to kubeflowpipeline docker hub

IronPan:
This is a follow up PR that change the third party images to GCR
https://github.com/kubeflow/kubeflow/pull/3665
https://github.com/kubeflow/pipelines/pull/1622

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/211)

### DIFF
--- a/argo/base/deployment.yaml
+++ b/argo/base/deployment.yaml
@@ -35,7 +35,7 @@ spec:
           value: "true"
         - name: BASE_HREF
           value: /argo/
-        image: argoproj/argoui:v2.3.0
+        image: gcr.io/ml-pipeline/argoui:v2.3.0
         imagePullPolicy: IfNotPresent
         name: argo-ui
         resources: {}
@@ -90,7 +90,7 @@ spec:
             fieldRef:
               apiVersion: v1
               fieldPath: metadata.namespace
-        image: argoproj/workflow-controller:v2.3.0
+        image: gcr.io/ml-pipeline/workflow-controller:v2.3.0
         imagePullPolicy: IfNotPresent
         name: workflow-controller
         resources: {}
--- a/argo/base/deployment.yaml
+++ b/argo/base/deployment.yaml
@@ -35,7 +35,7 @@ spec:
           value: "true"
         - name: BASE_HREF
           value: /argo/
-        image: gcr.io/ml-pipeline/argoui:v2.3.0
+        image: kubeflowpipeline/argoui:v2.3.0
         imagePullPolicy: IfNotPresent
         name: argo-ui
         resources: {}
@@ -90,7 +90,7 @@ spec:
             fieldRef:
               apiVersion: v1
               fieldPath: metadata.namespace
-        image: gcr.io/ml-pipeline/workflow-controller:v2.3.0
+        image: kubeflowpipeline/workflow-controller:v2.3.0
         imagePullPolicy: IfNotPresent
         name: workflow-controller
         resources: {}
--- a/argo/base/kustomization.yaml
+++ b/argo/base/kustomization.yaml
@@ -11,11 +11,11 @@ resources:
 commonLabels:
   kustomize.component: argo
 images:
-  - name: argoproj/argoui
-    newName: argoproj/argoui
+  - name: gcr.io/ml-pipeline/argoui
+    newName: gcr.io/ml-pipeline/argoui
     newTag: v2.3.0
-  - name: argoproj/workflow-controller
-    newName: argoproj/workflow-controller
+  - name: gcr.io/ml-pipeline/workflow-controller
+    newName: gcr.io/ml-pipeline/workflow-controller
     newTag: v2.3.0
 configMapGenerator:
 - name: workflow-controller-parameters
--- a/argo/base/kustomization.yaml
+++ b/argo/base/kustomization.yaml
@@ -11,11 +11,11 @@ resources:
 commonLabels:
   kustomize.component: argo
 images:
-  - name: gcr.io/ml-pipeline/argoui
-    newName: gcr.io/ml-pipeline/argoui
+  - name: kubeflowpipeline/argoui
+    newName: kubeflowpipeline/argoui
     newTag: v2.3.0
-  - name: gcr.io/ml-pipeline/workflow-controller
-    newName: gcr.io/ml-pipeline/workflow-controller
+  - name: kubeflowpipeline/workflow-controller
+    newName: kubeflowpipeline/workflow-controller
     newTag: v2.3.0
 configMapGenerator:
 - name: workflow-controller-parameters
--- a/argo/base/params.env
+++ b/argo/base/params.env
@@ -1,5 +1,5 @@
 namespace=
-executorImage=argoproj/argoexec:v2.3.0
+executorImage=gcr.io/ml-pipeline/argoexec:v2.3.0
 artifactRepositoryBucket=mlpipeline
 artifactRepositoryKeyPrefix=artifacts
 artifactRepositoryEndpoint=minio-service.kubeflow:9000
--- a/argo/base/params.env
+++ b/argo/base/params.env
@@ -1,5 +1,5 @@
 namespace=
-executorImage=gcr.io/ml-pipeline/argoexec:v2.3.0
+executorImage=kubeflowpipeline/argoexec:v2.3.0
 artifactRepositoryBucket=mlpipeline
 artifactRepositoryKeyPrefix=artifacts
 artifactRepositoryEndpoint=minio-service.kubeflow:9000
--- a/pipeline/minio/base/deployment.yaml
+++ b/pipeline/minio/base/deployment.yaml
@@ -17,7 +17,7 @@ spec:
           value: minio
         - name: MINIO_SECRET_KEY
           value: minio123
-        image: minio/minio:RELEASE.2018-02-09T22-40-05Z
+        image: gcr.io/ml-pipeline/minio:RELEASE.2018-02-09T22-40-05Z
         ports:
         - containerPort: 9000
         volumeMounts:
--- a/pipeline/minio/base/deployment.yaml
+++ b/pipeline/minio/base/deployment.yaml
@@ -17,7 +17,7 @@ spec:
           value: minio
         - name: MINIO_SECRET_KEY
           value: minio123
-        image: gcr.io/ml-pipeline/minio:RELEASE.2018-02-09T22-40-05Z
+        image: minio/minio:RELEASE.2018-02-09T22-40-05Z
         ports:
         - containerPort: 9000
         volumeMounts:
--- a/pipeline/minio/base/kustomization.yaml
+++ b/pipeline/minio/base/kustomization.yaml
@@ -21,7 +21,7 @@ vars:
   fieldref:
     fieldpath: data.minioPvcName
 images:
-- name: gcr.io/ml-pipeline/minio
+- name: minio/minio
   newTag: RELEASE.2018-02-09T22-40-05Z
 configurations:
 - params.yaml
--- a/pipeline/minio/base/kustomization.yaml
+++ b/pipeline/minio/base/kustomization.yaml
@@ -21,7 +21,7 @@ vars:
   fieldref:
     fieldpath: data.minioPvcName
 images:
-- name: minio/minio
+- name: gcr.io/ml-pipeline/minio
   newTag: RELEASE.2018-02-09T22-40-05Z
 configurations:
 - params.yaml
--- a/pipeline/mysql/base/deployment.yaml
+++ b/pipeline/mysql/base/deployment.yaml
@@ -12,7 +12,7 @@ spec:
         env:
         - name: MYSQL_ALLOW_EMPTY_PASSWORD
           value: "true"
-        image: gcr.io/ml-pipeline/mysql:5.6
+        image: mysql:5.6
         ports:
         - containerPort: 3306
           name: mysql
--- a/pipeline/mysql/base/deployment.yaml
+++ b/pipeline/mysql/base/deployment.yaml
@@ -12,7 +12,7 @@ spec:
         env:
         - name: MYSQL_ALLOW_EMPTY_PASSWORD
           value: "true"
-        image: mysql:5.6
+        image: gcr.io/ml-pipeline/mysql:5.6
         ports:
         - containerPort: 3306
           name: mysql
--- a/pipeline/mysql/base/kustomization.yaml
+++ b/pipeline/mysql/base/kustomization.yaml
@@ -20,7 +20,7 @@ vars:
   fieldref:
     fieldpath: data.mysqlPvcName
 images:
-- name: mysql
+- name: gcr.io/ml-pipeline/mysql 
   newTag: '5.6'
 configurations:
 - params.yaml
--- a/pipeline/mysql/base/kustomization.yaml
+++ b/pipeline/mysql/base/kustomization.yaml
@@ -20,7 +20,7 @@ vars:
   fieldref:
     fieldpath: data.mysqlPvcName
 images:
-- name: mysql 
+- name: mysql
   newTag: '5.6'
 configurations:
 - params.yaml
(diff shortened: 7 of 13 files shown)


jlewi:
@cheyang @swiftdiaries Thoughts on this PR?

cheyang:
@jlewi Thanks for letting me know. I think we are fine to this, because we will do the work on docker repo update like https://github.com/kubeflow/kubeflow/blob/372aede09a73bf09c8202cbb96128f843a4cf2e6/scripts/kfctl.sh#L108

jlewi:
@ironpan who's going to be responsible for sycing the images from DockerHub to GCR?

What is the expectation for future releases? I.e if Argo pushes updated images can we just update the manifests to use them or do we first need to replicate to GCR?

If we first need to replicate to GCR what's the reason for introducing that extra step?

swiftdiaries:
Sorry for the late reply.
+1 to @jlewi 's point about syncing with external releases.
If there's a way to automate the syncing to GCR, there's no overhead and would make the release process better. 
As it is, updating images is a hard task for releases. This extra step might be an added pain point for the release czar(s).

IronPan:
kubeflow pipeline will be responsible for upgrading image when needed. the argo might have backward incompatible changes and upgrading it requires cautious plan to prevent KFP from being broken. 
See https://github.com/kubeflow/pipelines/issues/1250 as an example.

IronPan:
Kubeflow Pipeline team would love to continue managing the argo components, this would prevent breaking changes in the future.

jlewi:
Thanks @IronPan 

How does replicating the images to GCR address the problem of ensuring pipelines doesn't break because Argo is upgraded? Couldn't we still pull the images from DockerHub?

To deal with the fact that pipelines has a dependency on a particular version of Argo; add a test [here](https://github.com/kubeflow/manifests/tree/master/tests) to verify that Argo is pinned to a particular version. That way the test will break if anyone tries to update Argo and alert the author that we need to verify that the change doesn't break pipelines.

IronPan:
@jlewi What is the generic framework for testing components?
KFP has complicated test infrastructure to test all KFP features preventing breaking changes in KFP repo
https://github.com/kubeflow/pipelines/tree/master/test

What is the recommended approach to run all those test to guard the change in this repo?

jlewi:
We define argo workflows
https://github.com/kubeflow/manifests/blob/master/prow_config.yaml

These Argo workflows will deploy Kubeflow using the latest manifests in kubeflow/manifest

IronPan:
@swiftdiaries I'd be happy to maintain the argo image. Do you have further concerns?

IronPan:
@jlewi about specific PR do you have further concerns?

The test infra would require a lot of work to duplicate existing test to that framework.

IronPan:
@jlewi @swiftdiaries FYI we've only bumped argo version once in the past 1.5 years. and it's initiated by Pipeline component.

IronPan:
@swiftdiaries What is the governance model for Argo/Pipeline component? Do you expecting to be owner of these components? 
If so could you ensure it align with https://github.com/kubeflow/pipelines/tree/master/manifests/kustomize in the future?

jlewi:
@IronPan I'm still unclear about what problem switching the image to GCR is supposed to solving? Is this supposed to prevent people from updating to newer versions of Argo that break KFP? How does switching to GCR prevent that?

IronPan:
@jlewi Use GCR is what google's corp policy required for it to be able to launched by https://deploy.kubeflow.cloud/

IronPan:
@swiftdiaries the purpose for this change is to ensure kubeflow project meets google's policy. Please let me know if this is a concern to you.

swiftdiaries:
I see.
If that is the case, then I think it makes sense to have an overlay that can refer the GCR images and the base can be left as it is. Since the pipelines team is going to maintain it anyway, I'm assuming it shouldn't be a problem to keep the base and the deployer overlay in sync.

IronPan:
Thanks @swiftdiaries. @jlewi can we unhold this PR?

swiftdiaries:
I believe that the PR is changing the base and not adding to an overlay. Could you please modify the PR to do that? 
My only concern is that I don't see a solid reason to change the base. 
If there's a google corp policy for the deployer then having a separate overlay seems like the best approach.

jlewi:
I agree with @swiftdiaries 

There are lots of reasons why people will want to host the images in different repositories; e.g. in DockerHub to support Chinese users.

Or to support different architectures like PowerPC(https://github.com/kubeflow/kubeflow/issues/819)

I don't think we should upstream vendor specific requirements.

I think the way to support this is to make it easy for vendors to create downstream forks. One of the big pieces to support that is

Creating a tool to do a bulk edit of the kustomization files https://github.com/kubeflow/kubeflow/issues/3210 to change the location of the images to some mirror.

Scripts to make it easy to bulk mirror all images listed in our kustomization files (https://github.com/kubeflow/kubeflow/issues/211)

The process would then be

1. Vendors mirror their images to whatever repo they want
1. They use the above tools to bulk edit the manifests
1. They push there changes to kubeflow/manfiests to a "downstream" fork they own
1. They publish a KFDef spec that refers to their fork rather than kubeflow/manifests

IronPan:
@jlewi where is the code that click-to-deploy app use? I just want to make sure that one has the GCR image for now. I thought it's using this repo but i could be wrong.

IronPan:
@jlewi I switch the image to kubeflow pipeline managed docker hub instead, assuming this is not the code that click-to-deploy app use. This ensures that if argoproj delete the image, we are not broken. 
And for the code that click-to-deploy use, we still need to use gcr image.

jlewi:
@IronPan How does moving the image to a different DockerHub registry address concerns about the image being deleted? Why would folks be less concerned about pipelines accidentally deleting the image compared to the Argo project deleting the image?

If you'd like to be opinionated about where the docker images should live; why not create a downstream fork of manifests and tell people to use that? We could probably create a branch in kubeflow/manifests if you'd like.

Do you want to check this PR into its own branch?

Using the publicly available Argo image that is built and maintained by the Argo team seems like a very sensible default. 

Would it be more productive to build out the tooling to make it easy to bulk replace the images (kubeflow/kubeflow#3210)?

How much effort do you think it would take to write a go or python script to do that? 

/cc @swiftdiaries

IronPan:
> How does moving the image to a different DockerHub registry address concerns about the image being deleted? Why would folks be less concerned about pipelines accidentally deleting the image compared to the Argo project deleting the image?

Argo can choose to delete it anytime, unlikely but possible. v.s. KF team will choose to delete only when project is cancelled.

> If you'd like to be opinionated about where the docker images should live; why not create a downstream fork of manifests and tell people to use that? 

if I understand right, we need a fork for click-to-deploy app. And that's what I plan to do. 

> Using the publicly available Argo image that is built and maintained by the Argo team seems like a very sensible default.

I disagree. 


@jlewi could you answer the question of where the code is click-to-deploy app use?

IronPan:
@jlewi I see this is going to your very personal opinionated and restricted direction that deliberately making Kubeflow Pipeline's team life harder and it's hard for me to believe this comes with a good intention. 

It's pretty much a double standard here. So let me try to flatten the logic and be very clear
1. Google has policy to review and maintain the images
2. Click-to-deploy app needs follow the Google policy
3. Because of 1 and 2, as well as the principal that "forking a branch for vendor specific config", click-to-deploy app need to use a forked branch. This is not what Kubeflow does as of today. click-to-deploy use same copy as kfctl from https://github.com/kubeflow/kubeflow/tree/master/kubeflow.
4. Why you require argo to be hosted on docker hub, while forking all other images to gcr
https://github.com/kubeflow/manifests/search?q=gcr.io%2Fkubeflow-images-public&unscoped_q=gcr.io%2Fkubeflow-images-public
instead of keeping them here 
https://hub.docker.com/u/kubeflow

what is the principal? why matters? what value does it add/deduct either way to the end user? should the effort be well spend on some better discussions than this?

IronPan:
@swiftdiaries do you have permission to push image to gcr.io/kubeflow-images-public ?

jlewi:
@IronPan I'm really sorry you feel that way and that this is creating extra work for your team. Let me try to articulate what our process is. Our process is designed to put everyone on equal footing and keep velocity high

1. Images built from source in Kubeflow can be pushed to GCR.
1. We've instrumented our test infrastructure so that our postsubmits can push images to this GCR registry
    * An example is our [notebook images](https://github.com/kubeflow/kubeflow/blob/f3b8d2337e8b29ac213fbf94e29a3783c76b5a90/prow_config.yaml#L214)
1. So anyone in the community can setup a continuous integration workflow to auto-publish their images
1. We don't have an explicit policy to use GCR only. The main reason we've used GCR todate is because it was the easiest to integrate with our test infra 
1. We have issues to mirror our images to DockerHub (kubeflow/kubeflow#211) and the primary reason we haven't done that is lack of resourcing to build out the infra 
1. Third-party images (not built from Kubeflow source) are generally pulled directly from wherever the upstream project publishes them

     * e.g. cert-manager https://github.com/kubeflow/manifests/blob/master/cert-manager/cert-manager/base/kustomization.yaml
     * modeldb https://github.com/kubeflow/manifests/blob/0d1b5c2bcfb1312d57416a61eb2b2d0ea43be554/modeldb/base/webapp-deplyment.yaml#L21
    * istio https://github.com/kubeflow/manifests/blob/0d1b5c2bcfb1312d57416a61eb2b2d0ea43be554/istio/istio-install/base/istio-noauth.yaml#L12602

IronPan:
@jlewi could you answer the question of where the code is click-to-deploy app use or will use? I want to make sure the image there is from GCR.

kkasravi:
@IronPan can this be resolved if a vendor specific config is provided which provides what registry should be used? 

The vendor specific config could live anywhere and be provided via the --config flag for kfctl. It would specify a registry value. I'm testing this approach but thought I would suggest this as a viable approach and solicit your feedback.

kkasravi:
here is an example of what I mentioned above #252 

the vendor specific config would need the argo section to look like

```
  - kustomizeConfig:
      parameters:
      - name: registry
        value: kubeflowpipeline
      overlays:
      - istio
      repoRef:
        name: manifests
        path: argo
    name: argo
```

eg, change [kfctl_gcp_iap.yaml](https://raw.githubusercontent.com/kubeflow/kubeflow/master/bootstrap/config/kfctl_gcp_iap.yaml)

kkasravi:
An alternative way to do vendor specific values is to have a kustomize generator plugin that could inject these values, however we would probably want to upgrade our kustomize version if we would call the plugin from within kfctl. I won't go into detail on this approach but plugins are described [here](https://github.com/kubernetes-sigs/kustomize/tree/master/docs/plugins)

IronPan:
Thanks @kkasravi this make sense. 
What would be the impact for no longer have images in kustomization.yaml
Also just wondering do we expect user to use kustomize without using kfctl or click-to-deploy app, to for example deploy to existing cluster?

jlewi:
@kkasravi That's a very elegant solution thanks.

swiftdiaries:
Thanks @kkasravi ! This is pretty neat.

@IronPan I believe since we are combining multiple overlays and we have a slight deviation from kustomize in kfctl, there's currently no good way to use `kubeflow/manifests` without kfctl or click to deploy.

kkasravi:
@IronPan the images tag in kustomization.yaml is how one should update images, however later versions of kustomize provide better ways to resolve vars that may allow us to keep images with a var in the image in kustomization.yaml. One improvement right now that's a PR in kustomize is 'automatic creation of vars and varReferences' (https://github.com/kubernetes-sigs/kustomize/pull/1217). This may not solve this problem but is a big improvement in kustomize usability.

kkasravi:
@IronPan 
> Also just wondering do we expect user to use kustomize without using kfctl or click-to-deploy app, to for example deploy to existing cluster?

we want to have each manifest target to be compatible with `kustomize build <target>` - short of var resolution and mixing in overlays (which come from the --config file).

Both var resolution and mixing in overlays can be done by a kustomize plugin. If we were to take part of kfctl that does the var and overlay logic and make it a kustomize plugin (go module), then kustoimize would load it automatically and we've enable the user to deploy/redeploy any manifest target individually - for example - different pipeline components could be individual redeployed.

jlewi:
@IronPan @kkasravi @swiftdiaries 

I responded too quickly and missed

> Note: the downside of the approach is we no longer have images in kustomization.yaml

I think this a pretty big downside. I think having a consistent and machine readable way to identify the images is pretty important. For example, this will make it much easier to automate mirror'ing all the images to some other repository.

kkasravi:
@jlewi 

> I think this a pretty big downside. I think having a consistent and machine readable way to identify the images is pretty important. For example, this will make it much easier to automate mirror'ing all the images to some other repository.

the images are resolved within each deployment yaml using a 'repository' var in the kustomization.yaml. EG deployment.yaml would have something like
```
    image: $(repository)/$(executorImage):$(executorVersion)
```

However each manifest target that had a deployment would require a params.env that contains
```
registry=foo
```

Jeremy - An alternative to this which might be more of what you had in mind is to have 'global' values in a top-level Application and 'registry' would be defined there. All deployments would still have $(registry) but it would be furnished by the Application or top-level Application.

jlewi:
@kkasravi Why not just rely on the images field in kustomization.yaml and build tooling around that? Your suggestion of using go plugin seems like a viable alternative.

Alternatively I could imagine baking this into `kustomize edit`. 

This was the script we were using before
https://github.com/kubeflow/kubeflow/blob/master/scripts/gke/use_gcr_for_all_images.sh

I was hoping that with kustomize it would be even easier to do this systematically.

Expecting all kustomize manifests to list images in kustomization.yaml seems pretty lightweight and gives us a lot of ability to do bulk operations.

kkasravi:
My apologies, this does work as expected using images in kustomization.yaml.
I've updated the [example](https://github.com/kubeflow/manifests/pull/252/files#). 
To review:

1. In deployment.yaml use 
```
        image: $(registry)/argoui:v2.3.0
```
2. In kustomization use the images tag and define a registry var
```
images:
  - name: $(registry)/argoui
    newName: $(registry)/argoui
    newTag: v2.3.0
  - name: $(registry)/workflow-controller
    newName: $(registry)/workflow-controller
    newTag: v2.3.0
...
vars:
- name: registry
  objref:
    kind: ConfigMap
    name: workflow-controller-parameters
    apiVersion: v1
  fieldref:
    fieldpath: data.registry
```

In params.env have the default value of registry be argoproj
but have the config file override the registry value with [kubeflowpipeline](https://github.com/kubeflow/kubeflow/pull/3814/files)
I've verified that the output of deployment.yaml has 
```
        image: kubeflowpipeline/argoui:v2.3.0
```


In order to get the same behavior as https://github.com/kubeflow/kubeflow/blob/master/scripts/gke/use_gcr_for_all_images.sh
we would need to add registry= to params.env in argo, ambassador, cloud-endpoint, katib, iap, metacontroller, pipeline* and override the value in the config file. 

We don't have a concept of a global var that can be used in all manifests right now.

@jlewi - a plugin could implement this behavior as well but we would be closer to pure kustomize using the above. I'll continue to look for a better solution that would allow a set of global vars to be applied to all manifests.

hemantha-kumara:
Based on above discussions, we have created a document [kfctl_registry.docx](https://github.com/kubeflow/manifests/files/3534512/kfctl_registry.docx) which summaries all options.
@jlewi @kkasravi @IronPan @swiftdiaries 

We(@SatwikBhandiwad @chethanmahadevaiah)  can submit a PR for **Option 2**: Registry path, image and tag can be taken from app.yaml in each component scope, pls do review and provide feedback

[KfDef.yaml](https://gist.github.com/hemantha-kumara/e5c387ed4230fee7d04026c346ca251a)   and related manifests will be like https://github.com/kubeflow/manifests/pull/252/files#

**Pros:**
1.	Only Manifests changes are required
2.	No changes are required in kfctl binary
3.	Existing kfdef files can work with this approach
4.	Multiple registries can be supported for kubeflow
5.	Single registry for each component

**Cons:**
1.	Defuncts Kustmoizations **images.newName**

PR #219, opened 2019-07-18
envoy-config: fix port name in setup_backend.sh

mattnworb:
This should fix https://github.com/kubeflow/kubeflow/issues/3628. See also https://github.com/kubeflow/kubeflow/pull/3691 (which modified the wrong file).

**Checklist:**
    1. `cd manifests/tests`
    2. `make generate`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/219)

### DIFF
--- a/gcp/iap-ingress/base/config-map.yaml
+++ b/gcp/iap-ingress/base/config-map.yaml
@@ -115,7 +115,7 @@ data:
       . /var/shared/healthz.env
 
       # If node port or backend id change, so does the JWT audience.
-      CURR_NODE_PORT=$(kubectl --namespace=${NAMESPACE} get svc ${SERVICE} -o jsonpath='{.spec.ports[0].nodePort}')
+      CURR_NODE_PORT=$(kubectl --namespace=${NAMESPACE} get svc ${SERVICE} -o jsonpath='{.spec.ports[?(@.name=="http2")].nodePort}')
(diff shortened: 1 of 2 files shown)


gcp/iap-ingress/base/config-map.yaml
```diff
@@ -115,7 +115,7 @@ data:
       . /var/shared/healthz.env
 
       # If node port or backend id change, so does the JWT audience.
-      CURR_NODE_PORT=$(kubectl --namespace=${NAMESPACE} get svc ${SERVICE} -o jsonpath='{.spec.ports[0].nodePort}')
+      CURR_NODE_PORT=$(kubectl --namespace=${NAMESPACE} get svc ${SERVICE} -o jsonpath='{.spec.ports[?(@.name=="'${PORT_NAME}'")].nodePort}')
```
lluunn: is PORT_NAME set? Should we just use http2 like other places:
https://github.com/kubeflow/manifests/blob/master/gcp/iap-ingress/base/config-map.yaml#L73

mattnworb: thanks, I didn't look closely enough at the structure of this file. updated

lluunn:
@mattnworb 
I think you need to cd tests, and do `make generate`

mattnworb:
@lluunn thanks for the help. I missed that originally as it seems that `hack/gen-test-target.sh` exits early if the repo is cloned to a directory named something other than `manifests`.

lluunn:
/lgtm
/approve

thanks for the help fixing this!

PR #235, opened 2019-07-22
Changes for workload identity

lluunn:
For https://github.com/kubeflow/kubeflow/issues/3638
- See diff with gcp in https://github.com/kubeflow/manifests/pull/226
- Pair with https://github.com/kubeflow/kubeflow/pull/3697

Basically, changes are:
- For workload related to ingress setup (backend-updater, iap-enabler etc), use k8s service account kf-admin, which is binded to the GCP admin service account.
- No more mounting gcp-admin-sa secret, and setting `GOOGLE_APPLICATION_CREDENTIALS`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/235)


jlewi:
Do all of the kustomize packages in in gcp need to change to support workload identity?

For some or all of the packages is it possible to use overlays to have workload identity and non-workload identity variants?

lluunn:
close for #238

PR #246, opened 2019-07-29
 Use managed cert

lluunn:
https://github.com/kubeflow/kubeflow/issues/3079

Make certmanager an overlay, and add managed cert as another overlay

**Checklist:**
    1. `cd manifests/tests`
    2. `make generate`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/246)


jlewi:
/lgtm
/approve
/hold

@lluunn This PR looks good to me but  do we also need to change the KFDef specs e.g.
https://github.com/kubeflow/kubeflow/blob/master/bootstrap/config/kfctl_gcp_iap.yaml
To use the new overlay? If so how do we sequence the changes?

I put a hold on this PR because I wasn't sure if it should be merged right away or whether we need to sequence it. 

I think one option would be in kfctl_gcp_iap.yaml if we pin to particular commits of kubeflow/manifests then its easier to make changes to kubeflow/manifests without worrying about breaking the configs on master.

lluunn:
It's already using the current certmanager one: https://github.com/kubeflow/kubeflow/blob/master/bootstrap/config/kfctl_gcp_iap.yaml#L287

PR #251, opened 2019-08-01
Combine the katib applications

wuchunghsuan:
Combine the katib applications into only two: katib-operators and katib-crds.

**Checklist:**
    1. `cd manifests/tests`
    2. `make generate`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/251)


johnugeorge:
@richardsliu Shouldn't we need make changes to  config files too https://github.com/kubeflow/kubeflow/blob/master/bootstrap/config/kfctl_k8s_istio.yaml  ?

richardsliu:
This is going to break deployment, we should first pin the kustomize version on kubeflow to a specific commit hash. See https://github.com/kubeflow/kubeflow/pull/3474/files for an example.

johnugeorge:
I feel that we can remove v1alpha1 now. It is not supported or maintained now anyways.  

@richardsliu @gaocegege What do you think?

richardsliu:
@johnugeorge We can do that in a separate PR.

richardsliu:
@wuchunghsuan @johnugeorge @gaocegege Instead of continuing this PR, I suggest we create the v1alpha3 manifests with just crd and operator packages. What do you think?

richardsliu:
I think this can be closed now.

PR #259, opened 2019-08-06
Add containerRuntimeExecutor field for other argo executors

Tomcli:
By default, Argo will use the `docker` containerRuntimeExecutor which it will not work on Kubernetes Cluster that doesn't use docker runtime. Therefore, we want to add the containerRuntimeExecutor field so that users who need different executor runtime can easily change the containerRuntimeExecutor params using kustomize. 

**Checklist:**
    1. `cd manifests/tests`
    2. `make generate`
    3. `make test`

cc: @animeshsingh

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/259)

### DIFF
--- a/argo/base/config-map.yaml
+++ b/argo/base/config-map.yaml
@@ -7,6 +7,7 @@ data:
   config: |
     {
     executorImage: $(executorImage),
+    containerRuntimeExecutor: $(containerRuntimeExecutor),
     artifactRepository:
     {
         s3: {
--- a/argo/base/kustomization.yaml
+++ b/argo/base/kustomization.yaml
@@ -30,6 +30,13 @@ vars:
     apiVersion: v1
   fieldref:
     fieldpath: data.executorImage
+- name: containerRuntimeExecutor
+  objref:
+    kind: ConfigMap
+    name: workflow-controller-parameters
+    apiVersion: v1
+  fieldref:
+    fieldpath: data.containerRuntimeExecutor
 - name: artifactRepositoryBucket
   objref:
     kind: ConfigMap
--- a/argo/base/params.env
+++ b/argo/base/params.env
@@ -1,5 +1,6 @@
 namespace=
 executorImage=argoproj/argoexec:v2.3.0
+containerRuntimeExecutor=docker
 artifactRepositoryBucket=mlpipeline
 artifactRepositoryKeyPrefix=artifacts
 artifactRepositoryEndpoint=minio-service.kubeflow:9000
--- a/tests/argo-base_test.go
+++ b/tests/argo-base_test.go
@@ -132,6 +132,7 @@ data:
   config: |
     {
     executorImage: $(executorImage),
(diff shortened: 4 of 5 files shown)


Tomcli:
Thanks @richardsliu, I updated the Argo tests and now the all the presubmit tests are passed.

swiftdiaries:
@Tomcli 
I didn't see this PR and I added this https://github.com/kubeflow/manifests/pull/266 to do the same via an overlay.
I think it'll be better to have an overlay so that a kfctl config can reference the overlay to install. I'm not sure how var substitution works otherwise at runtime from kfctl.
Do you think it makes sense to pull in those changes here? What are your thoughts?

swiftdiaries:
I used a config like this to init `kfctl` on the overlay
```
apiVersion: kfdef.apps.kubeflow.org/v1alpha1
kind: KfDef
metadata:
  name: kubeflow_app
  namespace: kubeflow
spec:
  repos:
  - name: manifests
    root: manifests-master
    uri: https://github.com/swiftdiaries/manifests/archive/containerd-cri-argo.tar.gz
  - name: kubeflow
    root: kubeflow-master
    uri: https://github.com/kubeflow/kubeflow/archive/master.tar.gz
...
```
And the containerd overlay for the argo application:
```
...
applications:
  - kustomizeConfig:
      overlays:
      - containerd
      repoRef:
        name: manifests
        path: argo
    name: argo
...
```

Tomcli:
Thanks @swiftdiaries, I agree having a containerd overlay is better to apply the changes via kfctl. My intent for this PR is to add the containerRuntimeExecutor variable to the base argo configmap so that other users can build their own custom overlays with the containerRuntimeExecutor variable as well.

swiftdiaries:
Oh cool, thanks for the clarification ! I can file a PR with the overlay in #266 after this one goes in.

Tomcli:
cc @animeshsingh

animeshsingh:
@kkasravi can we get this in? Its gating another PR, and is important for us at IBM to get Argo running with containerd

swiftdiaries:
I'll wait for the other reviewers as well but from my side this /lgtm

kkasravi:
@Tomcli will you update the config files to include this value
eg [kfctl_gcp_iap.yaml](https://github.com/kubeflow/kubeflow/blob/master/bootstrap/config/kfctl_gcp_iap.yaml)

```
  - kustomizeConfig:
      overlays:
      - istio
      parameters:
      - name: containerRuntimeExecutor
        value: docker
      repoRef:
        name: manifests
        path: argo
    name: argo
```

kkasravi:
other than the question about updating the config files which is in a different repository (kubeflow)
/lgtm

Tomcli:
Thanks @kkasravi for the review. I submitted a [PR to kubeflow](https://github.com/kubeflow/kubeflow/pull/3929) to update the corresponding config files.

PR #264, opened 2019-08-09
Tektoncd

kkasravi:
Prerequiste to kubeflow/testing#424

Adds tektoncd crds and install to manifests - similar to argo.

The tektoncd installation has the crds split out into a separate kustomize target
since kustomize v2.0.3 doesn't do ordering. The tektoncd manifest adheres to best practices 
and is similar to other kustomize targets where crds are separated.
 
**Checklist:**
    1. `cd manifests/tests`
    2. `make generate`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/264)


jlewi:
@kkasravi I don't think there is more work to get tekton installed in our test clusters; (i.e. lets not close  kubeflow/testing#424 after this PR is merged).

Test infra is here:
https://github.com/kubeflow/testing/tree/master/test-infra

So once we have a kustomize package for tekton we need to update that along with the instructions to actually get tekton deployed on our clusters.

kkasravi:
sure, i'll separate the others into a different PR

jlewi:
@kkasravi Is this PR ready for review? If it is could you update the PR description please to capture what this PR is doing? e.g. I don't think this PR is installing Tekton on our test clusters. I think this PR is just defining a kustomize manifest for installing tekton and possibly some tekton tasks and pipelines for Kubeflow.

kkasravi:
> @kkasravi Is this PR ready for review? If it is could you update the PR description please to capture what this PR is doing? e.g. I don't think this PR is installing Tekton on our test clusters. I think this PR is just defining a kustomize manifest for installing tekton and possibly some tekton tasks and pipelines for Kubeflow.

yes this is ready to merge. I did a rebase from main and a make generate;make test

jlewi:
/lgtm
/approve
/hold

@kkasravi Could you update the PR description to explain why you went with creating manifests for the individual components? I believe customize has a custom resource that can be used to install all of the Tekton components.

I don't think its worth changing right now; if this PR works then lets go with it. It would be good to document though why we went with this approach for future reference. i.e. if the reason is simply that we didn't know about the CR then lets just capture that in the PR description so that we know that it might be worth investigating later.

PR #266, opened 2019-08-10
adds containerRuntimeExecutor var for containerd overlay

swiftdiaries:
Resolves #

This PR is to reconcile changes in ksonnet that added a configmap variable for using alternative Container Runtimes like containerd in Argo.
It adds a kustomize var called `criExecutor` that defaults to docker but introduces an overlay for containerd.

The docker runtime in Argo is also unsecure as it mounts the docker socket directly onto each Pod as a volume mount. And is often rejected by OPA admission webhooks.

**Checklist:**
- [*] Unit tests have been rebuilt: 
    1. `cd manifests/tests`
    2. `make generate`
    3. `make test`

/cc @animeshsingh Tommy Li

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/266)

### DIFF
--- a/argo/overlays/containerd/params.env
+++ b/argo/overlays/containerd/params.env
@@ -1,0 +1,12 @@
+namespace=
+executorImage=argoproj/argoexec:v2.3.0
+containerRuntimeExecutor=pns
+artifactRepositoryBucket=mlpipeline
+artifactRepositoryKeyPrefix=artifacts
+artifactRepositoryEndpoint=minio-service.kubeflow:9000
+artifactRepositoryInsecure=true
+artifactRepositoryAccessKeySecretName=mlpipeline-minio-artifact
+artifactRepositoryAccessKeySecretKey=accesskey
+artifactRepositorySecretKeySecretName=mlpipeline-minio-artifact
+artifactRepositorySecretKeySecretKey=secretkey
+clusterDomain=cluster.local
--- a/tests/argo-overlays-containerd_test.go
+++ b/tests/argo-overlays-containerd_test.go
@@ -1,0 +1,505 @@
+package tests_test
+
+import (
+	"sigs.k8s.io/kustomize/k8sdeps/kunstruct"
+	"sigs.k8s.io/kustomize/k8sdeps/transformer"
+	"sigs.k8s.io/kustomize/pkg/fs"
+	"sigs.k8s.io/kustomize/pkg/loader"
+	"sigs.k8s.io/kustomize/pkg/resmap"
+	"sigs.k8s.io/kustomize/pkg/resource"
+	"sigs.k8s.io/kustomize/pkg/target"
+	"testing"
+)
+
+func writeArgoOverlaysContainerd(th *KustTestHarness) {
+	th.writeF("/manifests/argo/overlays/containerd/params.env", `
+namespace=
+executorImage=argoproj/argoexec:v2.3.0
+containerRuntimeExecutor=pns
+artifactRepositoryBucket=mlpipeline
+artifactRepositoryKeyPrefix=artifacts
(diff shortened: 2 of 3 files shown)


argo/base/params.env
```diff
@@ -1,4 +1,5 @@
 namespace=
+criExecutor=docker
```
Tomcli: I don't see you apply criExecutor variable to the argo configmap that actually read these variables at https://github.com/kubeflow/manifests/blob/master/argo/base/config-map.yaml

swiftdiaries: You're right, I forgot to add that

swiftdiaries: Thanks for the review, I'll add that in a commit

animeshsingh:
@Tomcli @swiftdiaries will this make Argo run on top of containerd environments? afaik we are using PNS executor at out end, rather than going through criExecutor?

Tomcli:
@swiftdiaries Looks like you also need to update the tests for argo and the overlays with the new argo config-map.yaml. You can refer to my PR #259 for the corresponding changes.

@animeshsingh Right now the kubeflow manifest doesn't give us the option to change different executor runtime like we have with ksonnet. This PR introduces a new overlay that can change the argo executor runtime when we deploy kubeflow. `criExecutor` is just a variable name in this PR, the actual variable name in the configmap is called `containerRuntimeExecutor`.

swiftdiaries:
@Tomcli Thanks! 
I've added the test, pre-submits are failing with https://github.com/kubeflow/kubeflow/issues/3883.
Will wait on the fix before retesting.

swiftdiaries:
@animeshsingh I am using the pns executor for Argo as well.
See: https://github.com/kubeflow/manifests/pull/266/files#diff-2a9f147588e9af81eb5224693fc80239R2

animeshsingh:
thanks - we should rename the var to something like containerExecutor.  crio is another mechanism in Kube community, and it will confuse

swiftdiaries:
I'll wait for the #259 to go in before reconciling those changes so that there's no divergence there

jlewi:
@animeshsingh @IronPan could you take another look so we could get this in for 0.7?

tests/argo-overlays-containerd_test.go
```diff
@@ -0,0 +1,505 @@
...
+	"sigs.k8s.io/kustomize/pkg/fs"
+	"sigs.k8s.io/kustomize/pkg/loader"
+	"sigs.k8s.io/kustomize/pkg/resmap"
+	"sigs.k8s.io/kustomize/pkg/resource"
+	"sigs.k8s.io/kustomize/pkg/target"
+	"testing"
+)
+
+func writeArgoOverlaysContainerd(th *KustTestHarness) {
+	th.writeF("/manifests/argo/overlays/containerd/params.env", `
```
Ark-kun: Would it be better to put these YAMLs into external test data files?

argo/overlays/containerd/params.env
```diff
@@ -0,0 +1,12 @@
+namespace=
+executorImage=argoproj/argoexec:v2.3.0
+containerRuntimeExecutor=pns
+artifactRepositoryBucket=mlpipeline
```
Ark-kun: Is it possible to have this part of configuration in a single place?

argo/overlays/containerd/params.env
```diff
@@ -0,0 +1,12 @@
+namespace=
+executorImage=argoproj/argoexec:v2.3.0
+containerRuntimeExecutor=pns
```
Ark-kun: Argo has 3 non-Docker executors.
Can you comment on their relative Pros and Cons? I do not disagree that PNS is probably the best one, but I lack information.

PR #274, opened 2019-08-14
parameterize userid header and prefix

### LINKED ISSUE #177: access-management-api: add userid-header and userid-prefix cli flags to parameters
Sub-issue for https://github.com/kubeflow/kubeflow/issues/3553.
Add parameters to configure:
* Which header we read for the user info (now it's only X-Goog-Authenticated-User-Email)
* What prefix is expected (now it's expected to be in the form of accounts.google.com:<user>, so the prefix is accounts.google.com:)

These are controlled by the cli flags `userid-header` and `userid-prefix`.

/priority p0

cc @jlewi @kunmingg

### LINKED ISSUE #178: central-dashboard: add userid-header and userid-prefix cli flags to parameters
Sub-issue for https://github.com/kubeflow/kubeflow/issues/3553.
Add parameters to configure:
* Which header we read for the user info (now it's only X-Goog-Authenticated-User-Email)
* What prefix is expected (now it's expected to be in the form of accounts.google.com:<user>, so the prefix is accounts.google.com:)

These are controlled by the cli flags `userid-header` and `userid-prefix`.

/priority p0

cc @jlewi

### LINKED ISSUE #179: access-management-ui: add userid-header and userid-prefix cli flags to parameters
Sub-issue for https://github.com/kubeflow/kubeflow/issues/3553.
Add parameters to configure:
* Which header we read for the user info (now it's only X-Goog-Authenticated-User-Email)
* What prefix is expected (now it's expected to be in the form of accounts.google.com:<user>, so the prefix is accounts.google.com:)

These are controlled by the cli flags `userid-header` and `userid-prefix`.

/priority p0

cc @jlewi @avdaredevil

### LINKED ISSUE #180: jupyter-web-app: add userid-header and userid-prefix cli flags to parameters
Sub-issue for https://github.com/kubeflow/kubeflow/issues/3553.
Add parameters to configure:
* Which header we read for the user info (now it's only X-Goog-Authenticated-User-Email)
* What prefix is expected (now it's expected to be in the form of accounts.google.com:<user>, so the prefix is accounts.google.com:)

These are controlled by the cli flags `userid-header` and `userid-prefix`.

/priority p0

cc @jlewi @kimwnasptd

### LINKED ISSUE #181: profile-controller: add userid-header and userid-prefix cli flags to parameters
Sub-issue for https://github.com/kubeflow/kubeflow/issues/3553.
Add parameters to configure:
* Which header we read for the user info (now it's only X-Goog-Authenticated-User-Email)
* What prefix is expected (now it's expected to be in the form of accounts.google.com:<user>, so the prefix is accounts.google.com:)

These are controlled by the cli flags `userid-header` and `userid-prefix`.

/priority p0

cc @jlewi @lluunn @kunmingg

yanniszark:
Resolves #177
Resolves #178
Resolves #179
Resolves #180
Resolves #181

Parameters to customize the userid header and prefix.
Essential for having multi-user capabilities in other-than-GCP platforms.

**Checklist:**
    1. `cd manifests/tests`
    2. `make generate`
    3. `make test`

/assign @kkasravi @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/274)

### DIFF
--- a/common/centraldashboard/base/kustomization.yaml
+++ b/common/centraldashboard/base/kustomization.yaml
@@ -35,6 +35,20 @@ vars:
     apiVersion: v1
   fieldref:
     fieldpath: data.clusterDomain
+- name: userid-header
+  objref:
+    kind: ConfigMap
+    name: parameters
+    apiVersion: v1
+  fieldref:
+    fieldpath: data.userid-header
+- name: userid-prefix
+  objref:
+    kind: ConfigMap
+    name: parameters
+    apiVersion: v1
+  fieldref:
+    fieldpath: data.userid-prefix
 configurations:
 - params.yaml
 
--- a/common/centraldashboard/base/deployment.yaml
+++ b/common/centraldashboard/base/deployment.yaml
@@ -21,4 +21,9 @@ spec:
         ports:
         - containerPort: 8082
           protocol: TCP
+        env:
+        - name: USERID_HEADER
+          value: $(userid-header)
+        - name: USERID_PREFIX
+          value: $(userid-prefix)
       serviceAccountName: centraldashboard
--- a/common/centraldashboard/base/params.env
+++ b/common/centraldashboard/base/params.env
@@ -1,1 +1,3 @@
 clusterDomain=cluster.local
+userid-header=
+userid-prefix=
--- a/common/centraldashboard/base/params.yaml
+++ b/common/centraldashboard/base/params.yaml
@@ -3,3 +3,7 @@ varReference:
   kind: Service
 - path: spec/http/route/destination/host
   kind: VirtualService
+- path: spec/template/spec/containers/0/env/0/value
+  kind: Deployment
+- path: spec/template/spec/containers/0/env/1/value
+  kind: Deployment
--- a/jupyter/jupyter-web-app/base/deployment.yaml
+++ b/jupyter/jupyter-web-app/base/deployment.yaml
@@ -18,6 +18,10 @@ spec:
             configMapKeyRef:
               name: parameters
               key: UI
+        - name: USERID_HEADER
+          value: $(userid-header)
+        - name: USERID_PREFIX
+          value: $(userid-prefix)
         image: gcr.io/kubeflow-images-public/jupyter-web-app:v0.5.0
         imagePullPolicy: $(policy)
         name: jupyter-web-app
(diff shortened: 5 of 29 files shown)


jlewi:
Thanks

So just to clarify. This is adding parameters for userid and prefix. The defaults for those will be the empty string.

If they are the empty string does that preserve existing behavior? i.e. GCP headers? Or do we need to update kfctl specs for GCP to start setting those headers correctly.

yanniszark:
> If they are the empty string does that preserve existing behavior? i.e. GCP headers?

GCP headers are hardcoded defaults in every application, so everything should still work.
However, GCP configs should edited to use these parameters and then these defaults should be removed from the respective applications.

jlewi:
@yanniszark seems good to me.  I might suggest opening issues to track.

/lgtm
/approve

 @kkasravi any comments?

common/centraldashboard/base/kustomization.yaml
```diff
@@ -35,6 +35,20 @@ vars:
     apiVersion: v1
   fieldref:
     fieldpath: data.clusterDomain
+- name: userid-header
+  objref:
+    kind: ConfigMap
+    name: parameters
```
kkasravi: i believe this will remain 'parameters' since no prefix is specified.
We should probably have an e2e that makes sure configmap names are unique per manifest.

yanniszark: Does this have any adverse effects right now?
Or can this PR go in?

jlewi:
Looks like there is an open thread with @kkasravi that needs to be resolved?

common/centraldashboard/base/kustomization.yaml
jlewi: IIUC that is independent of this PR. I don't believe this PR is adding the ConfigMap for central dashboard. And this manifest currently doesn't add a common prefix. So there's no guarantee the config map is going to be unique and it could conflict with configmaps from other applications. But I think that's already the case and unrelated to this PR.

Should we open an issue to track that and fix later?

yanniszark: @jlewi that is correct, the CentralDashboard ConfigMap isn't introduced by this PR.
I opened https://github.com/kubeflow/manifests/issues/288 to track the issue.

yanniszark:
@jlewi I opened https://github.com/kubeflow/manifests/issues/288 to track the issue @kkasravi mentioned.
I also rebased to get the tests passing.
If everything is ok, can I get another lgtm to this?

jlewi:
@yanniszark I think you just need to remove the hold.

PR #298, opened 2019-08-21
[CentralDashboard v2] Added env_var for KFAM Server address

avdaredevil:
## About
Added an environment variable to central dash to target the correct KFAM target

### Meta
/area centraldashboard
/area front-end
/priority p0
/assign @avdaredevil
/cc @prodonjs @kunmingg @abhi-g

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/298)

### DIFF
--- a/common/centraldashboard/base/deployment.yaml
+++ b/common/centraldashboard/base/deployment.yaml
@@ -27,5 +27,5 @@ spec:
         - name: USERID_PREFIX
           value: $(userid-prefix)
         - name: PROFILES_KFAM_SERVICE_HOST
-          value: profiles-kfam
+          value: profiles-kfam.kubeflow
       serviceAccountName: centraldashboard
--- a/common/centraldashboard/base/deployment.yaml
+++ b/common/centraldashboard/base/deployment.yaml
@@ -26,4 +26,6 @@ spec:
           value: $(userid-header)
         - name: USERID_PREFIX
           value: $(userid-prefix)
+        - name: PROFILES_KFAM_SERVICE_HOST
+          value: profiles-kfam
       serviceAccountName: centraldashboard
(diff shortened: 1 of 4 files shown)


common/centraldashboard/base/deployment.yaml
```diff
@@ -26,4 +26,6 @@ spec:
           value: $(userid-header)
         - name: USERID_PREFIX
           value: $(userid-prefix)
+        - name: PROFILES_KFAM_SERVICE_HOST
+          value: profiles-kfam
```
kunmingg: Can we have the default value point to service address?
`kfam.kubeflow:8081`

kunmingg:
/lgtm
/approve
Seems we only need service name

abhi-g:
you might have to update tests. 
go to the tests dir and 'make generate' followed by 'make test'. Then update the PR

common/centraldashboard/base/deployment.yaml
jlewi: +1

prodonjs: @avdaredevil - Why not change the default value in [server.ts](https://github.com/kubeflow/kubeflow/blob/840e24b5e86654d8e4a3596a3c91ffb2d198bca1/components/centraldashboard/app/server.ts#L12) so that it expects to resolve the `profiles-kfam` DNS name and has to overridden only when developing locally?

That would alleviate the need for this to be set in manifests.

avdaredevil: That's already tackled in kubeflow/kubeflow#3971 @prodonjs 

Sure @kunmingg @jlewi

avdaredevil: Port is a different argument and is already defaulted to 8081

Issue #296, opened 2019-08-21
Unit tests in v0.6-branch are not up-to-date and updating them makes tests fail


yanniszark:
Steps to reproduce:
* Checkout to v0.6-branch (eab125dd42c08b0e2495b7b2d7a7010ac64d3774).
* Regenerate tests: 
```bash
cd tests
make generate
```

Immediately, you will notice changes.
This means that the tests were out of sync, but tests passed either way.

* Run unit tests:

```bash
make test
```

Tests fail:
```
=== RUN   TestIstioBase                                                                                                                                                                                     
--- FAIL: TestIstioBase (0.01s)                                                                                                                                                                             
    istio-base_test.go:157: Err: cannot read file "/manifests/istio/istio/base/params.yaml"                                                                                                                 
```
This essentially means that tests on v0.6-branch are broken.

@jlewi @kkasravi why didn't CI catch this?

/priority p0

kkasravi:
i believe this may be due to not rebasing from master.

I think we add this to the PR template form

1. git fetch upstream master
2. git rebase -i upstream/master
3. cd tests
4. make generate
5. make test

yanniszark:
@kkasravi this is not master though.
This is v0.6-branch.
Shouldn't the E2E tests catch this ?

kkasravi:
yes - we need to update the e2e to do the following

mv tests tests_check
make generate
see if any *.go files differ between tests tests_check
if yes 
  fail
else
  make test
fi

kkasravi:
@gabrielwen you'll need to update run-tests.sh along the lines of what I [suggested](https://github.com/kubeflow/manifests/issues/296#issuecomment-523597616)

kkasravi:
Adding liuhougang6@126.com and adhita.selvaraj@gmail.com

I think this means that something was merged w/o doing a generate.

There is a flaw in the template we use for the PR, any PR should do the following
git fetch upstream master
git rebase -i upstream/master
cd tests
make generate
make test
Yannis do you know how to update the PR template?

swiftdiaries:
The test workflow seems to be running this [run-tests.sh](https://github.com/kubeflow/manifests/blob/master/tests/scripts/run-tests.sh)
But make generate is not part of that, is that intended ?

jlewi:
@kkasravi Are you saying that if PR authors don't run appropriate commands then our presubmits won't catch test failures?

kkasravi:
[...]

PR #299, opened 2019-08-21
adds make generate to the test workflow

swiftdiaries:
Resolves #
https://github.com/kubeflow/manifests/issues/296

Adds `make generate` as part of the test workflow after checkout of the manifests repo.

**Checklist:**
    1. `cd manifests/tests`
    2. `make generate`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/299)

### DIFF
--- a/tests/scripts/run-tests.sh
+++ b/tests/scripts/run-tests.sh
@@ -20,10 +20,19 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-cd tests
-
 # the tests depend on kustomize
 export PATH=${GOPATH}/bin:/usr/local/go/bin:${PATH}
 export GO111MODULE=on
 go get sigs.k8s.io/kustomize
-make test
+
+# Generate Go files on upstream HEAD 
+cd /src/kubeflow/manifests/tests
+make generate
+cd -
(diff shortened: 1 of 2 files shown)


tests/scripts/run-tests.sh
```diff
@@ -20,10 +20,18 @@ set -o errexit
...
+
+# Generate Go files on upstream HEAD 
+cd /src/kubeflow/manifests/tests
+make generate
+cd -
+if diff /src/kubeflow/manifests/tests/*.go tests/*.go ; then
+    cd tests
+    make test
+else
+    exit 1
```
yanniszark: Should we print a message here to help users who encounter this?

swiftdiaries: Thanks for the review @yanniszark, I've added a message to be printed before exiting. 
PTAL

jlewi:
It looks like this PR broke the presubmits on the v0.6-branch.
See #348 

It looks like there are two problems

1. EXTRA_REPOS is not correct; repos should be separated by a semi-colon not a common.
1. The changes to run_tests.sh don't work when run under the CI system.

I'm going to revert this change.

Issue #288, opened 2019-08-20
Multiple packages may generate the same parameters ConfigMap resource


yanniszark:
A pattern we use in kubeflow/manifests is defining a ConfigMap and using its values as variables, to achieve variable substitution.

In addition, there is no CI test to check if two packages in manifests generate the same ConfigMap resource (or any resource for that matter).

Normally, it is rare to have collisions between resources created in different packages.
However, for the parameters ConfigMap, this can easily happen as:
* There is a tendency to disable the ConfigMap name suffix.
* All packages use the same name for the parameters ConfigMap (copy-pasting probably).
* Some packages don't use a namePrefix, which would solve the problem.

Effects:
* `kustomize build` should still work as expected, as the variables are substituted from the ConfigMap generated at that moment, before it is applied to K8s.
* Our strategy now is create if not exists, so only the first ConfigMap of the same name will be applied. If the strategy changes to apply, then the last ConfigMap will be the one to exist in the end.
* In case a component uses a ConfigMap reference instead of a variable substitution, this could potentially break things (eg https://github.com/kubeflow/manifests/blob/8c5b9cee9c637fe6102603fdebfac569c178112e/jupyter/jupyter-web-app/base/deployment.yaml#L14)

Possible mitigations:
* Enforce namePrefix for all components.
* Add CI test.

/cc @kimwnasptd  @kkasravi @jlewi 

/priority p1

kkasravi:
thanks for opening this @yanniszark !

Issue #256, opened 2019-08-06
ConfigMap named "ui-parameters" generated by both pipelines-ui and metadata


mattnworb:
I am trying to install 0.6.1 for the first time and I noticed that two different components are trying to generate configmaps with the base name of `ui-parameters`:

https://github.com/kubeflow/manifests/blob/7cf5587dd9e2ec523bbdc99642d530f42fec4361/metadata/base/kustomization.yaml#L6-L8

https://github.com/kubeflow/manifests/blob/e2e7d42c4a95b374da674980542b019c7ed6fcad/pipeline/pipelines-ui/base/kustomization.yaml#L10-L12

I am not incredibly familiar with kustomize, but will this cause any problems? Neither file has `disableNameSuffixHash`, so both will probably generate uniquely named resources, but seems like it could lead to some confusion as well.

Issue #147, opened 2019-06-18
centraldashboard should provide a nameprefix


kkasravi:
centraldashboard creates a configMapGenerator with the name parameters. 
It should provide a nameprefix so the name is unique

Issue #306, opened 2019-08-23
How do we manually verify that the expected value in the unittests is correct?


jlewi:
How are we supposed to verify that the expected values used by unittests is correct?

As an example:  kubeflow/kubeflow#3986 appears to be due to a bug in the kustomize manifest resulting in environment variables not being set correctly to the variable value; instead they remained set to the variable name e.g. `$(userid-prefix)

I would expect this is something we test in the unittest. 

So my expectation would be that I can look at the unittest and verify by inspection that the expected value is what it should be.

But when I look at the unittest
https://github.com/kubeflow/manifests/blob/76bd875d65d39960ab2d75daa2ba2b7a2a9e82a4/tests/centraldashboard-base_test.go#L76

It looks like we are specifying the kustomize files (i.e. the input) and not the expected output.
For example here:
https://github.com/kubeflow/manifests/blob/76bd875d65d39960ab2d75daa2ba2b7a2a9e82a4/tests/jupyter-web-app-base_test.go#L242

We are setting

```
        image: gcr.io/kubeflow-images-public/jupyter-web-app:v0.5.0
        imagePullPolicy: $(policy)
```

The expected value for imagePullPolicy should not be $(policy). So either that is the input value and not the expected value or else our tests are wrong.

Can someone familiar with the unittest infrastructure help educate me?

My expectation is that unittests would work as follows

* A golden set of manifests for different scenarios would be checked in somewhere (either as go code or as separate files in a "golden") directory
   * Thus I can look at those golden files to see if the expected value is what it should be
* The tests would be comparing the generated files to the golden files.
* There might be some scripts or utilities to regenerate the golden files (e.g. by running kustomize)
    * During code review it should be easy to inspect the diff to the golden files to allow manual verification of the new golden data

/cc @johnugeorge @yanniszark @kkasravi

jlewi:
What is this code doing?
https://github.com/kubeflow/manifests/blob/76bd875d65d39960ab2d75daa2ba2b7a2a9e82a4/tests/jupyter-web-app-base_test.go#L451

Looks like we are running kustomize to get the expected value.

kkasravi:
if values aren't replaced and one or more warnings are emitted 
```
2019/08/23 12:29:59 well-defined vars that were never replaced: clusterDomain
```
we should FAIL on that unit test.

kkasravi:
@jlewi I've opened #308 PTAL

jlewi:
thanks @kkasravi 

[...]

PR #331, opened 2019-08-29
To support uniform and unique configmaps names across kubeflow applications

chethanmahadevaiah:
Resolves #
Currently multiple application creates config maps with the same generic name "parameters".
Changes planned here is to have a unique configmaps for each application in an uniform way using the namePrefix.
Similarly for services too.

https://github.com/kubeflow/manifests/issues/288
https://github.com/kubeflow/manifests/issues/256
https://github.com/kubeflow/manifests/issues/147

To support unique configmaps and service names
The pattern followed is inline with kustomize ability to prefix names.

Specific example for where its used like one below where namePrefix is used
https://github.com/kubeflow/manifests/blob/02ddd6616650642b2ff865f16b164ede3385c54b/jupyter/jupyter-web-app/base/kustomization.yaml#L12
and the config map is named
https://github.com/kubeflow/manifests/blob/02ddd6616650642b2ff865f16b164ede3385c54b/jupyter/jupyter-web-app/base/kustomization.yaml#L23

Not using namePrefix but hardcoding of the name as below and does not have any namePrefix:
https://github.com/kubeflow/manifests/blob/02ddd6616650642b2ff865f16b164ede3385c54b/argo/base/kustomization.yaml#L21

Tensorboard example where namePrefix is not present and configMap name is just “parameters”:
https://github.com/kubeflow/manifests/blob/02ddd6616650642b2ff865f16b164ede3385c54b/tensorboard/base/kustomization.yaml#L10

Multiple applications are using either “parameters” or “config-map” as name for configMap with and without namePrefix. The PR makes it uniform to have “parameters” consistently across all applications.

configmaps name: "<namePrefix>-parameters"
service name: "<namePrefix>-service"

Disabled name suffix Hash where ever missing.
generatorOptions:
  disableNameSuffixHash: true

**Checklist:**
    1. `cd manifests/tests`
    2. `make generate`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/331)

### DIFF
--- a/metadata/base/kustomization.yaml
+++ b/metadata/base/kustomization.yaml
@@ -1,11 +1,13 @@
-namePrefix: metadata-
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 commonLabels:
   kustomize.component: metadata
 configMapGenerator:
-- name: ui-parameters
+- name: parameters
   env: params.env
+generatorOptions:
+  disableNameSuffixHash: true
+namePrefix: metadata-
 resources:
 - metadata-db-pvc.yaml
 - metadata-db-secret.yaml
@@ -23,21 +25,21 @@ vars:
 - name: ui-namespace
   objref:
     kind: Service
-    name: ui
+    name: service
     apiVersion: v1
   fieldref:
     fieldpath: metadata.namespace
 - name: ui-clusterDomain
   objref:
     kind: ConfigMap
-    name: ui-parameters
+    name: parameters
     version: v1
   fieldref:
     fieldpath: data.uiClusterDomain
 - name: service
   objref:
     kind: Service
-    name: ui
+    name: service
     apiVersion: v1
   fieldref:
     fieldpath: metadata.name
--- a/profiles/base/kustomization.yaml
+++ b/profiles/base/kustomization.yaml
@@ -12,8 +12,10 @@ namePrefix: profiles-
 namespace: kubeflow
 commonLabels:
   kustomize.component: profiles
+generatorOptions:
+  disableNameSuffixHash: true
 configMapGenerator:
-  - name: profiles-parameters
+  - name: parameters
     env: params.env
 images:
   - name: gcr.io/kubeflow-images-public/profile-controller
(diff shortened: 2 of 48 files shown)


chethanmahadevaiah:
@googlebot I signed it!

metadata/base/kustomization.yaml
```diff
@@ -1,11 +1,13 @@
-namePrefix: metadata-
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 commonLabels:
   kustomize.component: metadata
 configMapGenerator:
-- name: ui-parameters
+- name: parameters
```
nrchakradhar: Since there are multiple params here, probably the name ui- still makes sense.

profiles/base/kustomization.yaml
```diff
@@ -23,28 +25,28 @@ vars:
...
-      name: profiles-parameters
+      name: parameters
       apiVersion: v1
     fieldref:
       fieldpath: data.userid-prefix
   - name: namespace
     objref:
       kind: Service
-      name: kfam
+      name: service
```
nrchakradhar: Here also for better clarity kfam can still be present right?

jlewi:
@chethanmahadevaiah Can you update the PR description please? Can you explain what the problem is and how this PR fixes it? 

One thing I'm confused about is if we want to generate unique names for the config maps why don't we use kustomize's ability to attach a common prefix like we do for other resources to make those resource names unique?

It looks like one problem we run into when doing that is that the prefix isn't automatically added in places where we are doing substitution. Is that accurate? is there anyway to fix that?

chethanmahadevaiah:
> @chethanmahadevaiah Can you update the PR description please? Can you explain what the problem is and how this PR fixes it?

@jlewi I have updated the description with the details.

> One thing I'm confused about is if we want to generate unique names for the config maps why don't we use kustomize's ability to attach a common prefix like we do for other resources to make those resource names unique?

 Yes, we are utilizing the kustomize's prefix option here. As it was done only for few of the application, we extended it for all the default applications.  Also disabled the disableNameSuffixHash. This is not only unique but also maintains the uniformity across the components.

> It looks like one problem we run into when doing that is that the prefix isn't automatically added in places where we are doing substitution. Is that accurate? is there anyway to fix that?

Yes this we verified and saw it was added to all the applications as expected.

jlewi:
Thanks @chethanmahadevaiah 

It looks like the tests didn't run. Its possible once the tests need to be run they will need to be generated. If the tests need to be regenerated then that suggests the output of kustomize build is changing.

Due to #306 its not easy to verify what is actually changing in the outputted manifests which make it hard to review.

If the output ends up changing should we come up with some fix to #306 before checking in this PR?

nrchakradhar:
@kkasravi, can you please suggest how to make the names unique? As per @jweli, nameprefix is desired. Do we go with namesuffix then.

jlewi:
@chethanmahadevaiah @nrchakradhar @kkasravi What is the path forward with this PR?

PR #336, opened 2019-09-02
Remove katib-v1alpha1

hougangliu:
Katib has removed v1alpha1, and katib v1alpha2 and the successors will not change a lot.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/336)


johnugeorge:
As we will have one more alpha release, should we skip renaming katib-v1alpha2 to katib?

hougangliu:
> As we will have one more alpha release, should we skip renaming katib-v1alpha2 to katib?

We had remained both katib-v1alpha1 and katib-v1alpha2, because the two versions have much difference. After v1alpha2, the change will not be big. we should remove version information in directory, keeping consistent with other Kubeflow components

johnugeorge:
@hougangliu I remember, two separate directories are added due to limitation in parsing logic in manifests. @richardsliu Is this fixed?

kkasravi:
@hougangliu could you rebase ?
`git fetch upstream master`
`cd tests && make generate && make test && git commit -a -m "rebase from upstream/master" && git push origin remove-katib-v1 -F`

johnugeorge:
This PR is obsolete now

PR #339, opened 2019-09-04
Parameterize container image name and image tag to read from kfDef fi…

chethanmahadevaiah:
Resolves #
https://github.com/kubeflow/manifests/pull/211
https://github.com/kubeflow/kubeflow/issues/3210 
https://github.com/kubeflow/manifests/pull/211#issuecomment-524304213

Resolves the issue with the hardcoded values of container registry, name and tag details of all the kubeflow applications.

Changes will allow us to parameterize the container Image Registry, Image Name and Image Tag for each application.
Values for each of these can be provided from the kfDef file (app.yaml)
If not present in the kfDef file then default value present in the params.env willbe considered.

**e.g: app.yaml:**
  - kustomizeConfig:
      parameters:
      - name: centraldashboardImageName
        value: gcr.io/kubeflow-images-public/centraldashboard
      - name: centraldashboardImageTag
        value: v20190823-v0.6.0-rc.0-69-gcb7dab59
      overlays:
      - istio
      repoRef:
        name: manifests
        path: common/centraldashboard
    name: centraldashboard

*********************
This does same job as and de-funcs newImage and newTag options available from kustomization.yaml at every base/ 

This is sample changes done for centraldashboard app, which needs to be replicated for all the applications to achieve these customization/parameterization.

**Checklist:**
    1. `cd manifests/tests`
    2. `make generate`
    3. `make test`

/assign @kkasravi @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/339)

### DIFF
--- a/common/centraldashboard/base/deployment.yaml
+++ b/common/centraldashboard/base/deployment.yaml
@@ -15,7 +15,7 @@ spec:
         app: centraldashboard
     spec:
       containers:
-      - image: gcr.io/kubeflow-images-public/centraldashboard:v0.5.0
+      - image: $(registry)/$(centraldashboardImageName):$(centraldashboardImageTag)
         imagePullPolicy: IfNotPresent
         name: centraldashboard
         ports:
--- a/common/centraldashboard/base/kustomization.yaml
+++ b/common/centraldashboard/base/kustomization.yaml
@@ -28,6 +28,27 @@ vars:
     apiVersion: v1
   fieldref:
     fieldpath: metadata.namespace
+- name: registry
+  objref:
+    kind: ConfigMap
+    name: parameters
+    apiVersion: v1
+  fieldref:
+    fieldpath: metadata.registry
+- name: centraldashboardImageName
+  objref:
+    kind: ConfigMap
+    name: parameters
+    apiVersion: v1
+  fieldref:
+    fieldpath: metadata.centraldashboardImageName
+- name: centraldashboardImageTag
+  objref:
+    kind: ConfigMap
+    name: parameters
+    apiVersion: v1
+  fieldref:
+    fieldpath: metadata.centraldashboardImageTag
 - name: clusterDomain
   objref:
     kind: ConfigMap
--- a/common/centraldashboard/base/params.env
+++ b/common/centraldashboard/base/params.env
@@ -1,3 +1,6 @@
 clusterDomain=cluster.local
 userid-header=
 userid-prefix=
+registry=gcr.io
+centraldashboardImageName=kubeflow-images-public/centraldashboard
+centraldashboardImageTag=v20190823-v0.6.0-rc.0-69-gcb7dab59
--- a/common/centraldashboard/base/params.yaml
+++ b/common/centraldashboard/base/params.yaml
@@ -7,3 +7,5 @@ varReference:
   kind: Deployment
 - path: spec/template/spec/containers/0/env/1/value
   kind: Deployment
+- path: spec/template/spec/containers/image
+  kind: Deployment


jlewi:
Do we need to create separate parameters for the registry?

kustomize edit set image makes it easy to rewrite the registry and the image tag.
See: kubeflow/kubeflow#3210

The current thinking is that if images are listed in kustomization.yaml files then its easy to

1. Process all kustomization.yaml files to get a list of all images
1. Do bulk transformations of the images (e.g. bulk substitute the registry for a new one).

hemantha-kumara:
@jlewi instead of changing each component container registries(using kustomize edit set), we can change it in manifests(one time effort) of all components params to take registry and tags.
registry value can be global or from component level from kfdef.
both options pros and cons are mentioned in https://docs.google.com/document/d/15_4ATuUDqLNNJF5_nv5uOliguUMWPLHXcK9mB8HqrPw/edit 
based on your view, it can be submitted.

krishnadurai:
@hemantha-kumara @jlewi if we choose to kustomize edit set image, should we allow for kfctl to set the registry as opposed to a script?

jlewi:
@hemantha-kumara It looks like you are going with option 2 in your doc.

Can you explain why you want to define a new pattern for doing image replacement rather than relying on the standard estabilished by kustomize for doing image replacement? 

The approach in this PR has some big disadvantages

1. I believe it breaks the ability of kustomize to do image replacement
1. Every kustomize application has to adopt this convention in order to ensure consistency? How are we going to ensure this overtime?

With kustomize it should be pretty straightforward to 

1. Recourse over the kustomization.yaml files
1. Read the YAML files to get the images
1. Do kustomize edit set image 

We should be able to script this fairly easily (see kubeflow/kfctl#30) and add whatever semantics we want in term of matching and replacing the registry.

This has a major advantage of not imposing any new patterns/conventions on our kustomize applications and instead relying on core kustomize patterns/functionality.

@krishnadurai 

> if we choose to kustomize edit set image, should we allow for kfctl to set the registry as opposed to a script?

Yes. I think though we should start by making it a separate binary or subcommand until we have the semantics right.

hemantha-kumara:
@jlewi  Thanks for your comments. I will look and get back. 
I will update the [doc](https://docs.google.com/document/d/15_4ATuUDqLNNJF5_nv5uOliguUMWPLHXcK9mB8HqrPw/edit) with comments.

hemantha-kumara:
@jlewi @krishnadurai @xaniasd
> Can you explain why you want to define a new pattern for doing image replacement rather than relying on the standard estabilished by kustomize for doing image replacement?

Initialy we were thinking plugin in kfctl takes some time and also not to change anything in kfctl, hence proposed changes in manifests(option-2).

> We should be able to script this fairly easily (see kubeflow/kfctl#30) and add whatever semantics we want in term of matching and replacing the registry.

With the new kfdef specification based on [kfctl design doc](https://docs.google.com/document/d/1kviNOdjoBPeoJ56KBvMKWq7SzrSN5rFUQoULK8muD-M/edit#), script/binary changes can it be added as plugin? or can it be added as sub-command?
New plugin can be called at either 
- End of build/generate step
- Starting of apply step

Pls note that with the existing design doc, plugins are called during **apply** step only. 

```
--- 
apiVersion: kubeflow.org/v1alpha1
kind: KfDef
spec: 
  plugins: 
    - 
      name: registryUpdater
      spec: 
        applications: 
          - 
            application: 
              images: 
                - 
                  name: argoproj/argoui
                  newName: argoproj/argoui
                  newTag: v2.3.0
                - 
                  name: argoproj/workflow-controller
                  newName: argoproj/workflow-controller
                  newTag: v2.3.0
              name: argo
          - 
            application: 
              images: 
                name: gcr.io/kubeflow-images-public/kubernetes-sigs/application
                newName: gcr.io/kubeflow-images-public/kubernetes-sigs/application
                newTag: '1.0-beta'
              name: application
        globalRegistry: test.io 
```
If globalRegistry is NOT blank AND
1. No applications are defined inside spec.applications - then value of globalRegistry will be added to the beginning of newName of all kustomize files
2. Some/All applications are defined inside spec.applications - then value of globalRegistry will be replaced/added to the beginning of newName of all kustomize files

If globalRegistry is blank AND
1. No applications are defined inside spec.applications - then no changes from plugin
2. Some/All applications are defined inside spec.applications - then value of newName, newTag  will be replaced in defined applications kustomize files. In this case each application can use different registries.

Incase of sub-command, it can be executed after build/generate and before apply

I will update doc based on feedback on this.

jlewi:
What is the next step with this PR?

kubeflow/kubeflow#3210 is the uber issue tracking how we want to go about doing bulk replacement of Kubeflow images with images in a different registry.

PR #352, opened 2019-09-10
Self signed certs for dex ingress endpoints as a k8s job

krishnadurai:
Addresses partially: kubeflow/kfctl#6

This PR contains a K8s job to generate self-signed certificates for Dex based ingress endpoints.
The scope of this PR doesn't include waiting for istio-ingress-gateway service's load-balancer IP to add it to 'domains' for this k8s job. This can be configured separately.

Run `kustomize build dex-auth/dex-ingress | kubectl apply -f -` to test.

**Params**
istioNamespace: Istio installation's namespace
domains: A comma-separated list of domains to alt-names

Note: We should adopt a cert-manager implementation to include LetsEnrypt based certs in the future.

**Checklist:**
    1. `cd manifests/tests`
    2. `make generate`
    3. `make test`

/assign @krishnadurai 
/cc @jlewi @yanniszark

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/352)

### DIFF
--- a/dex-auth/dex-ingress/base/Dockerfile
+++ b/dex-auth/dex-ingress/base/Dockerfile
@@ -1,0 +1,18 @@
+FROM alpine
+
+MAINTAINER Krishna Durai <krishadurai20@gmail.com>
+
+ENV KUBE_LATEST_VERSION="v1.15.0"
+
+RUN apk add --update ca-certificates \
+ && apk add --update -t deps curl \
+ && apk add --no-cache bash \
+ && curl -L https://storage.googleapis.com/kubernetes-release/release/${KUBE_LATEST_VERSION}/bin/linux/amd64/kubectl -o /usr/local/bin/kubectl \
+ && chmod +x /usr/local/bin/kubectl \
+ && apk add --no-cache openssl \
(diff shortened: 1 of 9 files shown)


yanniszark:
@krishnadurai any reason not to use cert-manager for this?

krishnadurai:
@yanniszark This is an initial implementation, there is no reason as such right now - other than limiting the scope. We should go ahead and use cert-manager along with LetsEnrypt in the near future. What I'm thinking is that we could use managed certificates along with cert-manager - then that would require some discussion around the tool which we can use for managed certificates.

krishnadurai:
This follows from kubeflow/kubeflow#3739

yanniszark:
@krishnadurai cert-manager can be used to generate self-signed certs.
My thinking is that we should standardize on using it  instead of writing custom code for every component that might need a certificate (eg webhook, ingress, etc).
It is less error-prone, reusable and battle-tested.

krishnadurai:
@yanniszark I agree with your thoughts on standardizing cert-manager. I will close this PR and replace this with a cert-manager implementation.

/close

PR #358, opened 2019-09-13
fix for backtick escape error

### LINKED ISSUE #356: unit test generation isn't properly escaping embedded backticks
the generator for go unit tests embeds the code in the unit test using the golang multi-line backtick. When embedding code that has backticks, the generator tries to use a unicode sequence. However the correct approach is to surround the backtick with double-quotes.
See https://stackoverflow.com/questions/21198980/golang-how-to-escape-back-ticks

kkasravi:
Resolves #356

backticks are escaped as described in the issue.
a second fix is related to returning the earliest 'manifests' in a directory path rather than the latest. If a subdir ends with manifests you will get an error in unit test generation.

**Checklist:**
    1. `cd manifests/tests`
    2. `make generate`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/358)

### DIFF
--- a/hack/gen-test-target.sh
+++ b/hack/gen-test-target.sh
@@ -19,8 +19,9 @@ kebab-case-2-PascalCase() {
 
 gen-target-start() {
   local dir=$(get-target $1) target fname
-  fname=/manifests${dir##*/manifests}
+  fname=/manifests${dir#*/manifests}
   target=$(kebab-case-2-PascalCase $(get-target-name $1))
+#echo 'gen-target-start dir='$dir' fname='$fname' target='$target
 
   echo 'package tests_test'
   echo ''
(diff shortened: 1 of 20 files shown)


jlewi:
@kkasravi Why use escaping at all? i.e. rather than embed YAML in go files; why not just create YAML files and have the unittests load those files? If this is a quick fix then doing that then I think its fine.

Longer term though it seems like #306 (verifying expected values) is pushing us in the direction of creating a test_data directory that would contain YAML files representing the expected output of kustomize build.

So storing YAML files as opposed to embedding YAML in go files seems more consistent with that.

That said, there's probably a lot more changes needed to realize #306 and I'm not sure that changing the current scripts to emit the kustomize files to YAML rather than embedding in the go code really moves the needle. So I'm fine with the workaround in this PR.

hougangliu:
please resolve the conflict. lgtm

kkasravi:
thanks @jlewi - it would be more straight forward to store the files in a different directory and read them in.

PR #369, opened 2019-09-19
Upgrades cert-manager to v0.11.0 - was an upgrade to v0.10.0 earlier

### LINKED ISSUE #365: Upgrade cert-manager to v0.10.0
The current version of cert-manager is at v0.4.0 and doesn't support flags like 'isCA' for Issuer resource.

Refer PR: #362 

/assign @krishnadurai 
/cc @yanniszark

krishnadurai:
~Upgrades cert-manager to v0.10.0~
Upgrades cert-manager to v0.11.0
Structures ClusterIssuers into 2 overlays:
- letsencrypt
- self-signed
As discussed in #362. Unblocks #362.

Resolves #365 

The current version of cert-manager doesn't have isCA support for self-signed certs. This PR updates cert-manager by using official templates: https://github.com/jetstack/cert-manager/releases/download/v0.10.0/cert-manager.yaml

**Checklist:**
    1. `cd manifests/tests`
    2. `make generate`
    3. `make test`

/assign @krishnadurai @yanniszark 
/cc @yanniszark @lluunn @kkasravi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/369)


yanniszark:
Great work @krishnadurai, I will take a look asap.
I wonder if disabling the validation is necessary, I saw in the release notes that they now have a controller that bootstraps certificates for their webhooks.
https://github.com/jetstack/cert-manager/releases/tag/v0.10.0

krishnadurai:
@yanniszark with the release notes it does look like the disabling is not required. Though the default configuration has it disabled still. We'll need to take a look at why.
I can confirm that the required secrets are already getting created in the `cert-manager` namespace. I think if we test this functionality once we should be able to enable validation.

yanniszark:
Cert-manager maintainer @munnerz confirmed that the label should no longer be necessary. (https://kubernetes.slack.com/archives/CAR30FCJZ/p1568903338043500)

I also opened an issue in cert-manager to track it (https://github.com/jetstack/cert-manager/issues/2084)

krishnadurai:
Thanks for clearing this out @yanniszark! I'll make this change in this PR.

cert-manager/cert-manager/base/cluster-roles.yaml
```diff
@@ -0,0 +1,294 @@
...
+    app.kubernetes.io/name: cert-manager
+    app.kubernetes.io/instance:  cert-manager
+rules:
+  - apiGroups: ["certmanager.k8s.io"]
+    resources: ["certificates", "certificaterequests"]
+    verbs: ["create", "update", "delete"]
+  - apiGroups: ["certmanager.k8s.io"]
+    resources: ["certificates", "certificaterequests", "issuers", "clusterissuers"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: ["extensions"]
```
jbrette: need to add networking.k8s.io. Ingress in "extensions" are deprecated since Kubernetes 1.14.

yanniszark: Nice find!
We should also open an issue upstream since these manifests were taken from there.

yanniszark:
@krishnadurai I tried out the cert-manager installation and found an unfortunate error.

First I installed and specified the namespace for each resource.
I quickly realized that kfctl makes a kustomization that uses `namespace: "kubeflow"` by default, meaning it overrides the namespaces I'm giving my resources.

Then I tried specifying `namespace: "cert-manager"` in parameters.
It specified the correct namespace, but there's a problem.
The webhook's RoleBinding must be in the `kube-system` namespace, that is a different namespace than other resources in this kustomize package.

It seems that this use-case can't be satisfied now, because the namespace field of the kustomization is set by default, so we can't skip it.

We need to have a way to satisfy the following use-case:
* Kustomize package that contains two resources in different namespaces.

Relevant issue from kustomize: https://github.com/kubernetes-sigs/kustomize/issues/880

/cc @kkasravi @jlewi

krishnadurai:
@yanniszark Thanks for reporting this.
Given the problem, one of the ways to work-around this temporarily would be to keep this webhook RoleBinding in a separate manifest within cert-manager directory. Until the time Kustomize can get in a solution for this use-case.

cert-manager/cert-manager/base/cluster-roles.yaml
krishnadurai: @jbrette sure I'll update this config.

krishnadurai:
@yanniszark @kkasravi What do you suggest for this problem of multiple namespaces. I think we could have a work-around for now and then correct it when Kustomize has a way to address multiple namespaces in a Kustomization.

kkasravi:
@krishnadurai @yanniszark 
you can use a var namespace and set the RoleBinding namespace to use the var.
You wouldn't override the namespace in the kfctl config file

params.env
```
namespace=XXXX
```

params.yaml
```
- path: metadata/namespace
  kind: RoleBinding
```

kustomization.yaml
```
vars:
- name: namespace
  objref:
    kind: ConfigMap
    name: cert-manager-parameters
    apiVersion: v1
  fieldref:
    fieldpath: data.namespace
```

role-binding.yaml
```
apiVersion: rbac.authorization.k8s.io/v1beta1
kind: RoleBinding
metadata:
  name: cert-manager-webhook:webhook-authentication-reader
  namespace: $(namespace)
  labels:
    app: webhook
    app.kubernetes.io/name: webhook
    app.kubernetes.io/instance:  cert-manager
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: Role
  name: extension-apiserver-authentication-reader
subjects:
- apiGroup: ""
  kind: ServiceAccount
  name: cert-manager-webhook
  namespace: cert-manager
```

krishnadurai:
Thanks @kkasravi for replying! I'll try this and see if it works.

krishnadurai:
@kkasravi I've managed to make this work without specifying the namespace attribute (not parameter) in the base Kustomization file. When I use the namespace attribute the RoleBinding mentioned gets overridden to cert-manager namespace instead of the desired kube-system value. 

What I've resorted to is to use two separate namespace variables:
- namespace: for the 'cert-manager' resources
- webhookRoleBindingNamespace: for the webhook RoleBinding we have mentioned

Please see if this is appropriate. If it is, then I think this PR is ready for review.

/cc @yanniszark

yanniszark:
@krishnadurai thanks for the continuous effort on this :smile: 
I'd like to mention that `namespace` is a magic parameter for kfctl.
When `namespace` is encountered, kfctl adds `namespace: <namespace>` to the merged kustomization.yaml.
Because of that, it's not necessary to add a separate parameter.

How about:
* Adding `namespace: cert-manager` in kustomization.go so that using kustomize works normally.
* Have KfDef plugins define parameter namespace=cert-manager. This doesn't require having a corresponding params.env file, since it's a magic parameter. See the istio-install package for example.

krishnadurai:
@yanniszark You are correct about the kfctl's namespace parameter being added. This means that each resource defined in the kustomization will be declared in the same namespace however we try to parameterize it - having tested all possible scenarios.
What I've changed right now is to separate the webhook RoleBinding to a different cert-manager-webhook-role-binding package within cert-manager folder. This seems to be the only possible solution currently.

/cc @kkasravi

krishnadurai:
Test failed with error:
```
Error: couldn't delete KfApp:  (kubeflow.error): Code 500 with message: coordinator Delete failed for gcp:  (kubeflow.erro
r): Code 400 with message: Error when cleaning IAM policy: googleapi: Error 409: There were concurrent policy changes. Please retry the whole read-modify-write wi
th exponential backoff., aborted
```

jlewi:
/unassign @jlewi 
I'll let others review this one.

krishnadurai:
Circling back in here to see if this can be reviewed once?

/cc @yanniszark @kkasravi

cert-manager/cert-manager-crds/base/crds.yaml
```diff
@@ -0,0 +1,7500 @@
...
+  - name: v1alpha1
+    served: true
+    storage: true
+status:
+  acceptedNames:
+    kind: ""
+    plural: ""
+  conditions: []
+  storedVersions: []
+---
```
kkasravi: nit - only need one `---`

cert-manager/cert-manager-crds/base/crds.yaml
```diff
@@ -0,0 +1,7500 @@
...
+
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  creationTimestamp: null
```
kkasravi: nit - you don't need creationTimestamp

cert-manager/cert-manager-webhook-role-binding/base/kustomization.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: kube-system
```
kkasravi: i believe this namespace will be over-ruled by the one kfctl generates at /manifests/cert-manager/cert-manager-webhook-role-binding/kustomization.yaml

cert-manager/cert-manager-webhook-role-binding/base/role-binding.yaml
```diff
@@ -0,0 +1,22 @@
+---
```
kkasravi: nit - not needed

cert-manager/cert-manager-webhook-role-binding/base/role-binding.yaml
```diff
@@ -0,0 +1,22 @@
...
+# apiserver gets the ability to read authentication. This allows it to
+# read the specific configmap that has the requestheader-* entries to
+# api agg
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: RoleBinding
+metadata:
+  name: cert-manager-webhook:webhook-authentication-reader
+  labels:
+    app: webhook
+    app.kubernetes.io/name: webhook
```
kkasravi: these 2 labels can be removed since we'll add 
overlays/application to 
cert-manager                      
cert-manager-crds                 
cert-manager-webhook-role-binding

Can be a follow-up PR but needed for upgrade and delete.

cert-manager/cert-manager-webhook-role-binding/base/role-binding.yaml
```diff
@@ -0,0 +1,22 @@
...
+    app.kubernetes.io/instance:  cert-manager
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: extension-apiserver-authentication-reader
+subjects:
+- apiGroup: ""
```
kkasravi: do we need apiGroup?

cert-manager/cert-manager/base/api-service.yaml
```diff
@@ -0,0 +1,19 @@
+---
+apiVersion: apiregistration.k8s.io/v1beta1
+kind: APIService
+metadata:
+  name: v1beta1.webhook.certmanager.k8s.io
+  labels:
+    app: webhook
+    app.kubernetes.io/name: webhook
```
kkasravi: you can leave out these labels - see earlier comment

cert-manager/cert-manager/base/cluster-role-bindings.yaml
```diff
@@ -0,0 +1,172 @@
+---
+
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
+metadata:
+  name: cert-manager-leaderelection
+  labels:
+    app: cert-manager
+    app.kubernetes.io/name: cert-manager
```
kkasravi: remove labels

cert-manager/cert-manager/base/cluster-role-bindings.yaml
```diff
@@ -0,0 +1,172 @@
...
+
+---
+
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
+metadata:
+  name: cert-manager-controller-issuers
+  labels:
+    app: cert-manager
+    app.kubernetes.io/name: cert-manager
```
kkasravi: remove labels

cert-manager/cert-manager/base/cluster-role-bindings.yaml
```diff
@@ -0,0 +1,172 @@
...
+
+---
+
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
+metadata:
+  name: cert-manager-controller-clusterissuers
+  labels:
+    app: cert-manager
+    app.kubernetes.io/name: cert-manager
```
kkasravi: remove labels

cert-manager/cert-manager/base/cluster-role-bindings.yaml
```diff
@@ -0,0 +1,172 @@
...
+
+---
+
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
+metadata:
+  name: cert-manager-controller-certificates
+  labels:
+    app: cert-manager
+    app.kubernetes.io/name: cert-manager
```
kkasravi: remove labels

cert-manager/cert-manager/base/cluster-role-bindings.yaml
```diff
@@ -0,0 +1,172 @@
...
+
+---
+
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
+metadata:
+  name: cert-manager-controller-challenges
+  labels:
+    app: cert-manager
+    app.kubernetes.io/name: cert-manager
```
kkasravi: remove these app.kubernetes.io labels from all resources

cert-manager/cert-manager/base/deployments.yaml
```diff
@@ -0,0 +1,145 @@
...
+              memory: 32Mi
+
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: cert-manager-webhook
+  labels:
+    app: webhook
+    app.kubernetes.io/name: webhook
```
kkasravi: remove these labels

cert-manager/cert-manager/base/service-accounts.yaml
```diff
@@ -0,0 +1,29 @@
+---
```
kkasravi: nit - even with multiple resources most manifests have just called it service-account.yaml

cert-manager/cert-manager/base/services.yaml
```diff
@@ -0,0 +1,38 @@
+---
```
kkasravi: nit - even with multiple resources most manifests have just called it service.yaml

krishnadurai:
@kkasravi Thanks for your review! I'll look to address these shortly.

cert-manager/cert-manager-webhook-role-binding/base/role-binding.yaml
krishnadurai: I'll remove them here and raise a follow-up PR for application.

cert-manager/cert-manager/base/api-service.yaml
krishnadurai: Sure

cert-manager/cert-manager/base/service-accounts.yaml
krishnadurai: I'll remove all 's' trailing chars from the resources.

cert-manager/cert-manager/base/services.yaml
krishnadurai: Same here.

cert-manager/cert-manager-webhook-role-binding/base/role-binding.yaml
krishnadurai: I'll remove them throughout this package, wherever I find them in the beginning of the file.

cert-manager/cert-manager-webhook-role-binding/base/kustomization.yaml
krishnadurai: I'll strip it off here.

cert-manager/cert-manager-webhook-role-binding/base/role-binding.yaml
krishnadurai: @kkasravi Looks like I might have to add the Application changes within this PR itself since these labes (especially the `app.kubernetes.io/name` labels) are spread across:
- cert-manager
- ca-injector
- webhook

I think we can break this down into 3 different folders representing these applications and then have an Application overlay for each, so that the right labels are applied. Folders would be: 
- cert-manager
- ca-injector
- webhook

yanniszark:
>  I think we can break this down into 3 different folders representing these applications and then have an Application overlay for each, so that the right labels are applied.

Should we use different applications for each of these?
Can't they all be under a cert-manager application?

cert-manager/cert-manager-webhook-role-binding/base/role-binding.yaml
krishnadurai: I think this signifies the Core (v1) API group.

cert-manager/cert-manager-webhook-role-binding/base/kustomization.yaml
krishnadurai: This makes it easier for using with Kustomize - so I've retained it.

krishnadurai:
> Should we use different applications for each of these?
> Can't they all be under a cert-manager application?

The problem here is that the Application overlay sets a parameter like `app.kubernetes.io/name: webhook` throughout all resources. This becomes a problem for selector resource for Deployments and Services.

cert-manager/cert-manager-webhook-role-binding/base/kustomization.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
krishnadurai: Since this resource is not a part of `cert-manager-webhook` application, this does not get tagged with its label. Should we add and overlay with another set of labels to track this through kfctl?

krishnadurai:
@kkasravi this is ready for another review. I've addressed your comments.

/cc @yanniszark

Here's a KfDef in case you would like to test it:

```
# This is the config to install Kubeflow on an existing k8s cluster.
# If the cluster already has istio, comment out the istio install part below.

apiVersion: kfdef.apps.kubeflow.org/v1alpha1
kind: KfDef
metadata:
  name: k8s-istio
  namespace: kubeflow
spec:
  repos:
  - name: manifests
    root: manifests-master
    #uri: https://github.com/kubeflow/manifests/archive/master.tar.gz
    uri: https://github.com/kubeflow/manifests/archive/pull/369/head.tar.gz
  applications:
  - kustomizeConfig:
      overlays:
        - application
        - self-signed
      parameters:
      - name: namespace
        value: cert-manager
      repoRef:
        name: manifests
        path: cert-manager/cert-manager
    name: cert-manager
  - kustomizeConfig:
      overlays:
        - application
      parameters:
      - name: namespace
        value: cert-manager
      repoRef:
        name: manifests
        path: cert-manager/cert-manager-webhook
    name: cert-manager-webhook
  - kustomizeConfig:
      overlays:
        - application
      parameters:
      - name: namespace
        value: cert-manager
      repoRef:
        name: manifests
        path: cert-manager/cert-manager-ca-injector
    name: cert-manager-ca-injector
  - kustomizeConfig:
      parameters:
      - name: namespace
        value: kube-system
      repoRef:
        name: manifests
        path: cert-manager/cert-manager-webhook-role-binding
    name: cert-manager-webhook-role-binding
  enableApplications: true
  packageManager: kustomize
  skipInitProject: true
  useBasicAuth: false
  useIstio: true
  version: master
```

yanniszark:
I see. It looks like we are stumbling upon kustomize issues.
I think having 4 packages to install just cert-manager is not a scalable at all.
* KfDef specs will become unmanageably big.
* Much more work generated for package maintainers.

The cause of the issue is that a package is created for each component of cert-manager.

A solution here would be to have everything under a common app-name: cert-manager label.
After all, the `app.kubernetes.io/name` is supposed to be for the whole application ([eg wordpress](https://github.com/kubernetes-sigs/application/blob/master/docs/example.yaml)).

I believe long-term vision is to have just one kustomization package for everything (and not 2 packages, one for crds, one for controllers like now), but we need ordered resource generation (kustomize 3) for that to happen, in order for the CRDs to be generated first.

jbrette:
@yanniszark @krishnadurai @kkasravi For info you already have kustomize 3.
Second kfctl is using kustomize as a library. If you check your code here: https://github.com/kubeflow/kubeflow/blob/master/bootstrap/pkg/kfapp/kustomize/kustomize.go#L1082 you see that the last thing which is done is to invoke the legacy ordering.

My project proposed an improvment on those last transformers applied by kustomize before outputting the yaml *or what ever tools leveraging kustomize). It allows to order or/and filter. This solve the issues with CRD that have to be applied first:
Have a look at this transformer [fitler](https://github.com/keleustes/kustomize/tree/allinone/plugin/builtin/kindfiltertransformer) and this [one](https://github.com/keleustes/kustomize/blob/allinone/plugin/builtin/kindordertransformer/KindOrderTransformer.go)

Please have a look at that [example](https://github.com/keleustes/kustomize/tree/allinone/examples/issues/issue_0318_b) and in your mind replace "Values" with "KfDef". You will see how everything become more manageable.

yanniszark:
@jbrette @kkasravi it seems that the legacy reordering does take CRDs into account.
With that in mind, it seems we can start having everything (CRDs + Controllers) into one package.
The only blocker is: https://github.com/kubeflow/kubeflow/issues/3989

However, this is out of scope for this PR.
@jbrette it seems that you have a lot of experience with kustomize.
Let's discuss in a new issue how we can improve the experience of manifests.

krishnadurai:
> A solution here would be to have everything under a common app-name: cert-manager label.
> After all, the `app.kubernetes.io/name` is supposed to be for the whole application ([eg wordpress](https://github.com/kubernetes-sigs/application/blob/master/docs/example.yaml)).

@yanniszark  If we choose to go this way, we'll need to identify deployments and services with other custom labels and selectors like the way its currently done for Katib. See the example here:

Katib Controller deployment
https://github.com/kubeflow/manifests/blob/master/katib/katib-controller/base/katib-controller-deployment.yaml#L6

Katib Controller Service
https://github.com/kubeflow/manifests/blob/master/katib/katib-controller/base/katib-controller-service.yaml#L11

Katib DB deployment
https://github.com/kubeflow/manifests/blob/master/katib/katib-controller/base/katib-db-deployment.yaml#L6-L7

Katib DB Service
https://github.com/kubeflow/manifests/blob/master/katib/katib-controller/base/katib-db-service.yaml#L15-L16

I'm not sure if we should introduce custom labels and how these affect the Application CR related scenarios, like an upgrade. Probably if @kkasravi can let us know a bit about it, we should be able to take the same approach.

@jbrette Thanks for mentioning and fixing the CRD ordering with legacy ordering!

jlewi:
A couple drive by comments; not directly related to the PR but some of the surrounding discussion

* Current recommendation is to consider splitting applications into 2 kustomize packages; 1 for cluster scoped resources one for namespace scoped resources
   
   * See [application requirements](https://github.com/kubeflow/community/blob/master/guidelines/application_requirements.md#requirements)
 
   * We've had customers ask for the ability to let cluster admins install the cluster scoped resources and leave other applications for others with out privileges to install cluster scoped resources to install

* Installing multiple instances of Kubeflow per cluster generally won't work; the only thing that we officially support is a single instance per cluster

    * So I think its reasonable for KF applications to assume there is a single instance installed at a time and to set application labels based on that
       * Possible exception is during an upgrade when we have 2 different versions temporarily installed; labels should be set to support upgrades

    * Kubeflow is heavily reliant on custom resources and I think there are all kinds of problems that you will encounter trying to install and manage multiple instances of a controller and CRD per cluster

        * It would be significant effort to do this consistently across all Kubeflow applications. 
        * So for now I think we can safely assume there is a single instance of KF per cluster

yanniszark:
Thanks for the clarification @jlewi. I have a question regarding the cluster scoped distinction.

> We've had customers ask for the ability to let cluster admins install the cluster scoped resources and leave other applications for others with out privileges to install cluster scoped resources to install

From the resources installed for a package, some are Cluster Scoped Objects (CRDs) and some others are Namespace Scoped but affect the whole Cluster (controller for a CRD, which is a Deployment). The recommendation is for the 2 to go into the same package, correct?

krishnadurai:
Thanks @jlewi for clarifying.

krishnadurai:
@kkasravi @yanniszark this is ready for review.

```
apiVersion: kfdef.apps.kubeflow.org/v1alpha1
kind: KfDef
metadata:
  creationTimestamp: null
  name: k8s-istio
  namespace: kubeflow
spec:
  appdir: /Users/duselvar/golang/src/github.com/kubeflow/kubeflow/bootstrap/cert-manager-test
  applications:
  - kustomizeConfig:
      repoRef:
        name: manifests
        path: cert-manager/cert-manager-crds
    name: cert-manager-crds
  - kustomizeConfig:
      parameters:
      - name: namespace
        value: kube-system
      repoRef:
        name: manifests
        path: cert-manager/cert-manager-webhook-role-binding
    name: cert-manager-webhook-role-binding
  - kustomizeConfig:
      overlays:
      - application
      - self-signed
      parameters:
      - name: namespace
        value: cert-manager
      repoRef:
        name: manifests
        path: cert-manager/cert-manager
    name: cert-manager
  enableApplications: true
  packageManager: kustomize
  repos:
  - name: manifests
    root: manifests-master
    uri: https://github.com/kubeflow/manifests/archive/pull/369/head.tar.gz
  skipInitProject: true
  useBasicAuth: false
  useIstio: true
  version: master
```

munnerz:
👋 just a heads up, we’ve just released v0.11 of cert-manager which updates our apiVersion to a new group, as well as now using v1alpha2. There’s full details and upgrade notes on our releases page, but it may be worth updating this PR to use the new version as the CRD changes require backing up, uninstalling, reinstalling and restoring from the backup.

Feel free to reach out on the k8s slack if you’ve got any queries or concerns, or just generally want to chat about using cert-manager in kubeflow 😀

krishnadurai:
@munnerz thanks for the heads up! I'll update cert-manager to it's next version here.

krishnadurai:
@kkasravi @yanniszark I've upgraded this to v0.11.0 and tested it to ensure this is working fine.

PTAL.

cert-manager/cert-manager-kube-system-resources/overlays/application/params.env
```diff
@@ -0,0 +1 @@
+generateName=cert-manager-webhook
```
kkasravi: you can remove (generateName)

cert-manager/cert-manager-kube-system-resources/overlays/application/kustomization.yaml
```diff
@@ -0,0 +1,28 @@
...
+- name: generateName
+  objref:
+    kind: ConfigMap
+    name: cert-manager-webhook-role-binding-parameters
+    apiVersion: v1
+  fieldref:
+    fieldpath: data.generateName
+commonLabels:
+  app.kubernetes.io/name: cert-manager-webhook
+  app.kubernetes.io/instance: $(generateName)
```
kkasravi: this has been changed to remove $(generateName).

So it should look like
```
  app.kubernetes.io/name: cert-manager-webhook
  app.kubernetes.io/instance: cert-manager-webhook-v0.7.0
  app.kubernetes.io/managed-by: kfctl
  app.kubernetes.io/component: cert-manager
  app.kubernetes.io/version: v0.7.0
```

cert-manager/cert-manager-kube-system-resources/base/params.env
```diff
@@ -0,0 +1 @@
+certManagerNamespace=cert-manager
```
kkasravi: can you use 
namespace=cert-manager 
instead of 
certManagerNamespace=cert-manager
?

Other manifest targets also use namespace. 
In the config file you would provide 
```
      parameters:
      - name: namespace
        value: cert-manager
```

cert-manager/cert-manager/overlays/application/params.env
```diff
@@ -0,0 +1 @@
+generateName=cert-manager
```
kkasravi: this (generateName) can be removed since we now use cert-manager-v0.7.0 for the instance label

cert-manager/cert-manager-kube-system-resources/overlays/application/kustomization.yaml
```diff
@@ -0,0 +1,28 @@
...
+commonLabels:
+  kustomize.component: cert-manager
+configMapGenerator:
+- name: cert-manager-webhook-role-binding-parameters
+  behavior: merge
+  env: params.env
+generatorOptions:
+  disableNameSuffixHash: true
+vars:
+- name: generateName
```
kkasravi: generateName isn't needed

cert-manager/cert-manager-kube-system-resources/base/params.env
krishnadurai: Here, this variable is different than `namespace` variable. This is used to have RoleBindings link to cert-manager's namespace services.

For these resources (roles and role-bindings), `kube-system` would be the value for `namespace`.

cert-manager/cert-manager-kube-system-resources/overlays/application/kustomization.yaml
krishnadurai: Sure, I will change this part.

cert-manager/cert-manager-kube-system-resources/overlays/application/params.env
krishnadurai: Sure

cert-manager/cert-manager-kube-system-resources/overlays/application/kustomization.yaml
krishnadurai: Sure

cert-manager/cert-manager/overlays/application/params.env
krishnadurai: Done

cert-manager/cert-manager-kube-system-resources/overlays/application/params.env
krishnadurai: @kkasravi I have omitted the overlay for kube-system-resources folder since this is not an application of its own - rather resources in `kube-system` namespace for cert-manager. One option is that we could treat this as a separate application and have the overlay.

kkasravi: @krishnadurai i think that's a good suggestion. The application overlays are being used to label all resources so that `kfctl delete ...` works correctly (specifically the instance label).  So we should try and have label coverage for all resources

cert-manager/cert-manager-kube-system-resources/base/params.env
kkasravi: got it - thanks for the clarification

kkasravi:
@krishnadurai 

Looks good. I do think we need an application overlay for the kube-system-resources so we can GC only the resources we created from that namespace. If you would like to do this as a follow-up PR - it's ok with me.

kkasravi:
/lgtm
/approve
/hold

@krishnadurai do a hold cancel if you're doing the kube-system-resources application overlay as a followup PR.

krishnadurai:
@kkasravi 
I'll add the overlay in a follow-up PR.

krishnadurai:
Can I get an LGTM on this?

johnugeorge:
Adding lgtm as @kam has already approved it. 
/lgtm

krishnadurai:
Did an amend on the same commit to see if GitHub reported a conflict like it did earlier, but surprisingly, it didn't.

PR #370, opened 2019-09-19
metadata: Updating metadata component to use mlmd-grpc server

dushyanthsc:
Fixes https://github.com/kubeflow/metadata/issues/127

Updating metadata component to use mlmd gRPC server

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/370)

### DIFF
--- a/metadata/base/metadata-deployment.yaml
+++ b/metadata/base/metadata-deployment.yaml
@@ -16,7 +16,7 @@ spec:
     spec:
       containers:
       - name: container
-        image: gcr.io/kubeflow-images-public/metadata:v0.1.8
+        image: gcr.io/kubeflow-images-public/metadata:v0.1.9
         env:
           - name: MYSQL_ROOT_PASSWORD
             valueFrom:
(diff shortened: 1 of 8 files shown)


dushyanthsc:
cc: @gaoning777

metadata/base/metadata-deployment.yaml
```diff
@@ -16,20 +16,21 @@ spec:
...
                 name: metadata-db-secrets
                 key: MYSQL_ROOT_PASSWORD
-        command: ["./server/server",
-                  "--http_port=8080",
-                  "--mysql_service_host=metadata-db.kubeflow",
-                  "--mysql_service_port=3306",
-                  "--mysql_service_user=root",
-                  "--mysql_service_password=$(MYSQL_ROOT_PASSWORD)",
-                  "--mlmd_db_name=metadb"]
+        command: ["/bin/metadata_store_server"]
```
zhenghuiwang: I think current UI and SDK still depends on current HTTP server. So can you keep the existing server and add the gPRC server with a different port like 8081? We can remove existing server later.

metadata/base/metadata-envoy-deployment.yaml
```diff
@@ -0,0 +1,24 @@
...
+  selector:
+    matchLabels:
+      component: metadata-envoy
+  template:
+    metadata:
+      labels:
+        component: metadata-envoy
+    spec:
+      containers:
+      - name: container
```
zhenghuiwang: Kubeflow deploys istio envoy. Do we need a dedicated envoy for metadata?

neuromage: @zhenghuiwang good point. However, the UI will need to know how to proxy requests to the MLMD grpc server, and that config will need to be in sync between KFP and KF. My suggestion is as follows: we use this for now, but get rid of this completely in Q4 in favour of a HTTP API on top of the MLMD server. What do you think? We can also wait to complete that before updating it this way, but we need this to work now so UI folks can build the lineage parts.

zhenghuiwang: I'm okay with having a dedicated metadata envoy for now and getting rid of it later.

metadata/base/metadata-deployment.yaml
dushyanthsc: Besides the port, the existing server is still installed as part of the setup.

metadata/base/metadata-envoy-deployment.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: apps/v1beta2
+kind: Deployment
+metadata:
+  name: metadata-envoy
```
zhenghuiwang: please name it `envoy-deployment`, string prefix `metadata-` is automatically added.

metadata/base/metadata-envoy-service.yaml
```diff
@@ -0,0 +1,14 @@
+kind: Service
+apiVersion: v1
+metadata:
+  labels:
+    app: metadata
+  name: metadata-envoy-service
```
zhenghuiwang: please name it `envoy-service`, string prefix `metadata-` is automatically added.

metadata/base/metadata-deployment.yaml
zhenghuiwang: Where is the container image `gcr.io/kubeflow-images-public/metadata:v0.1.8`deployed?

Can you make the port the same? All existing examples (and possible user code) depends on `kubeflow-metadata:8080` pointing to the `metadata:v0.1.8` server. To test it, I created https://github.com/kubeflow/metadata/pull/129, which is failing right now indicating a regression.

PR #374, opened 2019-09-21
Kubernetes RBAC ClusterRole definitions

krishnadurai:
Partially Addresses: kubeflow/kubeflow#3938

Related PR: 
kubeflow/kubeflow#4167

This PR implements and shows how each application can define its own Kubernetes RBAC ClusterRole definition by using Aggregation to Kubeflow based ClusterRoles.

**Checklist:**
    1. `cd manifests/tests`
    2. `make generate`
    3. `make test`

/cc @johnugeorge 
/assign @krishnadurai

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/374)


istio/istio/base/cluster-roles.yaml
```diff
@@ -0,0 +1,53 @@
...
+kind: ClusterRole
+metadata:
+  name: kubeflow-istio-edit
+  labels:
+    rbac.example.com/aggregate-to-kubeflow-edit: "true"
+    rbac.example.com/aggregate-to-kubeflow-istio-admin: "true"
+aggregationRule:
+  clusterRoleSelectors:
+  - matchLabels:
+      rbac.example.com/aggregate-to-kubeflow-isito-edit: "true"
```
yanniszark: nit: typo

yanniszark:
Thanks a lot for the PR @krishnadurai.
Seems like a great start!

Should we use a new kubernetes package for this, or maybe reuse the profile package, as these are meant to be used for Profiles?

krishnadurai:
@yanniszark thanks for having an early look.
I'm looking to take inputs from reviewers on this subject. Although it seems appropriate to have it in Profiles - Profiles application will have its own K8s resources and will define ClusterRoles for them. Even though Profiles will be using them, I think it might be better to have these 'kubeflow' ClusterRoles as a separate entity in manifests - helping us separate the ClusterRoles.
@yanniszark @kunmingg @jlewi WDYT?

krishnadurai:
Build seems to failing without a log!

krishnadurai:
Looks like params:
```
      - name: userid-header
        value: X-Goog-Authenticated-User-Email
      - name: userid-prefix
        value: "accounts.google.com:"
```

are not set in the new Kfdef folder of manifests.

krishnadurai:
/assign @jlewi
/assign @kunmingg 
This is ready for review.

kubernetes/base/cluster-roles.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: kubeflow-edit
+  labels:
+    rbac.authorization.kubeflow.org/aggregate-to-kubeflow-admin: "true"
+aggregationRule:
+  clusterRoleSelectors:
+  - matchLabels:
```
kunmingg: Will recursive `aggregationRule` works?
How will we aggregate default role `edit` to `kubeflow-edit`?

istio/istio/base/cluster-roles.yaml
```diff
@@ -0,0 +1,51 @@
...
+
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: kubeflow-istio-edit
+  labels:
+    rbac.authorization.kubeflow.org/aggregate-to-kubeflow-edit: "true"
+    rbac.authorization.kubeflow.org/aggregate-to-kubeflow-istio-admin: "true"
+rules:
+- apiGroups: ["istio.io"]
```
kunmingg: Should we add roles for other CRs as well?

krishnadurai:
@jlewi @kunmingg This is ready for another look. I have addressed your changes.

jlewi:
@krishnadurai Instead of relying on kfctl to attach labels to the default cluster roles, would it better to just create our own base rules that define the access we need rather than depending on the default rules?

I think this has a couple advantages

i) Its more explicit about what roles Kubeflow has; less likelihood we run into problems because of default rules being different across various Kubernetes versions or flavors

ii) Less reliance on kfctl magic to properly deploy Kubeflow

Lets resolve this issue in kubeflow/kubeflow#4167

The rest of this PR looks good on my end so just waiting on other reviewers for this PR.

krishnadurai:
@jlewi ok let's choose to define K8s resource ClusterRoles.
@kunmingg @yanniszark I'll wait for your comments here and we can go ahead and merge this.
Refer comment here:
https://github.com/kubeflow/kubeflow/pull/4167#discussion_r328294420

kunmingg:
@krishnadurai 
Do you want to include base rules in this PR?

krishnadurai:
I think it will be better to merge this in and have another PR address it. Since kubeflow/kubeflow#4167 isn't in, it shouldn't change the functionality.

kubeflow-roles/base/cluster-roles.yaml
```diff
@@ -0,0 +1,39 @@
+---
```
kunmingg: Can we put this file as part of profile entry?
Since we will bind them through profile & kfam?

krishnadurai: > Can we put this file as part of profile entry?
> Since we will bind them through profile & kfam?

@kunmingg We have discussed this above in the comments:
1. https://github.com/kubeflow/manifests/pull/374#issuecomment-533793390
1. https://github.com/kubeflow/manifests/pull/374#issuecomment-533810673
1. https://github.com/kubeflow/manifests/pull/374#pullrequestreview-292117387

It looks like we have 2 votes for a separate folder and 2 for having this in profiles.

krishnadurai:
@yanniszark If you don't have any comments I think we can go ahead to get this in.

jlewi:
/lgtm
/approve

@krishnadurai You can remove the hold once @yanniszark LGTMs

kfdef/kfctl_gcp_iap.yaml
```diff
@@ -49,6 +49,13 @@ spec:
         name: manifests
         path: application/application
     name: application
+  - kustomizeConfig:
```
yanniszark: Since we don't use them yet, should we wait and edit all configs later?

kubeflow-roles/README.md
```diff
@@ -0,0 +1,74 @@
+# Default Kubeflow ClusterRoles
+
+This manifest package contains the default ClusterRoles Kubeflow uses for defining roles for Kubeflow user Profiles.
+These roles are currently assigned to users by Profiles (profile-controller and kfam) Service with the help of Manage Users page in Central Dashboard.
+
+## Aggregating default Kubernetes ClusterRoles to Kubeflow ClusterRoles
+
+Using [ClusterRole Aggregation](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles) `kfctl` currently labels each of [the default Kubernetes ClusterRoles](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#default-roles-and-role-bindings) to its Kubeflow Roles defined here.
```
yanniszark: Should this be updated to reflect the latest decision:
that we aren't labeling the default K8s ClusterRoles, but defining our own.

kfdef/kfctl_gcp_iap.yaml
krishnadurai: Sounds good. I'll make this change.

krishnadurai: Sounds good. I'll make this change.

kubeflow-roles/README.md
krishnadurai: Sure I'll update it.

krishnadurai:
@yanniszark I've addressed these changes. We should be good to go.

/cc @jlewi @kunmingg

PR #376, opened 2019-09-22
Update manifest to deploy on Kubernetes 1.16.

jbrette:
**Which issue is resolved by this Pull Request:**
Resolves #
Numerous kubeflow manifests can't deploy on Kubernetes 1.16 #375

**Description of your changes:**
- Check the apiVersion where needed.
- Update the kustomization variables definition where needed.
- Add selector to Deployment where needed.
- Rerun make generate and make test but DID NOT CHECKIN the .go files on purpose yet to not clutter the PR.

**Checklist:**
- [ ] Unit tests have been rebuilt: 
    1. `cd manifests/tests`
    2. `make generate`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/376)

PR #383, opened 2019-09-24
kfdef: generic: revert #376 

yanniszark:
PR #376 introduced changes that broke the kfdef/generic/istio package and consequently the kfctl_existing_arrikto config on master. (istio-egressgateway selector doesn't match labels).
This PR reverts the changes to kfdef/generic.

/cc @jlewi @jbrette 

**Checklist:**
    1. `cd manifests/tests`
    2. `make generate`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/383)

### DIFF
--- a/kfdef/generic/istio/istio-noauth.yaml
+++ b/kfdef/generic/istio/istio-noauth.yaml
@@ -14889,10 +14889,10 @@ rules:
 - apiGroups: [""]
   resources: ["pods", "nodes", "services", "endpoints"]
   verbs: ["get", "list", "watch"]
-- apiGroups: ["extensions", "networking.k8s.io"]
+- apiGroups: ["extensions"]
   resources: ["ingresses"]
   verbs: ["get", "list", "watch"]
-- apiGroups: ["extensions", "apps"]
+- apiGroups: ["extensions"]
   resources: ["deployments/finalizers"]
   resourceNames: ["istio-galley"]
   verbs: ["update"]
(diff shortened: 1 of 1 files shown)


kfdef/generic/istio/istio-noauth.yaml
```diff
@@ -16000,7 +15993,7 @@ spec:
 ---
 # Source: istio/charts/gateways/templates/deployment.yaml
 
-apiVersion: apps/v1
+apiVersion: extensions/v1beta1
```
jbrette: This can't be right.....This does not deploy at all on Kubernetes 1.16.
There may be a issue with selector....but not with apiVersion

jbrette:
@yanniszark I was not sure about the added selector. What is sure is that a manifest using extensions/v1beta1 for deployment....it way outdated and should have been fixed since Kubernetes 1.9.

Kubernetes 1.16 does not support Deployment extensions/v1beta1....so that part has to remain. What has to be fixed is most likely the selector.

yanniszark:
@jbrette in general, we should not merge PRs that break master.
There is no E2E test for existing_arrikto yet (soon to come https://github.com/kubeflow/kubeflow/pull/4154), but if there was one, it would have broken now.

I agree that the manifests should be upgraded to use apps/v1.
However, it should happen in a way that does not break any config.

This PR reverts the changes and a follow-up PR can change kfdef/generic to use apps/v1.

jlewi:
Let's revert and then fix.

/LGTM
/Approve

SachinVarghese:
@yanniszark Still facing issues on the master branch 
```
ERRO[0127] Failed to create Istio Control Plane: Unauthorized  filename="existing_arrikto/existing.go:459"
WARN[0127] reconcile process has error:  (kubeflow.error): Code 500 with message: coordinator Apply failed for existing_arrikto:  (kubeflow.error): Code 500 with message: Unauthorized
```

yanniszark:
@SachinVarghese this seems like an issue with the permissions of the person deploying with kfctl.
Kfctl gets its config from the user's kubeconfig, so it should be configured appropriately, with the correct permissions.
If you're still having problems, let's open a separate issue.

PR #387, opened 2019-09-25
Upgrade to Kubernetes 1.14 for GCP deployments.

jlewi:
* 1.12 is now the oldest support version on GKE.
* We want to stay ahead of GKE deprecating the version we are using by
default so we don't get surprised.

* Related to kubeflow/kubeflow#417

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/387)


jlewi:
This is fixed by #382 
https://github.com/kubeflow/manifests/blob/master/gcp/deployment_manager_configs/cluster-kubeflow.yaml#L34

PR #392, opened 2019-09-27
Update pipelines manifest to ship 0.1.31

jlewi:
* Update the images to 0.1.31

* We need to include the pipeline-visualization-service

Related to kubeflow/pipelines#2239

Use a config map to specify the config for the ml-pipeline backend server
  * Currently it is hardcoded to a config file baked into the docker image.
  * Hardcoding a config into the dockerfile generally doesn't seem like a good
    idea since its not easy to configure.

  * Per kubeflow/pipelines#2101 the hard coded config is actually broken it
    is not specifying the hostname and port of the visualization server.

**Which issue is resolved by th

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/392)

### DIFF
--- a/kfdef/kfctl_aws.yaml
+++ b/kfdef/kfctl_aws.yaml
@@ -203,6 +203,11 @@ spec:
           path: pipeline/scheduledworkflow
       name: scheduledworkflow
     - kustomizeConfig:
+        repoRef:
+          name: manifests
+          path: pipeline/pipeline-visualization-service
+      name: pipeline-visualization-service
+    - kustomizeConfig:
         overlays:
           - istio
         repoRef:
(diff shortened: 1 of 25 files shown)


neuromage:
@jlewi would you mind rebasing on master (with #424 merged in), and try this again please? Thanks.

jlewi:
I rebased on master to pull in the metadata fix. Just waiting for the tests to pass and then this should be ready for review.

jlewi:
Tests passed before I rebased to squash the commits so this is ready for review.

/assign @neuromage 
/assign @Ark-kun

neuromage:
/lgtm
Thanks @jlewi

Issue #315, opened 2019-08-25
Update SeldonCore manifests to 0.4


jlewi:
We need to update the SeldonCore manifests to 0.4.

There was an attempt to update the manifests in #273 

But this appeared to break the kubeflow/kubeflow E2E tests (see kubeflow/kubeflow#4001). 

So we rolled back the change to the seldon manifests in #312 

The break should have been caught by kubeflow/manifests presubmits but I suspect it wasn't because the presubmits aren't using the correct version of the manifests repo.

jlewi:
kubeflow/manifests/issues/316 is why the presubmits didn't catch the break in the original PR.

cliveseldon:
Does kfctl modify the manifests in any way? As there is no PersistenVolumeClaim in the file:

https://github.com/kubeflow/manifests/blob/76bd875d65d39960ab2d75daa2ba2b7a2a9e82a4/seldon/seldon-core-operator/base/seldon-operator-controller-manager-statefulset.yaml

The kustomize works on its own. What's the correct way to test with kfctl PRs to manifests repo so we can find what the issue is?

cliveseldon:
I tested with master kfctl modified to print out what its creating and I see a VolumeClaimTemplate
```
[...]

PR #312, opened 2019-08-24
Revert "Update Seldon Manifest for 0.4 Release of Seldon (#273)"

kunmingg:
This reverts commit 76bd875d65d39960ab2d75daa2ba2b7a2a9e82a4.

Seem presubmit is broken by this PR. 

**Description of your changes:**

**Checklist:**
- [ ] Unit tests have been rebuilt: 
    1. `cd manifests/tests`
    2. `make generate`
    3. `make test`

Related to kubeflow/kubeflow#4001

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/312)

PR #403, opened 2019-09-30
Seldon 0.4.1

cliveseldon:
* Update Seldon to 0.4.1

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/403)

### DIFF
--- a/seldon/Makefile
+++ b/seldon/Makefile
@@ -1,6 +1,6 @@
 
 seldon-core:
-	git clone -b release-0.4 --single-branch git@github.com:SeldonIO/seldon-core.git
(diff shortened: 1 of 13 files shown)


jlewi:
@cliveseldon Do you recall what the issue was the last time we tried to update the manifests and had to revert it #312 ?

It looks like at the time our presubmits may not have been properly configured to test changes but I believe that should be fixed now. So the fact the presubmit is passing is encouraging.

cliveseldon:
@jlewi yeah its https://github.com/kubeflow/manifests/issues/315#issuecomment-524799621

PR #406, opened 2019-10-01
Katib v1alpha3 manifests

richardsliu:
This PR contains the manifests for Katib v1alpha3. I've also combined the Katib components into two packages:
- katib-crds
- katib-controller

This is similar to what https://github.com/kubeflow/manifests/pull/251 was trying to achieve. 

I've also named the top-level package as just "katib" as opposed to "katib-<version>". Going forward we should not be installing multiple katib versions in the same deployment.

Also note that the tag is not updated - we will update this when the Katib v0.7 release candidate is finalized.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/406)


katib/katib-controller/base/katib-configmap.yaml
```diff
@@ -0,0 +1,39 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: katib-config
+  namespace: kubeflow
```
johnugeorge: Namespace is not needed

katib/katib-controller/base/trial-template-configmap.yaml
```diff
@@ -0,0 +1,28 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: trial-template
+  namespace: kubeflow
```
johnugeorge: Namespace is not needed. There are many occurrences in this patch.

richardsliu:
Presubmit failure should be fixed by https://github.com/kubeflow/kubeflow/pull/4218.

PR #419, opened 2019-10-02
make generate about to break when kustomize new release is published.

jbrette:
Resolves #
Kustomize is about to change its interfaces, and this will breaks the "make generate" and "make test"
- Changes in the fs packages.
- Internalization of the build command.
- Various cleanups.

Just add the go.mod and remove it from the .gitignore. This will force make generate and make test to use a version of kustomize compatible with kubeflow instead of letting it pickup the latest published version of kustomize.

Another PR will have to be created to adapt the code to the new version of kustomize once it is released.

**Checklist:**
    1. `cd manifests/tests`
    2. `make generate`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/419)


go.mod
```diff
@@ -0,0 +1,5 @@
+module github.com/kubeflow/manifests
```
kkasravi: shouldn't we have the full go.mod?

jbrette: not sure. I ran GO111MODULE=on go mod tidy....and it seems this is the only dependency we need.....which is good.

jbrette: removed go get ...

jbrette:
@kkasravi 
- Done requested code review.
- Tested inside and outside GOPATH.
- Added test.go that where missing after yaml was changed with rerunning generation

/assign @jlewi

kkasravi:
@jbrette could i ask you to rebase from master?
`git fetch upstream master`

jbrette:
> @jbrette could i ask you to rebase from master?
> git fetch upstream master
> git rebase -i upstream/master

done. reran make generated and git push --force. You will have to reapply labels

tests/notebook-controller-overlays-application_test.go
```diff
@@ -49,7 +49,7 @@ spec:
      - jupyter
      - notebook
      - notebook-controller
-     - jupyterhub
+     - jupyterhub
```
jlewi: Why do we still have jupyterhub here? That package no longer exists.

jbrette: This is why: https://github.com/kubeflow/manifests/blob/master/jupyter/notebook-controller/overlays/application/application.yaml#L35

@jlewi @kkasravi I real don't want to be in the same situation than we you asked to remove Tiller and Heritage at the same time people were reverting changes to go back to deprecated version of Ingress, boggus changes which are still here by way as as the Tiller stuff.

So tell me what you want....I can drop the "generation code" commit from this PR so that you can merge...and one of you will have to run make generate again later

kkasravi: it is unrelated to what you're trying to get merged. 
It's ok with me if we merge this and then remove jupyter in a subsequent PR.
@jlewi ?

/lgtm

PR #421, opened 2019-10-03
error when deploying istio #420

### LINKED ISSUE #420: error when deploying istio
several problems when adding the application overlay to istio manifests: istio-crds, istio-install

1. istio-crds shouldn't set the namespace to be istio-system and use the default namespace. This is because the application overlay for istio-crds will attempt to add the application CR and ConfigMap to istio-system which isn't created until istio-install. The error is below:

```
customresourcedefinition.apiextensions.k8s.io/templates.config.istio.io created
customresourcedefinition.apiextensions.k8s.io/tracespans.config.istio.io created
customresourcedefinition.apiextensions.k8s.io/virtualservices.networking.istio.io created
customresourcedefinition.apiextensions.k8s.io/zipkins.config.istio.io created
Error from server (NotFound): error when creating "/tmp/kout164753882": namespaces "istio-system" not found
Error from server (NotFound): error when creating "/tmp/kout164753882": namespaces "istio-system" not found
```

2. During istio-install it patches deployment selectors which have already been set by the application overlay. Subsequently we hit `selector is immutable` - see https://github.com/kubernetes/kubernetes/pull/50719#.  This fix will require we remove selector fields in all resources in istio-noauth.yaml and add these selector fields to commonLabels in istio/istio-install/overlays/application/kustomization.yaml.

kkasravi:
Resolves #420

setting the namespace parameter for istio-crds to istio-system causes it to try and put the application CR and related configmap in the istio-system namespace which hasn't been created yet

adding the application overlay to istio-install adds selectors to istio deployments. Istio attempts to do sidecar injections which generate selector patches to these deployments. Selectors are [immutable](https://github.com/kubernetes/kubernetes/pull/50719#) so these operations fail.  

**Checklist:**
    1. `cd manifests/tests`
    2. `make generate`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/421)

### DIFF
--- a/gcp/iap-ingress/overlays/application/params.yaml
+++ b/gcp/iap-ingress/overlays/application/params.yaml
@@ -3,6 +3,12 @@ varReference:
   kind: Application
 - path: spec/selector/matchLabels/app.kubernetes.io\/instance
   kind: Application
+- path: spec/selector/app.kubernetes.io\/instance
+  kind: Service
+- path: spec/selector/matchLabels/app.kubernetes.io\/instance
+  kind: Deployment
+- path: spec/template/metadata/labels/app.kubernetes.io\/instance
+  kind: Deployment
 - path: spec/selector/matchLabels/app.kubernetes.io\/instance
   kind: StatefulSet
 - path: spec/template/metadata/labels/app.kubernetes.io\/instance
--- a/gcp/iap-ingress/overlays/application/params.yaml
+++ b/gcp/iap-ingress/overlays/application/params.yaml
@@ -7,3 +7,5 @@ varReference:
   kind: StatefulSet
 - path: spec/template/metadata/labels/app.kubernetes.io\/instance
   kind: StatefulSet
+- path: spec/template/metadata/labels/app.kubernetes.io\/instance
+  kind: Job
--- a/kfdef/kfctl_aws.yaml
+++ b/kfdef/kfctl_aws.yaml
@@ -7,6 +7,11 @@ spec:
   platform: aws
   applications:
     - kustomizeConfig:
+        repoRef:
+          name: manifests
+          path: application/application-crds
+      name: application-crds
+    - kustomizeConfig:
         overlays:
           - application
         parameters:
@@ -36,11 +41,6 @@ spec:
           name: manifests
           path: istio/istio
       name: istio
-    - kustomizeConfig:
-        repoRef:
-          name: manifests
-          path: application/application-crds
-      name: application-crds
     - kustomizeConfig:
         overlays:
           - application
--- a/kfdef/kfctl_aws.yaml
+++ b/kfdef/kfctl_aws.yaml
@@ -17,11 +17,7 @@ spec:
           path: istio/istio-crds
       name: istio-crds
     - kustomizeConfig:
-        overlays:
-          - application
         parameters:
-          - name: generateName
-            value: istio-install-
           - name: namespace
             value: istio-system
         repoRef:
--- a/kfdef/kfctl_aws.yaml
+++ b/kfdef/kfctl_aws.yaml
@@ -12,8 +12,6 @@ spec:
         parameters:
           - name: generateName
             value: istio-crds-
-          - name: namespace
-            value: istio-system
         repoRef:
           name: manifests
           path: istio/istio-crds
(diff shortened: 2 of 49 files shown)


gcp/iap-ingress/overlays/application/params.yaml
```diff
@@ -7,3 +7,5 @@ varReference:
   kind: StatefulSet
 - path: spec/template/metadata/labels/app.kubernetes.io\/instance
   kind: StatefulSet
+- path: spec/template/metadata/labels/app.kubernetes.io\/instance
+  kind: Job
```
krishnadurai: I'm trying to understand this. Just correct me here if I'm not getting this right - the Application overlay should label all resources belonging to `iap-ingress`. So here all resources:
- BackendConfig.yaml
- CloudEndpoint
- ClusterRoleBinding.yaml
- ClusterRole
- ConfigMap
- Deployment
- Ingress
- Policy
- ServiceAccount
- Service

Should get labelled with `app.kubernetes.io/instance`.

And likewise, for all other application/params.yaml in this PR. This, of course, doesn't need to be a part of this PR.

krishnadurai: Ok, I think I figured this out.

Looks like the labels not in metadata/labels require a varReference.

kkasravi: @krishnadurai @richardsliu @jlewi @yanniszark 
here are some details. I think your understanding is correct, but I'll explain as well:

Adding the application overlay creates an Application CR for that manifest. Applications will track their components and report their status. This is key for e2e tests since we now have a better approach to health monitoring than just checking deployments ([the current e2e tests](https://github.com/kubeflow/kubeflow/blob/master/testing/kfctl/kf_is_ready_test.py#L35)). 

We add labels to all resources per application - for example in [centraldashboard](https://github.com/kubeflow/manifests/blob/master/common/centraldashboard/overlays/application/kustomization.yaml#L20). Notice the $(generateName) in the labels. Each application has a unique name which kfctl will generate by adding a suffix to the name. For example centraldashboard will have a name like [centraldashboard-akdfjls](https://github.com/kubeflow/manifests/blob/master/kfdef/kfctl_gcp_iap.yaml#L93).

When kustomize sees commonLabels it adds these to all resources labels. This is [recommended](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels) by kubernetes. Kustomize will also add selectors to resources like deployments, statefulsets, daemonsets, etc. Unfortunately kustomize [isn't doing var replacement](https://github.com/kubernetes-sigs/kustomize/issues/1585) when it adds selectors. 

Additionally adding the instance label to all resources allows us to delete all resources including cluster wide resources even if we're not adding that resource to the application.componentKinds. Deletion of cluster wide resources has been broken for a while in kubeflow so this is a high priority.

Finally this is needed for app upgrades.

krishnadurai: Thanks @kkasravi!
I think we could capture the application CR part in docs too.

kfdef/kfctl_aws.yaml
```diff
@@ -6,24 +6,23 @@ metadata:
...
+    - kustomizeConfig:
+        repoRef:
+          name: manifests
+          path: application/application-crds
+      name: application-crds
     - kustomizeConfig:
         overlays:
           - application
         parameters:
           - name: generateName
```
swiftdiaries: I saw some failures with the generateName as well. The Application CRD didn't have a name field and the deployment failed as a result.

kkasravi: in the config file - if it has
```
overlays:
- application
```
it will need
```
parameters:
- name: generateName
   value: <name>-
```

swiftdiaries: Got it, thanks :) 
I must've missed that

jlewi:
Does #420 mean the manifests for ISTIO currently on master are broken? If so why wasn't the breaking change caught by presubmits?

I really think we need to prioritize fixing #302 so its easy to manually verify what the generated manifests look like.

swiftdiaries:
I think when the PR was merged it was using the existing KfDef to test and not the incoming one. So it probably passed presubmit.

kkasravi:
closing since we reverted

PR #424, opened 2019-10-03
Metadata fix

neuromage:
Partially resolves kubeflow/pipelines#2239

Add virtual service for metadata grpc server and point mlmd grpc deployment to newer image with the right service name for the deployed grpc server.

**Checklist:**
    1. `cd manifests/tests`
    2. `make generate`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/424)

### DIFF
--- a/metadata/base/kustomization.yaml
+++ b/metadata/base/kustomization.yaml
@@ -43,6 +43,13 @@ vars:
     apiVersion: v1
   fieldref:
     fieldpath: metadata.name
+- name: metadata-envoy-service
+  objref:
+    kind: Service
+    name: envoy-service
+    apiVersion: v1
+  fieldref:
+    fieldpath: metadata.name
 images:
 - name: mysql
   newName: mysql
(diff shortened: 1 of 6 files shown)


neuromage:
/assign @zhenghuiwang 
/assign @jlewi 
/assign @abhi-g 

I think this should fix the problem for KFP UI in 0.7 deployment.

abhi-g:
Thanks @neuromage for the change. I am assuming we'll need this in conjunction with  @jlewi 's work in progress PR #392 ?

There seem to be quite a few files in the diff that are primarily arising due to just the formatting change of some of the configs. Is there a way to remove those changes or does Git have a way to diff ignoring those?

It seems like there are 2 critical changes: one of them to the metadata-envoy container image and other other with the virtual service for metadata-envoy with the correct URL prefix.

neuromage:
@abhi-g yes, we'll still need #392 as well for all the fixes to go in. Agreed about those spurious changes. I can try to revert those, but would you know how I can do this automatically? I just ran make generate which seems to have done that. I'm not sure why that happened though. Let me try to see if I can revert them.

neuromage:
@abhi-g @zhenghuiwang I've reset the spurious changes. PTAL, thanks.

PR #431, opened 2019-10-04
Removing unnecessary configs from Pytorch operator 

johnugeorge:
v1beta2 support is removed. 
ConfigMap is not needed any more.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/431)


richardsliu:
Are the corresponding tests updated? Will we have test failures because v1beta2 is deprecated?

johnugeorge:
@richardsliu which tests are you referring to?

PR #432, opened 2019-10-04
Removing tf-job dashboard from deployment

johnugeorge:
TF job dashboard is removed from deployment spec
Unnecessary rbac permissions are removed.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/432)


songkane:
Any reason remove tf-job dashboard ?

Issue #426, opened 2019-10-03
kubeflow/manifests @ master is broken - uber issue


jlewi:
Master of kubeflow/manifests appears to be broken.

There appear to be three known issues

1. #420 and #425 are breaking the ISTIO and knative install
1. #342 pinned kubeflow/manifests in kfctl to some presubmits

At issue appears to be some very large PRs #342 and #410 that touched a number of different components.

kunmingg:
We can 
1. set test config to 
```
uri: https://github.com/kubeflow/manifests/archive/3e85b3c83de12fed7d675151d0d4f8474eb04790.tar.gz
```
2. revert broken PRs.

jlewi:
I think we need to roll back all PRs on kubeflow/manifests 1 by 1 so that we roll back master to 
3e85b3c83de12fed7d675151d0d4f8474eb04790.

jlewi:
I suspect that one reason the presubmits didn't catch the breaks to ISTIO is we don't run the test_endpoint test are part of our presubmits.

jlewi:
I manually deployed using 
https://github.com/kubeflow/manifests/pull/429

I hit https://github.com/kubeflow/kfserving/issues/399 but when I retried kfctl apply it succeeded.

I'm still waiting for my endpoint to be ready but otherwise everything looks good.

jlewi:
#429 is merged. 

Here's the follow on work

* Roll forward #424 metadata fix
* Roll forward #401 update jupyter images

Not sure what to do about the 3 other massive PRs

* #342 app instance
* #410 first pass remaining applications
* #405 missing images to kustomization

#405 was probably not the culprit but #342 and #410 were probably causing problems. 

Should we improve the testing before rolling those forward

* Enable test_endpoint to try to ensure ISTIO install is working
* Fix #306 so its easier to manually verify what the expected output of the kustomization is

jlewi:
@yanniszark @neuromage @Tabrizian FYI that your PRs were at least temporarily rolled back see above.

yanniszark:
@jlewi I have #444 made to revert #405.

jlewi:
I'm going to close this issue.

* Master has been fixed
* All the PRs that should be rolled forward either have been rolled forward or are out for review
* @kkasravi is working on revising his PRs to define Application resources for all apps and will break it up into smaller PRs.

PR #435, opened 2019-10-04
just add application overlays and generateName to current application…

kkasravi:
Resolves #

Adding the application overlay creates an Application CR for that manifest. Applications will track their components and report their status. This is key for e2e tests since we now have a better approach to health monitoring than just checking deployment status (the current e2e tests).

We also add recommended labels to all resources that have an application overlay as [suggested](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels) by kubernetes. For the **app.kubernetes.io/instance** recommended label we use the kustomize var $(generateName). When kfctl sees this parameter in a config file such as below:
```
      overlays:
      - istio
      - application
      parameters:
      - name: generateName
        value: jupyter-webapp-
```
it will generate a suffix and use this unique name in the instance label. Generating a unique name is also [specified](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#applications-and-instances-of-applications) in the kubernetes documentation. For example centraldashboard will have a name generated like centraldashboard-akdfjls. Why don't we use [resource.metadata.generateName](https://kubernetes.io/docs/reference/using-api/api-concepts/#generated-values) instead of $(generateName)? This can only be used on kubectl create, whereas we run the equivalent of kubectl apply. 

Adding the instance label to all resources allows us to delete all resources including cluster wide resources even if we're not adding that resource to the application.componentKinds. Deletion of cluster wide resources has been broken for a while in kubeflow so this is a high priority.

Finally this is needed for app upgrades.
**Checklist:**
    1. `cd manifests/tests`
    2. `make generate`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/435)


jlewi:
Before doing more mass refactorings of the manifests should we 

1. Restore the other, unrelated PRs that were reverted as part of #426 ?
1. At least partially address how evaluate the correctness of our manifests #306?

Can you update the PR description to explain the motivation behind the introduction of a parameter for the application name? Is this because we want to give different versions of an application (e.g. v0.6.2 vs. v0.7.) different instance names? And we can't rely on kustomize's suffix generator or similar because those names would change every time we run kustomize build?

Although looking at the KFDef changes it looks like each application is setting generateName to some non-unique value "kubeflow-"

With respect to better evaluating the correctness of our manifests #306 can we do something like the following

1. Include a suitable kustomization.yaml for every application that includes the suitable overlays
    * e.g. include the istio and application overlays
    * e.g jupyter-web-app is missing a suitable kustomizatin.yaml file in https://github.com/kubeflow/manifests/tree/master/jupyter/jupyter-web-app

1. Provide a script that runs `kustomize build` for all of these applications and emits the output to a suitable directory that would get checked in

    * As a reviewer I can now see the actual output and a diff with respect to previous output
       * e.g. I can verify  how generateName is changed

    * These can become the golden manifests (or at least 1 set) against which our unittests run

jlewi:
@kkasravi Instead of relying on kfctl to generate a unique instance name why can't each Application just set the appropriate labels in the YAML defining the application? 

IIUC the primary reason we need uniqueness for the instance names is because during upgrades we want to distinguish between the 0.X version of the application and the 0.Y application of the instance. But we could just make it the responsibility of the maintainers of the application manifest to bump the instance name on each version. We could even build tooling to enforce this consistently across all applications.

Going a step further we could establish a convention that instance names should follow a specific naming schema like

```
${APPNAME}-${VERSION}
```

So for jupyter webapp it would be

```
jupyter-webapp-v0-7-0
```

This way we don't have to push additional complexity onto kfctl and KFDef.


> Adding the instance label to all resources allows us to delete all resources including cluster wide resources even if we're not adding that resource to the application.componentKinds. Deletion of cluster wide resources has been broken for a while in kubeflow so this is a high priority.

How are instance labels being added to all the resources? Is this kfctl magic?

jlewi:
> Include a suitable kustomization.yaml for every application that includes the suitable overlays

We are building more "magic" into kfctl; i.e. we are relying on kfctl to generate kustomization files.

This makes testing harder because we need to recreate the kfctl magic in order to generate a kustomization file to be used for testing. 

Ideally all of the kustomize applications should be buildable on their own and get a suitable/working set of spec. 

For example I should be able to do the following

```
git clone https://github.com/kubeflow/manifests.git
cd manifests/jupyter/jupyter-web-app
kustomize build
```
This doesn't work right now because we don't have a kustomization.yaml in that directory because we depend on KFctl to generate it with various parameters such as the overlays.

But we should have a default manifests/jupyter/jupyter-webapp/kustomization.yaml that includes suitable defaults (e.g. the Application and ISTIO overalys).

From a testing perspective I should now be able to do something like

```
cd manifests/jupyter/jupyter-web-app
kustomize build
```

And compare the output to some expected set of YAML manifests. That won't cover all possible cases but it should cover X% of them. Then we would just need to add more specialized use cases for different bits of customization.

jlewi:
> Restore the other, unrelated PRs that were reverted as part of #426 ?

I already sent out PRs to restore the 2 most important ones and they have been merged so I think we are good to go.

jlewi:
Thanks @kkasravi Could you explain though  why we are setting instanceName using generateName which is getting set in the kfdefs.yaml file?

Why don't we just set instanceName in the kustomization file?

For example, the jupyter-web-app already defines commonLabels here
https://github.com/kubeflow/manifests/blob/ea58d15690ffe02b3c388390460db1b8cbbf918c/jupyter/jupyter-web-app/base/kustomization.yaml#L14

So why can't we just change that to 

```
commonLabels:
  app: jupyter-web-app
  kustomize.component: jupyter-web-app
  app.kubernetes.io/instance: jupyter-web-app-vX-Y-Z
```

jupyter/jupyter-web-app/overlays/application/application.yaml
```diff
@@ -1,16 +1,11 @@
...
 spec:
   selector:
     matchLabels:
-      app.kubernetes.io/name: jupyter-web-app
-      app.kubernetes.io/instance: jupyter-web-app
-      app.kubernetes.io/managed-by: kfctl
-      app.kubernetes.io/component: jupyter
-      app.kubernetes.io/part-of: kubeflow
-      app.kubernetes.io/version: v0.6
+      app.kubernetes.io/instance: $(generateName)
```
jlewi: @kkasravi What's the reasoning for introducing generateName here? The selector was already matching a set of labels that was already being consistently applied across all the components via kustomization.yaml commonLabels.

Why are we parameterizing this using generateName?

jupyter/jupyter-web-app/overlays/application/kustomization.yaml
```diff
@@ -4,9 +4,22 @@ bases:
...
+    name: jupyter-web-app-parameters
+    apiVersion: v1
+  fieldref:
+    fieldpath: data.generateName
+configurations:
+- params.yaml
 commonLabels:
   app.kubernetes.io/name: jupyter-web-app
-  app.kubernetes.io/instance: jupyter-web-app
+  app.kubernetes.io/instance: $(generateName)
```
jlewi: Why are we parameterizing the instance name rather than just explicitly coding it here in common labels?

jupyter/jupyter-web-app/overlays/application/application.yaml
```diff
@@ -1,16 +1,11 @@
 apiVersion: app.k8s.io/v1beta1
 kind: Application
 metadata:
-  name: jupyter-web-app
+  name: $(generateName)
```
jlewi: Why are we parameterizing the name rather than just explicitly setting it here? What was wrong with just hardcoding the name here?

jlewi:
I'm still not sure I fully follow here.

Lets start with the KFDef files. Why is KFDef being used to set the instance label by adding a stanza like this for **every** application to **every* KFDef file?

```
      parameters:
      - name: generateName
        value: notebook-controller-
```

This stanza doesn't look like its conveying any information that an application shouldn't already know; i.e. what does the notebook controller need the KFDef file to set a parameter to know that its name is "notebook-controller"?

Related to that why are we generating random suffixes to use as the instance label? For example why would the instance label for centraldashboard be like `centraldashboard-akdfjls` where the suffix is some randomly generated value as opposed to something like `centraldashboard-vX-Y-Z` that is explicitly set in the kustomize manifests for central dashboard?

> the instance name should also be the name of the application CR, this is how the uber application CR collects the nested application CRs.

The labels and selector just need to be set consistently. So can't we just do this in the kustomize manifests? IIUC for every application we would just need to set two places consistently

1. We would have to set common labels in kustomization.yaml
1. We would have to set an appropriate label selector in the Application CR

It looks like that was already happening e.g. in jupyter_web_app. See my inline comments on jupyter_web_app.

jlewi:
Kubeflow only currently supports one instance of each Kubeflow application per cluster.

To support more than one instance of an application per cluster we would have to solve all kinds of other problems. 

For example, if you install multiple versions of CR controller per cluster, how do you decide which resources get handled by which controller?

So if we assume that we only need to support one instance per cluster; does that allow us to simplify things?

For example, do we need to use random salt for the instance names?

The only use case we have right now for multiple instances per cluster would be when we are upgrading Kubeflow and we have a 0.X and 0.Y instance both installed for a brief period of time before cleaning up 0.X.

kkasravi:
closing this PR - reopening with just one application (jupyter-web-app)

PR #440, opened 2019-10-06
Prep for kfctl_k8s_istio.yaml E2E tests

swiftdiaries:
Resolves #
https://github.com/kubeflow/kubeflow/issues/3496
Addresses issue about testing kfctl_k8s_istio KfDef.

KinD is a lightweight Kubernetes distribution that is meant for
testing environments and local development. We use this primarily for
testing the `kfctl_k8s_istio` config and the related components in Kubeflow CI.
Like KinD, this is NOT meant for production workloads.
Please refer to MiniKF for a production-ready environment :)

This PR adds config files required for the `platform: kind`

Goes together with PR: https://github.com/kubeflow/kubeflow/pull/4237

**Checklist:**
    1. `cd manifests/tests`
    2. `make generate`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/440)


kind/local-path-storage.yaml
```diff
@@ -0,0 +1,109 @@
...
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: local-path-provisioner-role
+subjects:
+- kind: ServiceAccount
+  name: local-path-provisioner-service-account
+  namespace: local-path-storage
+---
+apiVersion: apps/v1beta2
```
nrchakradhar: nit: Is it better to use apps/v1?

kind/local-path-storage.yaml
```diff
@@ -0,0 +1,109 @@
...
+metadata:
+  name: local-path-storage
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: local-path-provisioner-service-account
+  namespace: local-path-storage
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
```
nrchakradhar: nit: Is it better to use v1 for ease in working with later kubernetes version?

kind/local-path-storage.yaml
```diff
@@ -0,0 +1,109 @@
...
+  resources: ["endpoints", "persistentvolumes", "pods"]
+  verbs: ["*"]
+- apiGroups: [""]
+  resources: ["events"]
+  verbs: ["create", "patch"]
+- apiGroups: ["storage.k8s.io"]
+  resources: ["storageclasses"]
+  verbs: ["get", "list", "watch"]
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
```
nrchakradhar: nit: Is it better to move to v1 for easier compatibility with future kuberenetes versions?

nrchakradhar:
Question: Other than what is recommended by KinD for CPU, RAM, does Kubeflow need any extra resources for CPU, RAM. I understand that storage is changed.

swiftdiaries:
I use KinD locally but with Istio + Knative installed it starts to become really slow, without Istio + Knative it seems to run ok. 
I have 7 CPUs and 4Gi of memory set in my Docker resources. But ymmv.

swiftdiaries:
Actually I haven't switched out the storage class yet. I should do that in a follow-up PR. But for the E2E tests it shouldn't matter.

jlewi:
If the goal is to test kfctl_k8s_istio.yaml; why are you defining a new KFDef file?And why are there 2 different ones kfctl_k8s_kind.yaml and kfctl_k8s_istio_kind.yaml?

I would expect an E2E test to work as follows

1. Provision a suitable K8s cluster (e.g. using kops or kind)
1. Run `kfctl apply -f kfctl_k8s_istio.yaml`

If we start defining and using new KFDef files then how are we E2E testing our existing KFDef files?

swiftdiaries:
@johnugeorge also had a similar concern.
I'll re-do the PR to have the kind-config alone for the E2E tests to pick it up from here.
The E2E test in the `kubeflow/kubeflow` would then download and install `kind` to create a cluster using the CLI using the `kind-config` and `kfctl_k8s_istio.yaml` here.

yanniszark:
@swiftdiaries this seems to be a general concern for configs that don't include cluster creation.
Maybe we can add new steps to create/delete a K8s cluster?

swiftdiaries:
Yes. I tried going in the direction that GCP does today by including a cluster creation/deletion in the config. But that just adds another config.

Adding a new step in the E2E workflow seems like the better approach.

jlewi:
100% agree; adding cluster creation/deletion steps in the E2E test is the better approach.

jlewi:
@swiftdiaries can we close this PR? IIUC we are going to add appropriate create/delete steps to the E2E workflow rather than creating a new KFDef and plugin?

swiftdiaries:
Oh I need the config for the KinD cluster and the local-storage-class for it.
So I'm adding a KinD cluster creation / deletion as separate steps in the workflow.
Does it make sense to put those here or put them in `kubeflow/kubeflow`?

swiftdiaries:
/close
Moving the config files into the `kubeflow/kubeflow` repo

PR #443, opened 2019-10-07
Update the notebook-controller image.

jlewi:
* It looks like the code may not have actually changed since we last
  built the image. Nonethless, lets update the tag to use the sha
  as opposed to a tag like v20190911-e8193317-dirty-cd2831

Related to: kubeflow/kubeflow#3655

* Disable the test that images are listed in kustomization.yaml because it doesn't work with digests
* Related to #449 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/443)

### DIFF
--- a/jupyter/notebook-controller/base/kustomization.yaml
+++ b/jupyter/notebook-controller/base/kustomization.yaml
@@ -15,7 +15,7 @@ commonLabels:
 images:
 - name: gcr.io/kubeflow-images-public/notebook-controller
   newName: gcr.io/kubeflow-images-public/notebook-controller
-  newTag: v20190911-e8193317-dirty-cd2831
+  digest: sha256:6490f737000bd1d2520ac4b8cbde2b09749cdb291b1967ddda95d05131db49db
 configMapGenerator:
 - name: parameters
   env: params.env
--- a/tests/notebook-controller-base_test.go
+++ b/tests/notebook-controller-base_test.go
@@ -254,7 +254,7 @@ commonLabels:
 images:
 - name: gcr.io/kubeflow-images-public/notebook-controller
   newName: gcr.io/kubeflow-images-public/notebook-controller
-  newTag: v20190911-e8193317-dirty-cd2831
+  digest: sha256:6490f737000bd1d2520ac4b8cbde2b09749cdb291b1967ddda95d05131db49db
 configMapGenerator:
 - name: parameters
   env: params.env
--- a/tests/notebook-controller-overlays-application_test.go
+++ b/tests/notebook-controller-overlays-application_test.go
@@ -310,7 +310,7 @@ commonLabels:
 images:
 - name: gcr.io/kubeflow-images-public/notebook-controller
   newName: gcr.io/kubeflow-images-public/notebook-controller
-  newTag: v20190911-e8193317-dirty-cd2831
+  digest: sha256:6490f737000bd1d2520ac4b8cbde2b09749cdb291b1967ddda95d05131db49db
 configMapGenerator:
 - name: parameters
   env: params.env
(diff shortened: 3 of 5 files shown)


jupyter/notebook-controller/base/kustomization.yaml
```diff
@@ -15,7 +15,7 @@ commonLabels:
 images:
   - name: gcr.io/kubeflow-images-public/notebook-controller
     newName: gcr.io/kubeflow-images-public/notebook-controller
-    newTag: v20190911-e8193317-dirty-cd2831
+    digest: sha256:6490f737000bd1d2520ac4b8cbde2b09749cdb291b1967ddda95d05131db49db
```
lluunn: I think the 0911 should also be the latest.
Btw why use digest?
I think tag with the date should be more informative?

jlewi: shas are immutable whereas tags are not.
Using tags has a couple side effects

1) The tag names often end up including information we don't want

     * E.g; do we really want dirty in the tag?
     * If you rebuild the image using GCB from head you end up getting the following tag
       
       ```
       v20191007-v0.6.0-rc.0-134-g6ddf28cc-dirty-560d95 
       ```

     * This would likely confuse people because it says v0.6.0 but really its a v0.7.0 image
     * So one way to avoid that type of confusion is to just use the sha we can then map the sha to additional meta information as needed
    * e.g [http://gcr.io/kubeflow-images-public/notebook-controller@sha256:6490f737000bd1d2520ac4b8cbde2b09749cdb291b1967ddda95d05131db49db](http://gcr.io/kubeflow-images-public/notebook-controller@sha256:6490f737000bd1d2520ac4b8cbde2b09749cdb291b1967ddda95d05131db49db)  takes you to the GCR page for the image where you can see all the metadata infromation

1. We are trying to create automatic workflows to update the images and this will likely be easier to make work with the shas.

lluunn: got it, thanks for the explanation

lluunn:
```
Error: images missing from kustomization files.
diff --git a/jupyter/notebook-controller/base/kustomization.yaml b/jupyter/notebook-controller/base/kustomization.yaml
index f41c547..fefdc76 100644
--- a/jupyter/notebook-controller/base/kustomization.yaml
+++ b/jupyter/notebook-controller/base/kustomization.yaml
@@ -16,6 +16,7 @@ images:
- name: gcr.io/kubeflow-images-public/notebook-controller
  newName: gcr.io/kubeflow-images-public/notebook-controller
  digest: sha256:6490f737000bd1d2520ac4b8cbde2b09749cdb291b1967ddda95d05131db49db
+  newTag: v20190614-v0-160-g386f2749-e3b0c4
```

https://github.com/kubeflow/manifests/blob/abace77e6734d31afd2f398c3e8449580cf5dece/tests/scripts/run-tests.sh#L38

weird, image is there

jlewi:
I think its the new script that @yanniszark wrote.

yanniszark:
Let me take a look.

yanniszark:
I see what is going on.
The script also fills in newTag with the tag from the image it parsed.
So it asks of you to also fill in the newTag field while that is not necessary.

I believe these lines can safely go away and will solve your issue:
https://github.com/kubeflow/manifests/blob/abace77e6734d31afd2f398c3e8449580cf5dece/tests/scripts/extract_images.py#L44-L47

They are filling newTag and digest but I don't think that's necessary, name and newName is enough for the set-image-name script to function.

jlewi:
@yanniszark Can you open a fix in a separate PR. I don't want to block this PR trying to fix the new test.

PR #447, opened 2019-10-08
add kfserving to existing_arrikto

ryandawsonuk:
existing_arrikto wasn't included in https://github.com/kubeflow/kubeflow/pull/3953 but kfserving can be used with the rest of the existing_arrikto setup

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/447)

### DIFF
--- a/kfdef/kfctl_anthos.yaml
+++ b/kfdef/kfctl_anthos.yaml
@@ -95,6 +95,32 @@ spec:
         path: metadata
     name: metadata
   - kustomizeConfig:
+      parameters:
+      - name: namespace
+        value: knative-serving
+      repoRef:
+        name: manifests
+        path: knative/knative-serving-crds
+    name: knative-crds
+  - kustomizeConfig:
+      parameters:
+      - name: namespace
+        value: knative-serving
+      repoRef:
+        name: manifests
+        path: knative/knative-serving-install
+    name: knative-install
+  - kustomizeConfig:
+      repoRef:
+        name: manifests
+        path: kfserving/kfserving-crds
+    name: kfserving-crds
+  - kustomizeConfig:
+      repoRef:
+        name: manifests
+        path: kfserving/kfserving-install
+    name: kfserving-install
+  - kustomizeConfig:
       overlays:
       - istio
       - application
--- a/kfdef/kfctl_anthos.yaml
+++ b/kfdef/kfctl_anthos.yaml
@@ -95,32 +95,6 @@ spec:
         path: metadata
     name: metadata
   - kustomizeConfig:
-      parameters:
-      - name: namespace
-        value: knative-serving
-      repoRef:
-        name: manifests
-        path: knative/knative-serving-crds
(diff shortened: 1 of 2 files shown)


ryandawsonuk:
@johnugeorge @yanniszark

yanniszark:
That's great @ryandawsonuk.
Have you tried deploying it with existing_arrikto?
Does it work out of the box or are changes needed?
Can you point me to an example I can run to confirm it works?

ryandawsonuk:
I've certainly been using it without any changes.

jlewi:
@yanniszark Do you plan on testing this? How do you want to proceed?

jlewi:
@yanniszark should we go ahead with this change? I think we want to get this change in because the v0.7.0 blog post will highlight KFServing so it would be odd not to have KFServing included in that config.

@ryandawsonuk I think we need to make two changes 

1. On master kfctl_existing_arrikto.yaml was renamed to kfctl_istio_dex
    https://github.com/kubeflow/manifests/blob/master/kfdef/kfctl_istio_dex.yaml

    * Hence the conflict

We should probably cherrypick this into v0.7.0 for kfctl_existing_arrikto.yaml; it would be good to get @yanniszark to sign off on that.

yanniszark:
@jlewi I agree about including KFServing, my only concern is that it comes with its own Istio Gateway and I haven't validated that the configuration works as expected.
cc @krishnadurai

krishnadurai:
@yanniszark looks like the gateway here is just for knative and it shouldn't affect Auth.

The usage for kfserving right now is through Jupyter NB and kubectl which should be covered with Dex's auth.

krishnadurai:
@ryandawsonuk you could just resolve the conflict since the file has been renamed to kfctl_istio_dex.yaml. I could LGTM as soon as that change is made.

ryandawsonuk:
@krishnadurai have resolved conflict
@yanniszark , as @krishnadurai says KFServing is currently independent of the auth setup because it uses its own gateway. This means KFServices can currently be accessed without authentication - there's an [open ticket to bring it under a single auth gateway in the future](https://github.com/kubeflow/kfserving/issues/371).

jlewi:
@ryandawsonuk @yanniszark the 0.7 blog post is making a big deal about KFServing so it would be great to get it cherry-picked onto the existing_arrikto configs on the v0.7 branch as well. I believe the other deployments (GKE and AWS) already have it.

yanniszark:
The existing_arrikto config is not yet ready to add kfserving.
We use a commit separate from the v0.7 branch because we need some extra commits (authservice, cert manager)

I would suggest getting those commits cherry picked to v0.7-branch, pointing the existing_arrikto repo there and then adding kfserving.
/hold

I'll make an issue asap.

yanniszark:
Update: issue tracked here in #616 
After #616 is resolved (I have made the necessary PRs), we should be able to use v0.7-branch and go forward with the kfserving addition.

PR #462, opened 2019-10-09
Replaces null values at ClusterRoles with [] - for kubectl CLI compatibility

krishnadurai:
Kustomize CLI manifest application, for e.g. `kustomize build . | kubectl apply -f -` fails with:
```
error validating "STDIN": error validating data: ValidationError(ClusterRole): missing required field "rules" in io.k8s.api.rbac.v1.ClusterRole; if you choose to ignore these errors, turn validation off with --validate=false
```

This seems to be a kubectl client-side validation bug as mentioned in this [comment](https://github.com/kubernetes/kubernetes/issues/73050#issuecomment-457068012)

Another problem is with `make test`, which warns about:
```
nil value at `rules.resourceNames` ignored in mutation attempt
```

**Checklist:**
    1. `cd manifests/tests`
    2. `make generate`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/462)

### DIFF
--- a/admission-webhook/webhook/base/cluster-role.yaml
+++ b/admission-webhook/webhook/base/cluster-role.yaml
@@ -28,7 +28,7 @@ aggregationRule:
   clusterRoleSelectors:
   - matchLabels:
       rbac.authorization.kubeflow.org/aggregate-to-kubeflow-poddefaults-admin: "true"
-rules: null
+rules: []
 
 ---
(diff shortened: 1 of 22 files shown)


krishnadurai:
@kunmingg can we please try to get this minor fix in?

PR #470, opened 2019-10-09
add application overlays for pipeline

### LINKED ISSUE #469: add application overlays for pipeline
```
└── pipeline
    ├── api-service
    ├── minio
    ├── mysql
    ├── persistent-agent
    ├── pipeline-visualization-service
    ├── pipelines-runner
    ├── pipelines-ui
    ├── pipelines-viewer
    └── scheduledworkflow
```

kkasravi:
Resolves #469 

```
└── pipeline
    ├── api-service
    ├── minio
    ├── mysql
    ├── persistent-agent
    ├── pipeline-visualization-service
    ├── pipelines-runner
    ├── pipelines-ui
    ├── pipelines-viewer
    └── scheduledworkflow
```
**Checklist:**
    1. `cd manifests/tests`
    2. `make generate`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/470)

### DIFF
--- a/istio/istio-install/base/istio-noauth.yaml
+++ b/istio/istio-install/base/istio-noauth.yaml
@@ -14332,6 +14332,7 @@ spec:
           - /etc/config/validatingwebhookconfiguration.yaml
           - --monitoringPort=15014
           - --log_output_level=default:info
+          - --enable-validation=true
           volumeMounts:
           - name: certs
             mountPath: /etc/certs
--- a/kfdef/kfctl_anthos.yaml
+++ b/kfdef/kfctl_anthos.yaml
@@ -180,11 +180,15 @@ spec:
         path: katib/katib-controller
     name: katib-controller
   - kustomizeConfig:
+      overlays:
+      - application
       repoRef:
         name: manifests
         path: pipeline/api-service
     name: api-service
   - kustomizeConfig:
+      overlays:
+      - application
       parameters:
       - name: minioPvcName
         value: minio-pv-claim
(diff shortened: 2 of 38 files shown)


jlewi:
Pipelines has their own versioning scheme. The release is 0.1.31
https://github.com/kubeflow/pipelines/releases

So that's what we should set in the application resources.

yanniszark:
@kkasravi isn't the application `Kubeflow Pipelines` and runner, persistence-agent, etc components of that application?
Shouldn't we group all of these components under a single Kubeflow Pipelines Application, as different components?
It seems more in line with the [official Application CRD example](https://github.com/kubernetes-sigs/application/blob/master/docs/example.yaml), where a Server, a Database and a Dashboard are grouped under a wordpress Application.

kkasravi:
> @kkasravi isn't the application `Kubeflow Pipelines` and runner, persistence-agent, etc components of that application?
> Shouldn't we group all of these components under a single Kubeflow Pipelines Application, as different components?
> It seems more in line with the [official Application CRD example](https://github.com/kubernetes-sigs/application/blob/master/docs/example.yaml), where a Server, a Database and a Dashboard are grouped under a wordpress Application.

@yanniszark yes the top level directories under manifests do suggest these types of groupings: gcp, katib, admission-webhook, pipeline, though KfDef isn't organized this way. Upgrades (which use applications) would be much more granular and the Application CR doesn't provide any type of sub-grouping. You would have Application.componentKinds which would include the aggregate list of k8 resources across manifests/pipeline/*. Health level info would also be more granular and possibly more difficult to diagnose. If there were any ordering dependencies under pipeline (as there is under admission-webhook), the Application.componentKinds array doesn't capture this either. I think once we replace the overlays/mixins with kustomize/v3 this might be easier to do.

PR #493, opened 2019-10-11
update profile config to v1beta1

kunmingg:
Resolves #https://github.com/kubeflow/kubeflow/issues/3822
Resolves #https://github.com/kubeflow/kubeflow/issues/3823

**Checklist:**
    1. `cd manifests/tests`
    2. `make generate`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/493)


profiles/base/deployment.yaml
```diff
@@ -13,6 +14,8 @@ spec:
         - $(userid-header)
         - "-userid-prefix"
         - $(userid-prefix)
+        - "-workload-identity"
```
jlewi: Should this be a GCP specific overlay? What does gcp-sa get set for in non GCP configs? The empty string?

jlewi: Actually won't we get a bug if the gcp-sa parameter isn't defined?
I think a GCP specific overlay might be better.

profiles/base/kustomization.yaml
```diff
@@ -17,11 +18,11 @@ configMapGenerator:
   env: params.env
 images:
 - name: gcr.io/kubeflow-images-public/profile-controller
-  newName: gcr.io/kubeflow-images-public/profile-controller
-  newTag: v20190619-v0-219-gbd3daa8c-dirty-1ced0e
+  newName: gcr.io/kubeflow-dev/profile-controller
```
jlewi: Why aren't the updated images in kubeflow-images-public?

profiles/base/kustomization.yaml
```diff
@@ -17,11 +18,11 @@ configMapGenerator:
...
+  newTag: v20191010-v0.6.0-rc.0-155-g1486acd7-dirty-087a92
```
jlewi: How about using digest rather than tags?

profiles/base/kustomization.yaml
```diff
@@ -30,6 +31,13 @@ vars:
     apiVersion: v1
   fieldref:
     fieldpath: data.admin
+- name: gcp-sa
```
jlewi: I think this will cause an error if gcp-sa isn't explicitly defined in the parameters.
Forcing every KFDef to explicitly set GCP specific parameters to the empty string doesn't seem scalable.

profiles/base/deployment.yaml
```diff
@@ -13,6 +14,8 @@ spec:
...
+        - $(gcp-sa)
```
jlewi: Longer term:

An alternative pattern that might be more convenient would be to mount the config map as a file in the volume and then pass the config path as a command line argument.

Then overlays can just add parameters to that config map but the deployment doesn't have to change. 

Right now we are relying on kustomize to do the substitution when kustomize build is called.

Lets not do that now though because it would require changes to the profiler code.

profiles/base/deployment.yaml
kunmingg: Sure
if it's empty string, then the plugin would not be executed.

profiles/base/kustomization.yaml
kunmingg: The default value is empty string, if param not set, the plugin will be ignored.

profiles/base/deployment.yaml
jlewi: I stand corrected. It looks like the way this works is that kfctl is editing
base/params.env
https://github.com/kubeflow/kubeflow/blob/fa8886eae006dee0270bcc8bc1f5efc9e3c82e48/bootstrap/pkg/kfapp/kustomize/kustomize.go#L645

The logic updates any parameters that are set there.
So if the default in params.env is gcp_sa="" then it should evaluate correctly.

jlewi:
@kunmingg The test failure of kfctl_is_ready is most likely because you are missing kubeflow/manifests#498.

Can you please rebase on the latest master?

jlewi:
Test failures looks like a resource quota issue in kubeflow-ci-deployment

```
util.py                     71 INFO     time="2019-10-14T18:10:31Z" level=error msg="Creating kfctl-0697 error: &{Code:RESOURCE_ERROR Location:/deploymen
ts/kfctl-0697/resources/kfctl-0697 Message:{\"ResourceType\":\"gcp-types/container-v1beta1:projects.locations.clusters\",\"ResourceErrorCode\":\"FAILED_O
PERATION\",\"ResourceErrorMessage\":\"The network \\\"default\\\" does not have available private IP space in 10.0.0.0/8 to reserve a /14 block for conta
iners for cluster {us-central1-a, 29647740582, kubeflow-ci-deployment, kfctl-0697, a944743a20d815f1431c88c2965811e1afbacb4eb48712637ecba03d7a7e1cfe}. - h
ttps://container.googleapis.com/v1beta1/projects/29647740582/zones/us-central1-a/clusters/kfctl-0697\"} ForceSendFields:[] NullFields:[]}" filename="gcp/
gcp.go:380"
util.py                     71 INFO     Error: couldn't apply KfApp:  (kubeflow.error): Code 500 with message: coordinator Apply failed for gcp:  (kubefl
ow.error): Code 400 with message: gcp apply could not update deployment manager Error could not update deployment manager entries; Creating kfctl-0697 er
ror(400): BAD REQUEST
```

kunmingg:
Seems auto clean up not working, manual delete is working.

profiles/base/crd.yaml
```diff
@@ -2,17 +2,18 @@ apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   creationTimestamp: null
-  labels:
-    controller-tools.k8s.io: "1.0"
-  name: profiles.kubeflow.org
+  name: profiles.profile.kubeflow.org
 spec:
-  group: kubeflow.org
+  group: profile.kubeflow.org
```
jlewi: Lets leave the group alone for now to avoid risk.

profiles/base/crd.yaml
```diff
@@ -2,17 +2,18 @@ apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   creationTimestamp: null
-  labels:
-    controller-tools.k8s.io: "1.0"
-  name: profiles.kubeflow.org
+  name: profiles.profile.kubeflow.org
```
jlewi: Lets change the name back to profiles.kubeflow org for now.

profiles/base/cluster-role-binding.yaml
```diff
@@ -1,11 +1,12 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name: cluster-role-binding
+  name: profiles-cluster-role-binding
```
jlewi: Why do these need to change? Doesn't kustomization.yaml add a unique prefix?

profiles/base/deployment.yaml
```diff
@@ -1,8 +1,10 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: deployment
+  name: profiles-deployment
```
jlewi: Why are you adding profiles as a prefix? Doesn't kustomization.yaml take care of that?

profiles/base/deployment.yaml
```diff
@@ -1,8 +1,10 @@
...
+  namespace: kubeflow
```
jlewi: Why set namespace? Doesn't kustomization.yaml take care of that?

jlewi:
Chatted with @kunmingg the reason he is manually setting prefix is because he added to the package a cluster scoped resource to define an actual Profile which is what kfctl will use to create the profile. 

He's going to move that into a separate kustomize package.

profiles/base/crd.yaml
kunmingg: https://github.com/kubeflow/kubeflow/pull/4299

kfdef/kfctl_gcp_iap.yaml
```diff
@@ -358,6 +358,16 @@ spec:
         name: manifests
         path: seldon/seldon-core-operator
     name: seldon-core-operator
+  - kustomizeConfig:
+      parameters:
+        - name: user
+          value: SET_EMAIL
```
jlewi: Why do you have dummy parameters here but not in kfctl_basic_auth?

profiles/overlays/debug/deployment.yaml
```diff
@@ -1,7 +1,8 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: deployment
+  name: profiles-deployment
+  namespace: kubeflow
```
jlewi: Should these changes be reverted?

profiles/overlays/devices/deployment.yaml
```diff
@@ -1,7 +1,8 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: deployment
+  name: profiles-deployment
+  namespace: kubeflow
```
jlewi: Should this change be reverted?

kfdef/kfctl_gcp_iap.yaml
kunmingg: kfctl_basic_auth will use default param value "anonymous".

profiles/overlays/debug/deployment.yaml
kunmingg: Good catch

kfdef/kfctl_gcp_iap.yaml
jlewi: Lets add a comment explaining that this gets overwritten by kfctl.

profiles/base/service.yaml
```diff
@@ -2,6 +2,7 @@ apiVersion: v1
 kind: Service
 metadata:
   name: kfam
+  namespace: kubeflow
```
jlewi: Revert this?

kunmingg:
rebase to resolve conflicts

kunmingg:
related: https://github.com/kubeflow/kubeflow/pull/4309

kunmingg:
Failure was caused by timeout deleting DM entry.
/retest

jlewi:
kubeflow/kfctl#56 for kfctl_test_delete_failures

PR #513, opened 2019-10-12
On GCP use BackendConfig to set the backend timeout and other improvments

jlewi:
* As of GKE 1.13 we can now set the backend timeout using a BackendConfig
  https://cloud.google.com/kubernetes-engine/docs/how-to/configure-backend-service

* BasicAuth doesn't have a backendconfig so we need to add one.

  * In BasicAuth we are still using Ambassador to handle the auth.

* Improve the setup_backend.sh script for IAP

  * This script is only updating the JWT based on IAP; there's no reason
    the script should be checking if the backend timeout is set.

  * Make the script run for ever to avoid confusing container restarts

* Related to kubeflow/kubeflow#4269 backend timeout not set
* Related to kubeflow/kubeflow#4043 crash looping of iap enabler and backend_updater pods.

* For the kustomize test target generation
  * Add an option to regenerate all tests.
  * use the current branch when computing the diff rather than assuming
    the remote fork name is origin.

Resolves #

**Checklist:**
    1. `cd manifests/tests`
    2. `make generate`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/513)

### DIFF
--- a/common/ambassador/base/service.yaml
+++ b/common/ambassador/base/service.yaml
@@ -17,6 +17,9 @@ spec:
 apiVersion: v1
 kind: Service
 metadata:
+  annotations:
+    # Ambassador is only used on GCP with basic auth.
+    beta.cloud.google.com/backend-config: '{"ports": {"ambassador":"basicauth-backendconfig"}}'
   labels:
     service: ambassador
   name: ambassador
--- a/gcp/basic-auth-ingress/base/config-map.yaml
+++ b/gcp/basic-auth-ingress/base/config-map.yaml
@@ -3,8 +3,8 @@ data:
   update_backend.sh: |
     #!/bin/bash
     #
-    # A simple shell script to configure the backend timeouts and health checks by using gcloud.
-
+    # A simple shell script to configure the health checks by using gcloud.
+    set -x 
     [ -z ${NAMESPACE} ] && echo Error NAMESPACE must be set && exit 1
     [ -z ${SERVICE} ] && echo Error SERVICE must be set && exit 1
     [ -z ${INGRESS_NAME} ] && echo Error INGRESS_NAME must be set && exit 1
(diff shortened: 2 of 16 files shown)


jlewi:
I deployed manually with IAP and verified the backend timeout is set to 3600.

```
gcloud --project=jlewi-dev compute backend-services describe --global k8s-be-31380--3ec84ae56f727bf0
affinityCookieTtlSec: 0
backends:
- balancingMode: RATE
  capacityScaler: 1.0
  group: https://www.googleapis.com/compute/v1/projects/jlewi-dev/zones/us-east1-d/instanceGroups/k8s-ig--3ec84ae56f727bf0
  maxRatePerInstance: 1.0
connectionDraining:
  drainingTimeoutSec: 0
creationTimestamp: '2019-10-12T17:16:06.845-07:00'
description: '{"kubernetes.io/service-name":"istio-system/istio-ingressgateway","kubernetes.io/service-port":"80"}'
enableCDN: false
fingerprint: q6hdyut7CW4=
healthChecks:
- https://www.googleapis.com/compute/v1/projects/jlewi-dev/global/healthChecks/k8s-be-31380--3ec84ae56f727bf0
iap:
  enabled: true
  oauth2ClientId: 147474701642-rn0ksbu4eikorfk8j0uvknubbke57ri1.apps.googleusercontent.com
  oauth2ClientSecretSha256: 5aa08d6104d663f1f6588be497db5dea6dee39f4cde87daf98b156183c70c910
id: '2103215684271985193'
kind: compute#backendService
loadBalancingScheme: EXTERNAL
name: k8s-be-31380--3ec84ae56f727bf0
port: 31380
portName: port31380
protocol: HTTP
selfLink: https://www.googleapis.com/compute/v1/projects/jlewi-dev/global/backendServices/k8s-be-31380--3ec84ae56f727bf0
sessionAffinity: NONE
timeoutSec: 3600
```

jlewi:
Test error looks like a flake

```
util.py                     71 INFO     error: error when creating "/tmp/kout501751445": Unexpected error &net.OpError{Op:"read", Net:"tcp", Source:(*net
.TCPAddr)(0xc00001fb60), Addr:(*net.TCPAddr)(0xc00001fb90), Err:(*os.SyscallError)(0xc000650de0)} when reading response body. Please retry.
generated xml file: /mnt/test-data-volume/kubeflow-manifests-presubmit-kfctl-go-513-d2bf292-0848-1675/output/artifacts/junit_kfctl-build-testiap-istio.xml
```

gcp/basic-auth-ingress/base/config-map.yaml
```diff
@@ -15,58 +15,67 @@ data:
...
-    do BACKEND_SERVICE=$(gcloud --project=${PROJECT} compute backend-services list --filter=name~k8s-be-${NODE_PORT}- --uri);
-    echo "Waiting for the backend-services resource PROJECT=${PROJECT} NODEPORT=${NODE_PORT} SERVICE=${SERVICE}...";
-    sleep 2;
-    done
-
-    while [[ -z ${HEALTH_CHECK_URI} ]];
-    do HEALTH_CHECK_URI=$(gcloud compute --project=${PROJECT} health-checks list --filter=name~${BACKEND_NAME} --uri);
-    echo "Waiting for the healthcheck resource PROJECT=${PROJECT} NODEPORT=${NODE_PORT} SERVICE=${SERVICE}...";
-    sleep 2;
+    set_health_check() {
```
lluunn: According to this: https://stackoverflow.com/questions/44584270/how-to-get-a-custom-healthcheck-path-in-a-gce-l7-balancer-serving-a-kubernetes-i

It seems that if we properly set the readiness probe, GCE ingress should pick up that as the health check path.
So we can fix here https://github.com/kubeflow/manifests/blob/master/istio/istio-install/base/istio-noauth.yaml#L14658, and get rid of this update backend?

common/ambassador/base/service.yaml
```diff
@@ -17,6 +17,8 @@ spec:
 apiVersion: v1
 kind: Service
 metadata:
+  annotations:
+    beta.cloud.google.com/backend-config: '{"ports": {"ambassador":"iap-backendconfig"}}'
```
lluunn: should be basicauth-backendconfig?

gcp/basic-auth-ingress/base/config-map.yaml
jlewi: I think it would be great if we could get rid of it but I'm wary of removing it without more testing and investigation. 

IIRC there might have been some sort of race condition where if the ingress is created before the deployment the health check won't get set correctly to the value used in the deployment.

We should be able to test that by modifying the deployment and seeing if the backend health check is updated correctly.

jlewi: Filed: kubeflow/manifests#521

lluunn: thanks

common/ambassador/base/service.yaml
jlewi: Fixed

jlewi:
@lluunn PTAL addressed your comments.

Issue #524, opened 2019-10-14
upgrade all application instance labels to v0.7.0


kkasravi:
$ ☞  vgrep 'v0.6'|grep -v tests
admission-webhook/webhook/overlays/application/kustomization.yaml
admission-webhook/webhook/overlays/application/application.yaml
admission-webhook/bootstrap/overlays/application/kustomization.yaml
admission-webhook/bootstrap/overlays/application/application.yaml
application/application/overlays/application/kustomization.yaml
application/application/overlays/application/application.yaml
aws/aws-alb-ingress-controller/overlays/application/kustomization.yaml
aws/aws-alb-ingress-controller/overlays/application/application.yaml
jupyter/jupyter-web-app/overlays/application/kustomization.yaml
jupyter/jupyter-web-app/overlays/application/application.yaml
jupyter/notebook-controller/overlays/application/kustomization.yaml
jupyter/notebook-controller/overlays/application/application.yaml
mxnet-job/mxnet-operator/overlays/application/kustomization.yaml
mxnet-job/mxnet-operator/overlays/application/application.yaml
pytorch-job/pytorch-operator/base/deployment.yaml
pytorch-job/pytorch-operator/base/kustomization.yaml
seldon/seldon-core-operator/overlays/application/kustomization.yaml
spark/spark-operator/overlays/application/kustomization.yaml
spark/spark-operator/overlays/application/application.yaml
tektoncd/tektoncd-install/overlays/application/kustomization.yaml
tektoncd/tektoncd-dashboard/overlays/application/kustomization.yaml

richardsliu:
Can we put this in a 1-line script?

kkasravi:
yep

PR #533, opened 2019-10-15
Add a script to update app instance and version labels

richardsliu:
Fix https://github.com/kubeflow/manifests/issues/524

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/533)

### DIFF
--- a/hack/update-instance-labels.sh
+++ b/hack/update-instance-labels.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+# Replace 'app.kubernetes.io/version: v0.6.2' with 'app.kubernetes.io/version: v0.7.0'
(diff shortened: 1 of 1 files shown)


hack/update-instance-labels.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+# Replace 'app.kubernetes.io/version: v0.6.x' with 'app.kubernetes.io/version: v0.7.0'
+grep -rl --exclude-dir=hack 'app.kubernetes.io/version: v0.6' ./ | xargs sed -i -E 's/app.kubernetes.io\/version: v0.6(.*)/app.kubernetes.io\/version: v0.7.0/g'
```
kkasravi: it would be nice to pull these versions from git tags or something eventually.

kkasravi: we may want to exclude what hack/gen-target-tests.sh excludes which is

kfdef|gatekeeper|gcp/deployment_manager_configs|aws/infra_configs|doc|tests|hack|plugins

for example I think it will pick up files under tests

PR #534, opened 2019-10-15
Update labels to v0.7.0

richardsliu:
Resolves https://github.com/kubeflow/manifests/issues/524

**Checklist:**
    1. `cd manifests/tests`
    2. `make generate`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/534)

### DIFF
--- a/hack/update-instance-labels.sh
+++ b/hack/update-instance-labels.sh
@@ -1,7 +1,9 @@
 #!/bin/bash
 
-# Replace 'app.kubernetes.io/version: v0.6.x' with 'app.kubernetes.io/version: v0.7.0'
-grep -rl --exclude-dir=hack 'app.kubernetes.io/version: v0.6' ./ | xargs sed -i -E 's/app.kubernetes.io\/version: v0.6(.*)/app.kubernetes.io\/version: v0.7.0/g'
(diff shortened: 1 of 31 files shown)


hack/update-instance-labels.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+# Replace 'app.kubernetes.io/version: v0.6.x' with 'app.kubernetes.io/version: v0.7.0'
+grep -rl --exclude-dir=hack 'app.kubernetes.io/version: v0.6' ./ | xargs sed -i -E 's/app.kubernetes.io\/version: v0.6(.*)/app.kubernetes.io\/version: v0.7.0/g'
```
kkasravi: may want to include the same directories for exclusion as hack/gen-test-targets.sh just so we don't inadvertently change something we don't want to

kfdef|gatekeeper|gcp/deployment_manager_configs|aws/infra_configs|doc|tests|hack|plugins

richardsliu: We do want to update tests, right?

kkasravi: yep, this does what it needs to then
/lgtm
/approve

richardsliu:
@kkasravi Updated.

PR #541, opened 2019-10-16
kubeflow on GCP: create empty GPU pool by default

kunmingg:
DM will create empty GPU pool by default when user deploy on GKE.

**Checklist:**
    1. `cd manifests/tests`
    2. `make generate`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/541)

### DIFF
--- a/tests/profiles-overlays-test_test.go
+++ b/tests/profiles-overlays-test_test.go
@@ -22,13 +22,24 @@ bases:
 #generators:
 #- app_test.yaml
 `)
+	th.writeF("/manifests/profiles/base/cluster-role-binding.yaml", `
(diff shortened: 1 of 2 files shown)


tests/profiles-overlays-test_test.go
```diff
@@ -21,23 +21,37 @@ bases:
 - ../../base
 #generators:
 #- app_test.yaml
+`)
```
lluunn: is the change in this file needed?

kunmingg: Yes, seems this test on master head is broken

PR #550, opened 2019-10-21
add dockerSockPath to argo config map

adrian555:

Since Argo v2.4.0, new config `dockerSockPath` is added to allow different location of docker.sock. Ref. https://github.com/argoproj/argo/pull/1419.
To reflect the change in Argo and to better support different docker installation, add the `dockerSockPath` to the Argo config map.

Add `dockerSockPath` to argo base config map and related kustomization yaml as well as the tests.

**Checklist:**
    1. `cd manifests/tests`
    2. `make generate`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/550)

### DIFF
--- a/tests/argo-base_test.go
+++ b/tests/argo-base_test.go
@@ -135,6 +135,7 @@ data:
     {
     executorImage: $(executorImage),
     containerRuntimeExecutor: $(containerRuntimeExecutor),
+    dockerSockPath: $(dockerSockPath),
     artifactRepository:
     {
         s3: {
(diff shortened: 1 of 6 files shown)


tests/argo-base_test.go
```diff
@@ -135,6 +135,7 @@ data:
     {
     executorImage: $(executorImage),
     containerRuntimeExecutor: $(containerRuntimeExecutor),
+    dockerSockPath: $(dockerSockPath),
```
hougangliu: for now, Kubeflow install v2.3.0 argo. does `dockerSockPath` work for v2.3.0?

adrian555: @hougangliu this just adds one config to the workflow-controller-configmap which is not used in v2.3.0. It does not break argo workflow functionality. I have tested this with a fresh install of kubeflow 0.6.2 and run the sample pipeline successfully.

PR #557, opened 2019-10-22
make namespace explicit

ryandawsonuk:
I found that this does get installed in the istio-system namespace so would be better to make that explicit

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/557)

### DIFF
--- a/istio/oidc-authservice/base/envoy-filter.yaml
+++ b/istio/oidc-authservice/base/envoy-filter.yaml
@@ -2,6 +2,7 @@ apiVersion: networking.istio.io/v1alpha3
 kind: EnvoyFilter
 metadata:
   name: authn-filter
+  namespace: istio-system
 spec:
   workloadLabels:
     istio: ingressgateway


istio/oidc-authservice/base/envoy-filter.yaml
```diff
@@ -2,6 +2,7 @@ apiVersion: networking.istio.io/v1alpha3
 kind: EnvoyFilter
 metadata:
   name: authn-filter
+  namespace: istio-system
```
gabrielwen: this is controlled by kustomize overlays.

ryandawsonuk: Ok makes sense. How come not with https://github.com/kubeflow/manifests/blob/1bb77f3f28ce901947d80161f8e6ff83cafbfa1d/kfdef/generic/auth_oidc/envoy-filter.yaml#L4 ?

gabrielwen: I think someone forgot to remove it?  @kkasravi

krishnadurai: @ryandawsonuk the manifests you mention are not built through kustomize.go. These are included in the platform specific code - bypassing kustomize's namespace feature.

/cc @gabrielwen

Issue #267, opened 2019-08-10
Integrate Dex Auth with Multi User Story


krishnadurai:
Current Dex implementation needs to integrate well with Multi-User Story in the following areas:

- User RBAC is generated as per the roles created through ProfilesController
- Ingress Auth is enabled for Dex based Auth

Reference issue/PR:
https://github.com/kubeflow/manifests/pull/195

/cc @jlewi @kkasravi 

/assign @krishnadurai

jlewi:
@krishnadurai Can you please add

* Priority for this issue
* Area for this issue
* A kind (bug, enhancement etc...)
* Any projects that this should belong to

jlewi:
I think one of the issues that came up in PR #195 was reconciling the default RBAC and ISTIO RBAC roles and bindings created in #195 with the roles and bindings created by the profile custom resource

jlewi:
@krishnadurai What are the next steps here? I believe the existing_arrikto.yaml config is already using Dex and ISTIO.

I think one question that came up is can we generalize what's used in existing_arrikto.yaml to make it easier to compose the Dex setup with installations? (see #3739).

For GCP we would like to get rid of our custom basic auth setup and just install Dex with an appropriate authenticator.

Another question that has come up is where is the source code for the OIDC connector used in kubeflow_existing_arrikto.yaml? I believe the config is currently referencing a container in an Arrikto GCR container.

How does key keycloak-gatekeeper interact with Dex? Is keycloak trying to do the same thing Dex is doing by creating a proxy for OpenID providers?

/cc @yanniszark

krishnadurai:
@jlewi keycloak-gatekeeper is being used as an authentication proxy for Kubeflow like what IAP does for GCP IAP config. This will also be used to add additional headers required for multi-user in the future. For now, I've been able to make this proxy work on the web UI for Authentication (login/logout) flow. I'm finalizing it to work with kfctl and HTTPS. Soon Multi-User support will follow.

> How does key keycloak-gatekeeper interact with Dex? Is keycloak trying to do the same thing 
Dex is doing by creating a proxy for OpenID providers?

keycloak-gatekeeper acts as a client application to Dex, which means - if it doesn't find the user session information - it will interact with Dex (making the user login) and set the appropriate session headers or cookie information for its endpoint. Here, Dex acts as a unifying connector to all these identification mechanisms and not a proxy.
[...]

PR #560, opened 2019-10-23
Introduces kfctl_k8s_istio_dex.yaml config

krishnadurai:
Related to #267 

This PR introduces a dex auth based kubernetes deployment for kubeflow.

Blocked on: #362 

/cc @johnugeorge @swiftdiaries

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/560)

### DIFF
--- a/kfdef/kfctl_k8s_istio_dex.yaml
+++ b/kfdef/kfctl_k8s_istio_dex.yaml
@@ -1,0 +1,375 @@
+# This is the config to install Kubeflow with Dex on an existing k8s cluster.
+# If the cluster already has istio, comment out the istio install part below.
+
+apiVersion: kfdef.apps.kubeflow.org/v1alpha1
+kind: KfDef
+metadata:
+  name: kubeflow_app
+  namespace: kubeflow
+spec:
+  repos:
+  - name: manifests
+    uri: https://github.com/kubeflow/manifests/archive/master.tar.gz
+  applications:
+  - kustomizeConfig:
+      repoRef:
+        name: manifests
+        path: application/application-crds
+    name: application-crds
+  - kustomizeConfig:
+      overlays:
+      - application
+      repoRef:
+        name: manifests
+        path: application/application
+    name: application
+  # Istio install. If not needed, comment out istio-crds and istio-install.
+  - kustomizeConfig:
+      parameters:
+      - name: namespace
+        value: istio-system
+      repoRef:
+        name: manifests
+        path: istio/istio-crds
+    name: istio-crds
+  - kustomizeConfig:
+      parameters:
+      - name: namespace
+        value: istio-system
+      repoRef:
+        name: manifests
+        path: istio/istio-install
+    name: istio-install
+  # This component is the istio resources for Kubeflow (e.g. gateway), not about installing istio.
+  - kustomizeConfig:
+      parameters:
+        - name: clusterRbacConfig
+          value: "ON"
+      repoRef:
+        name: manifests
+        path: istio/istio
+    name: istio
+  - kustomizeConfig:
+      parameters:
+        - name: namespace
+          value: cert-manager
+      repoRef:
+        name: manifests
+        path: cert-manager/cert-manager-crds
+    name: cert-manager-crds
+  - kustomizeConfig:
+      parameters:
+        - name: namespace
+          value: kube-system
+      repoRef:
+        name: manifests
+        path: cert-manager/cert-manager-kube-system-resources
+    name: cert-manager-kube-system-resources
+  - kustomizeConfig:
+      overlays:
+        - self-signed
+        - application
+      parameters:
+        - name: namespace
+          value: cert-manager
+      repoRef:
+        name: manifests
+        path: cert-manager/cert-manager
+    name: cert-manager
+  - kustomizeConfig:
+      overlays:
+        - application
+      parameters:
+        - name: namespace
+          value: istio-system
+        - name: userid-header
+          value: kubeflow-userid
+        - name: oidc_provider
+          value: http://dex.auth.svc.cluster.local:5556/dex
+        - name: oidc_redirect_uri
+          value: /login/oidc
+        - name: oidc_auth_url
+          value: /dex/auth
+        - name: skip_auth_uri
+          value: /dex
+        - name: client_id
+          value: kubeflow-oidc-authservice
+      repoRef:
+        name: manifests
+        path: istio/oidc-authservice
+    name: oidc-authservice
+  - kustomizeConfig:
+      overlays:
+        - istio
+      parameters:
+        - name: namespace
+          value: auth
+        - name: issuer
+          value: http://dex.auth.svc.cluster.local:5556/dex
+        - name: client_id
+          value: kubeflow-oidc-authservice
+        - name: oidc_redirect_uris
+          value: '["/login/oidc"]'
+        - name: static_email
+          value: admin@kubeflow.org
+        # Password is "password", 12-round bcrypt-hashed.
+        - name: static_password_hash
+          value: $2y$12$9Ukc61M.v83Nidxvcln9VeOtJRBBulSyH81hfnHzKAc53uZafn.6O
+      repoRef:
+        name: manifests
+        path: dex-auth/dex-crds
+    name: dex
+  - kustomizeConfig:
+      repoRef:
+        name: manifests
+        path: metacontroller
+    name: metacontroller
+  - kustomizeConfig:
+      overlays:
+      - istio
+      - application
+      repoRef:
+        name: manifests
+        path: argo
+    name: argo
+  - kustomizeConfig:
+      repoRef:
+        name: manifests
+        path: kubeflow-roles
+    name: kubeflow-roles
+  - kustomizeConfig:
+      overlays:
+      - istio
+      - application
+      parameters:
+      - name: userid-header
+        value: kubeflow-userid
+      repoRef:
+        name: manifests
+        path: common/centraldashboard
+    name: centraldashboard
+  - kustomizeConfig:
+      overlays:
+      - application
+      repoRef:
+        name: manifests
+        path: admission-webhook/bootstrap
+    name: bootstrap
+  - kustomizeConfig:
+      overlays:
+      - application
+      repoRef:
+        name: manifests
+        path: admission-webhook/webhook
+    name: webhook
+  - kustomizeConfig:
+      parameters:
+      - name: userid-header
+        value: kubeflow-userid
+      overlays:
+      - istio
+      - application
+      repoRef:
+        name: manifests
+        path: jupyter/jupyter-web-app
+    name: jupyter-web-app
+  - kustomizeConfig:
+      overlays:
+      - istio
+      repoRef:
+        name: manifests
+        path: metadata
+    name: metadata
+  - kustomizeConfig:
+      overlays:
+      - istio
+      - application
+      repoRef:
+        name: manifests
+        path: jupyter/notebook-controller
+    name: notebook-controller
+  - kustomizeConfig:
+      overlays:
+      - application
+      repoRef:
+        name: manifests
+        path: pytorch-job/pytorch-job-crds
+    name: pytorch-job-crds
+  - kustomizeConfig:
+      overlays:
+      - application
+      repoRef:
+        name: manifests
+        path: pytorch-job/pytorch-operator
+    name: pytorch-operator
+  - kustomizeConfig:
+      overlays:
+      - application
+      parameters:
+      - name: namespace
+        value: knative-serving
+      repoRef:
+        name: manifests
+        path: knative/knative-serving-crds
+    name: knative-crds
+  - kustomizeConfig:
+      overlays:
+      - application
+      parameters:
+      - name: namespace
+        value: knative-serving
+      repoRef:
+        name: manifests
+        path: knative/knative-serving-install
+    name: knative-install
+  - kustomizeConfig:
+      overlays:
+      - application
+      repoRef:
+        name: manifests
+        path: kfserving/kfserving-crds
+    name: kfserving-crds
+  - kustomizeConfig:
+      overlays:
+      - application
+      repoRef:
+        name: manifests
+        path: kfserving/kfserving-install
+    name: kfserving-install
+  - kustomizeConfig:
+      overlays:
+      - application
+      parameters:
+      - initRequired: true
+        name: usageId
+        value: <randomly-generated-id>
+      - initRequired: true
+        name: reportUsage
+        value: "true"
+      repoRef:
+        name: manifests
+        path: common/spartakus
+    name: spartakus
+  - kustomizeConfig:
+      overlays:
+      - istio
+      repoRef:
+        name: manifests
+        path: tensorboard
+    name: tensorboard
+  - kustomizeConfig:
+      overlays:
+      - application
+      repoRef:
+        name: manifests
+        path: tf-training/tf-job-crds
+    name: tf-job-crds
+  - kustomizeConfig:
+      overlays:
+      - application
+      repoRef:
+        name: manifests
+        path: tf-training/tf-job-operator
+    name: tf-job-operator
+  - kustomizeConfig:
+      repoRef:
+        name: manifests
+        path: katib/katib-crds
+    name: katib-crds
+  - kustomizeConfig:
+      overlays:
+      - application
+      - istio
+      repoRef:
+        name: manifests
+        path: katib/katib-controller
+    name: katib-controller
+  - kustomizeConfig:
+      overlays:
+      - application
+      repoRef:
+        name: manifests
+        path: pipeline/api-service
+    name: api-service
+  - kustomizeConfig:
+      overlays:
+      - application
+      parameters:
+      - name: minioPvcName
+        value: minio-pv-claim
+      repoRef:
+        name: manifests
+        path: pipeline/minio
+    name: minio
+  - kustomizeConfig:
+      overlays:
+      - application
+      parameters:
+      - name: mysqlPvcName
+        value: mysql-pv-claim
+      repoRef:
+        name: manifests
+        path: pipeline/mysql
+    name: mysql
+  - kustomizeConfig:
+      overlays:
+      - application
+      repoRef:
+        name: manifests
+        path: pipeline/persistent-agent
+    name: persistent-agent
+  - kustomizeConfig:
+      overlays:
+      - application
+      repoRef:
+        name: manifests
+        path: pipeline/pipelines-runner
+    name: pipelines-runner
+  - kustomizeConfig:
+      overlays:
+      - istio
+      - application
+      repoRef:
+        name: manifests
+        path: pipeline/pipelines-ui
+    name: pipelines-ui
+  - kustomizeConfig:
+      overlays:
+      - application
+      repoRef:
+        name: manifests
+        path: pipeline/pipelines-viewer
+    name: pipelines-viewer
+  - kustomizeConfig:
+      overlays:
+      - application
+      repoRef:
+        name: manifests
+        path: pipeline/scheduledworkflow
+    name: scheduledworkflow
+  - kustomizeConfig:
+      overlays:
+      - application
+      repoRef:
+        name: manifests
+        path: pipeline/pipeline-visualization-service
+    name: pipeline-visualization-service
+  - kustomizeConfig:
+      overlays:
+      - application
+      - istio
+      parameters:
+      - name: userid-header
+        value: kubeflow-userid
+      repoRef:
+        name: manifests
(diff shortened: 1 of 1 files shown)


yanniszark:
Hi @krishnadurai.

I took a look at the new config and I cannot spot any differences from the existing_arrikto config, apart from the KFServing installation, which is coming to existing_arrikto as well.

I don't think we need two identical configs, it will just confuse people. And we can work fast together if we want to accelerate the merge of KFServing support in existing_arrikto.

Is there something else, except KFServing, which I might have missed?

krishnadurai:
Hello @yanniszark,
With this configuration, we are looking to have a complete Dex Auth based Kubernetes installation. This will let us test the complete feature matrix of Kubeflow.

As far as the similarity is concerned, this config aims to be Dex based with all Kubernetes features included and as well as a few more components like [dex-authenticator](https://github.com/kubeflow/manifests/tree/master/dex-auth/dex-authenticator/base) and ldap(https://github.com/kubeflow/manifests/tree/master/dex-auth/dex-ldap/base). This is just one of the few things which might differ from the existing_arrikto config, as work on this config continues.

Likewise, I assumed the purpose of existing_arrikto config is an opinionated config from Arrikto - which we might not adopt for this purpose.

krishnadurai:
/hold

For reviews and a few PRs to get in: #362, #562

LeonardAukea:
What about renaming `existing_arrikto` to `kfctl_k8s_istio_dex` ?

yanniszark:
@krishnadurai thanks for your answer. Let me elaborate on that.

The whole purpose of Arrikto's config is to have a configuration that will let people install the complete feature matrix of Kubeflow, without depending on any cloud-specific features.

This is what we had argued when introducing it, and it seems that people indeed need this config and use it extensively with v0.6. This is why we had initially proposed for it to become a community config and not mark it with the "arrikto" wording. At that point, @jlewi pushed for using a name that declares who the arbitrator is and thus a vendor name.

So, we should either mark your proposed config with a "cisco" name, and then have two configs doing the same thing coming from two different vendors, which doesn't make much sense, or have one config, so that users don't get confused. From our side, we don't have a problem renaming the "existing_arrikto", if we all decide that this is the config which will always include the whole feature matrix without depending on any cloud-specific feature , so it is the "batteries included" config. This was our initial argument in the first place.

What do you think? cc: @jlewi

krishnadurai:
@yanniszark I appreciate your concern. As I've made it clear, we are looking for a config which can let us use the full feature surface for Kubeflow. Growingly, authentication and authorization are playing an important part of how Kubeflow is shaping up and we required the need of this config to be a part of this pre 1.0 release.

jlewi:
Regarding the name I believe my original thinking was 

* To name it based on what the config does; IIRC correctly I think I pushed for something like kfctl_dex_istio.yaml precisely to @LeonardAukea's point that we should name it based on what it does.
* IIRC the argument against a name like `kfctl_dex_istio.yaml` was that it was too narrow; overtime the config was expected to evolve to track additional opinions around what a recommended deployment of Kubeflow looks like
* At which point I suggested that if the config was intended to track Arrikto's opinions/recommendations about what a Kubeflow deployment should be then it would make sense to pick a name that indicate that it was a vendor, opinionated solution.

Taking a page from the Kubernetes playbook 
https://kubernetes.io/docs/setup/

I think we will have two categories of solutions in Kubeflow

1. "Turnkey" solutions 
   
     * These will probably be vendor specific/maintained
     * These will probably target more production ready environments

1. Deployments with vendor neutral tools (e.g. minikube, kops etc...)

    * These will likely be more vendor neutral and community maintained
    * These will likely be targeted more at people looking to kick the tires using free tools or advanced users (e.g. folks doing bare metal deployments)

I think the names should reflect that.

swiftdiaries:
💯 to @jlewi 's comment. `kfctl_k8s_istio` and `kfctl_k8s_dex_istio` are vendor neutral.

I was under the impression that the `existing_arritko` config gives Arritko the freedom to add components like Rok into the config (even though it doesn't today) and hence maintain it themselves.

kfdef/kfctl_k8s_istio_dex.yaml
```diff
@@ -0,0 +1,387 @@
...
+      repoRef:
+        name: manifests
+        path: cert-manager/cert-manager
+    name: cert-manager
+  - kustomizeConfig:
+      parameters:
+        - name: namespace
+          value: istio-system
+        - name: domain
+          value: DOMAIN #Replace with your kubeflow domain
```
krishnadurai: Replace this value with an actual domain or IP resolving to the NodePort or LoadBalancer of istio-ingressgateway service.

swiftdiaries: Can this be an environment variable that is set during invocation ?

kfdef/kfctl_k8s_istio_dex.yaml
```diff
@@ -0,0 +1,387 @@
...
+          value: http://dex.auth.svc.cluster.local:5556/dex
+        - name: client_id
+          value: kubeflow-oidc-authservice
+        - name: oidc_redirect_uris
+          value: '["/login/oidc"]'
+        - name: static_email
+          value: admin@kubeflow.org
+        # Password is "password", 12-round bcrypt-hashed.
+        - name: static_password_hash
+          value: $2y$12$9Ukc61M.v83Nidxvcln9VeOtJRBBulSyH81hfnHzKAc53uZafn.6O
```
swiftdiaries: Can this be an environment variable as well?

swiftdiaries:
Overall, looks good
Just a few minor comments on how we set the variables

swiftdiaries:
This is what I was thinking:

```
export KFAPP= some-name
export EMAIL=admin@kubeflow.dev
export PASSWORD=<encryted password>
export DOMAIN=<some-ip>.IstioNodePort
export CONFIG=dex-config.yaml
kfctl apply -V -f ${CONFIG}
```

yanniszark:
@swiftdiaries @jlewi thank you for your feedback.
 
 Let's discuss this on Tuesday's community meeting.
 I believe it's a matter of interest to all Kubeflow contributors.

krishnadurai:
> Overall, looks good
> Just a few minor comments on how we set the variables

Thanks for your review!

> This is what I was thinking:
> 
> ```
> export KFAPP= some-name
> export EMAIL=admin@kubeflow.dev
> export PASSWORD=<encryted password>
> export DOMAIN=<some-ip>.IstioNodePort
> export CONFIG=dex-config.yaml
> kfctl apply -V -f ${CONFIG}
> ```

@swiftdiaries I'm not sure if `kfctl` has the option to substitute ENV vars into KfDef right now. Can you suggest if there's an option available?
We might want to implement it otherwise for subsequent releases.

swiftdiaries:
@yanniszark 
Yes, it is a matter of importance to all Kubeflow users and contributors.
I strongly believe this is a step in the right direction.

These are the Kubernetes community values:
https://github.com/kubernetes/community/blob/master/values.md
> We are here as a community first, our allegiance is to the intentional stewardship of the Kubernetes project for the benefit of all its members and users everywhere.

It's important we share not only the API conventions from the Kubernetes community but also their values. As I personally believe, that is what made Kubernetes what it is today.

https://groups.google.com/forum/#!topic/knative-dev/KbrYbiF4XCs
> This project is bigger than any one company or individual.

jlewi:
@krishnadurai a KFDef can refer to secrets using environment variables. You can use this to avoid storing sensitive information in KFDef that shouldn't be checked into source control

krishnadurai:
@jlewi thanks for this useful information.
Though it looks like we would require a plugin implementation for either:
1. Modifying the application parameters in the KfDef through ENV vars
2. Setting these secrets from KfDef to K8s

The GCP based basic auth and IAP implementation use plugins to interpret environment variables and set these secrets.

For example:
https://github.com/kubeflow/kubeflow/blob/a992645fe88fe01eff5636cfccde9ae0607874a2/bootstrap/pkg/kfapp/gcp/gcp.go#L1946-L1965

Is there any method through which we could set these application parameters in the KfDef without a plugin?

Please correct me if my understanding is wrong.

jlewi:
@krishnadurai I believe that's correct; if you want to get secrets from the environment where kfctl is running and pass them into the Kubernetes cluster you'll need a plugin.

This PR though is just checking in a K8s config file?

@yanniszark and @krishnadurai is the difference between kfctl_k8s_istio_dex.yaml and kfctl_existing_arrikto.yaml just the name or are they doing something different in how they handle ingress and auth?

Can the code in [existing_arrikto.go](https://github.com/kubeflow/kubeflow/blob/master/bootstrap/pkg/kfapp/existing_arrikto/existing.go) be refactored into a a DEX plugin that could be composed with other plugins?

We (GCP) would like that because we'd like to be able to combine the dex plugin with the GCP plugin  in order to deploy dex on GKE cluster.

So I guess I have the following questions

1. Can we create a reusable/composable plugin for configuring DEX that can be used across configurations?
1. Are kfctl_k8s_istio_dex.yaml and kfctl_existing_arrikto.yaml functionally different or is it just a question of naming?
1. Do we need to have two separate configs kfctl_k8s_istio_dex.yaml and kfctl_k8s_istio.yaml?
    * In general fewer configs seems better just from a support load
    * Would it be sufficient to have a config that installs istio and dex and if people really don't want to install dex then we could just provide instructions on how to remove dex from the KFDef file?

krishnadurai:
> @krishnadurai I believe that's correct; if you want to get secrets from the environment where kfctl is running and pass them into the Kubernetes cluster you'll need a plugin.
> 
> This PR though is just checking in a K8s config file?

Yes this PR is for just the config file.

> @yanniszark and @krishnadurai is the difference between kfctl_k8s_istio_dex.yaml and kfctl_existing_arrikto.yaml just the name or are they doing something different in how they handle ingress and auth?

There is a minor difference in the way ingress is handled. In this file, I'm trying to add a SSL certificate to ingress gateway, issued on a provided 'DOMAIN'. kfctl_existing_arrikto has so far not given an option for ingress through it's KfDef.

> Can the code in [existing_arrikto.go](https://github.com/kubeflow/kubeflow/blob/master/bootstrap/pkg/kfapp/existing_arrikto/existing.go) be refactored into a a DEX plugin that could be composed with other plugins?
>
> We (GCP) would like that because we'd like to be able to combine the dex plugin with the GCP plugin in order to deploy dex on GKE cluster.

I'm looking to add a Dex plugin to do the same.
A question I have is should we look to follow the ingress back-end setup pattern for GCP? This will be similar to what IAP and Basic Auth setups have with cloud end-points and Kubernetes Ingress.
If yes, should GCP plugin handle it?

> So I guess I have the following questions
> 
> 1. Can we create a reusable/composable plugin for configuring DEX that can be used across configurations?

We should look to add a minimal Dex based plugin which does the following:

1. Optionally sets up staticPasswords configuration for a single user.
1. Fills in env var DOMAIN as a parameter in KfDef for issuing SSL certificate used by Istio ingress gateway service's tls secret. This has to be optional in case GCP chooses to have the cloud-endpoint based ingress pattern.

> 2. Are kfctl_k8s_istio_dex.yaml and kfctl_existing_arrikto.yaml functionally different or is it just a question of naming?

Right now it looks like both look to serve the same purpose. @yanniszark should be able to confirm.

> 3. Do we need to have two separate configs kfctl_k8s_istio_dex.yaml and kfctl_k8s_istio.yaml?
>    
>    * In general fewer configs seems better just from a support load
>    * Would it be sufficient to have a config that installs istio and dex and if people really don't want to install dex then we could just provide instructions on how to remove dex from the KFDef file?

I would prefer having a single config with Dex and then having instructions to disable it. 
One of the loose ends would be whether if we can be supporting no-auth based setups (without multi user) in tests without a config present in manifests/kfdef folder?

jlewi:
I have a couple questions; it would be good to capture them in your design doc: http://bit.ly/2kxbgbE


1. This might be an obvious question but is the ingress the [ISTIO Ingress](https://istio.io/docs/tasks/traffic-management/ingress/ingress-control/)

    * On GCP for example you could potentially either use the ISTIO ingress which I think would expose it using an L4 loadbalancer. Or you could do what we do with IAP which is expose the ISTIO gateway on a nodeport and use a GCP Ingress

1. Are you relying on the ISTIO ingress to do SSL termination? 

> There is a minor difference in the way ingress is handled. In this file, I'm trying to add a SSL certificate to ingress gateway, issued on a provided 'DOMAIN'. kfctl_existing_arrikto has so far not given an option for ingress through it's KfDef.

Could you explain how this would work?

i) Does the user have to use a domain registrar (e.g. Google Domains) to register a domain for the ip address?
ii) Would the user have to upload a cert for the domain or is the cert going to be generated for the domain using lets-encyrpt & cert manager?

I think right now kfctl_existing_arrikto.yaml supports exposing Kubeflow via IP address using a self-signed certificate. That's nice because it doesn't require the user to register a domain. Using a self-signed certificate should also be fast.

Do you also intend to support that mode?

> A question I have is should we look to follow the ingress back-end setup pattern for GCP? This will be similar to what IAP and Basic Auth setups have with cloud end-points and Kubernetes Ingress.
If yes, should GCP plugin handle it?

If there is some configuration that needs to be set that isn't known until runtime then using the pattern of running an agent in the cluster that will monitor it and then set appropriate fields is a good one.

e.g. if some configuration needs the IP address of the ingress which isn't known until after the ingress is created then one pattern is to run an agent in the cluster that would monitor the ingress and then create a configmap with the ip address once its available.

> I would prefer having a single config with Dex and then having instructions to disable it.

Agreed. Could we then modify and/or rename the existing kfctl_k8s_istio.yaml config?

krishnadurai:
@jlewi thanks for the information in your comments. I'll have this documented as suggested.

>     1. This might be an obvious question but is the ingress the [ISTIO Ingress](https://istio.io/docs/tasks/traffic-management/ingress/ingress-control/)

Yes, it is the Istio ingress.

>        * On GCP for example you could potentially either use the ISTIO ingress which I think would expose it using an L4 loadbalancer.

We could choose this option for now, IMO.

> Or you could do what we do with IAP which is expose the ISTIO gateway on a nodeport and use a GCP Ingress

And we should shortly follow up with a GCP Ingress solution, since this will not require the user to set up their own domain.

>     2. Are you relying on the ISTIO ingress to do SSL termination?

Yes, Istio ingress would do the SSL termination. This is required especially if the Istio ingress gateway service is exposed as an end-point.

> > There is a minor difference in the way ingress is handled. In this file, I'm trying to add a SSL certificate to ingress gateway, issued on a provided 'DOMAIN'. kfctl_existing_arrikto has so far not given an option for ingress through it's KfDef.
> 
> Could you explain how this would work?

The user input 'DOMAIN' will be required since I'm looking to issue an SSL certificate on the Isito ingress gateway service. Here are the cases:
1. DOMAIN is provided: Issue certificate on the domain (Irrespective of whether Istio ingress gateway service is of LoadBalancer or NodePort type).
1. DOMAIN is not provided and the user requires a load balancer configuration: The plugin will wait for Istio ingress gateway service to get its external IP and issue a self-signed certificate on the external IP of the Load Balancer.

Note: In case if the user does not want to use LoadBalancer and Serve with NodePort they would supply DOMAIN with a Node IP.

Regarding DOMAIN automation: This is platform-specific and I think we could configure this as a part of platform-specific plugins later.

> i) Does the user have to use a domain registrar (e.g. Google Domains) to register a domain for the ip address?

Yes, since this should be relatively easy to document users shouldn't be much affected during setup.

> ii) Would the user have to upload a cert for the domain or is the cert going to be generated for the domain using lets-encrypt & cert manager?

The cert creation for the DOMAIN would be automated using cert-manager's options: self-signed or lets-encrypt production.

> I think right now kfctl_existing_arrikto.yaml supports exposing Kubeflow via IP address using a self-signed certificate. That's nice because it doesn't require the user to register a domain. Using a self-signed certificate should also be fast.
> 
> Do you also intend to support that mode?

We should support this in case if the user does not provide a DOMAIN and wants to configure a LoadBalancer based setup.

> Agreed. Could we then modify and/or rename the existing kfctl_k8s_istio.yaml config?

Let's have this configuration replace kfctl_k8s_istio.yaml. The docs should reflect that.

Another question is how do we intend to test no-auth mode (config without multi-user) without a pre-baked KfDef? Should test configurations be separate from production ones?

jlewi:
@krishnadurai why would this configuration support exposing the istio ingress gateway via a node-port as opposed to a LoadBalancer?

Requiring that a user's Kubernetes cluster to provision an IP address (either public or on internal network given the Kubernetes install) for a Service of type LoadBalancer seems like a reasonable assumption.

What would be the use case for supporting NodePort? 

On GCP with IAP we support NodePort for ISTIO ingress gateway because we want to combine it with a GCP ingress resource.

But if we are creating a reference architecture then can we standardize on a Service of type LoadBalancer?

On GCP for example, this configuration is valuable precisely because its not using the GCP ingress because using an L4 loadbalancer (service type Loadbalancer) is much faster.

kfdef/kfctl_k8s_istio_dex.yaml
jlewi: +1 
password should be set via plugin based on environment variable

jlewi:
From a code review perspective; I think it would be valuable to modify the existing kfctl_k8s_istio.yaml config file. This way the diff would highlight whats changing about this PR.

jlewi:
Currently kfctl_k8s_istio.yaml is a configuration that works for everyone but is not identity aware and doesn't have a good story for how you access it from outside the cluster (kubectl port-forward works but isn't great).

Is the proposal to change that to

* Exposing the istio ingress gateway on an IP address (which could be either internal or public depending on the "Cloud" Provider's implementation of service type loadbalancer and possibly additional service annotations)
* Always including DEX for auth
  * Defaulting to basic auth with username and password set by environment variables
* Using a self sign certificate to do https termination 

Adding advanced options to

* Use LetsEncrypt to generate a signed certificate and/or upload a signed certificate
  * I don't think we want to use LetsEncrypt by default because it requires an external DNS name and public ip 
  * LetsEncrypt ACME challenge is also slow
* Specify a DOMAIN that would be registered in some external registrar
  * I don't think we want to require this by default because that requires users to perform another manual step in a cloud/infra dependent way

krishnadurai:
> From a code review perspective; I think it would be valuable to modify the existing kfctl_k8s_istio.yaml config file. This way the diff would highlight whats changing about this PR.

I'll make this change in this PR.

krishnadurai:
> @krishnadurai why would this configuration support exposing the istio ingress gateway via a node-port as opposed to a LoadBalancer?
> 
> Requiring that a user's Kubernetes cluster to provision an IP address (either public or on internal network given the Kubernetes install) for a Service of type LoadBalancer seems like a reasonable assumption.
> 
> What would be the use case for supporting NodePort?

@jlewi Here are some reasons why I would like to support NodePort configuration:

1. Most users tend to deploy Kubeflow in a minimal configuration of Kubernetes to test it out.
1. On-premise setups do not have a reference load balancer, likewise, OTS Kubernetes + Kubeflow could possibly work with NodePort. Otherwise, we introduce a necessary LoadBalancer installation for Kubeflow to work.

For both these cases, NodePort could be an off the shelf configuration.

In case if we do not want to support these use-cases

krishnadurai:
> Currently kfctl_k8s_istio.yaml is a configuration that works for everyone but is not identity aware and doesn't have a good story for how you access it from outside the cluster (kubectl port-forward works but isn't great).
> 
> Is the proposal to change that to
> 
>     * Exposing the istio ingress gateway on an IP address (which could be either internal or public depending on the "Cloud" Provider's implementation of service type loadbalancer and possibly additional service annotations)
> 
>     * Always including DEX for auth
>       
>       * Defaulting to basic auth with username and password set by environment variables
> 
>     * Using a self sign certificate to do https termination
> 
> 
> Adding advanced options to
> 
>     * Use LetsEncrypt to generate a signed certificate and/or upload a signed certificate
>       
>       * I don't think we want to use LetsEncrypt by default because it requires an external DNS name and public ip
>       * LetsEncrypt ACME challenge is also slow
> 
>     * Specify a DOMAIN that would be registered in some external registrar
>       
>       * I don't think we want to require this by default because that requires users to perform another manual step in a cloud/infra dependent way

This captures the proposal in a well-summarized manner.
The only thing I would add is that we could add comments to optionally disable dex based auth for install as discussed earlier.

yanniszark:
Following from the discussion around the existing_arrikto config and the agreement that the two configs converge, I have made the following issue and PR to address it as agreed:

Issue: https://github.com/kubeflow/manifests/issues/599
PR: https://github.com/kubeflow/manifests/pull/600

I believe we can close this PR or repurpose it as a PR that builds on the base added by https://github.com/kubeflow/manifests/pull/600.

/cc @krishnadurai @ddutta @jlewi @johnugeorge @vkoukis @cvenets

PR #573, opened 2019-10-24
reset backend name

### LINKED ISSUE #561: [GCP] iap-enabler won't pick up changes to the backend id in JWT policy
IAP enabler script will not pick up changes to the backend id.
https://github.com/kubeflow/manifests/blob/973fba5a2012e7c9e305092727b2da446a2b6b81/gcp/iap-ingress/base/config-map.yaml#L92

We don't reset backend id and backend name each time through the loop so if backend id or backend changes it won't get picked.

The behavior was recently changed #551 to periodically check and update the value

Related to  kubeflow/kubeflow#4342

lluunn:
Resolves #561

Reset backendName each time so that it will pick up new changes.

**Checklist:**
    1. `cd manifests/tests`
    2. `make generate`
    3. `make test`

/cc @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/573)

### DIFF
--- a/gcp/iap-ingress/base/config-map.yaml
+++ b/gcp/iap-ingress/base/config-map.yaml
@@ -81,6 +81,7 @@ data:
       NODE_PORT=$(kubectl --namespace=${NAMESPACE} get svc ${SERVICE} -o jsonpath='{.spec.ports[?(@.name=="http2")].nodePort}')
       echo "node port is ${NODE_PORT}"
 
+      BACKEND_NAME=""
       while [[ -z ${BACKEND_NAME} ]]; do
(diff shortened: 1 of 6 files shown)


jlewi:
@lluunn can you revert any tests not related to this change?

abhi-g:
/hold

Address @jlewi's comments

lluunn:
Those are result of make test. Isn't that supposed to be run, and maybe previous PR didn't run?

jlewi:
@lllunn `make generate` regenerates everything there's another make command that's only supposed to generate changed files but that never works for me.

I usually run `make generate` and then revert any files not related to the manifest that changed. The tests should still pass assuming HEAD isn't broken.

PR #576, opened 2019-10-25
Fix KFServing preventing notebook (and other pods) form starting.

jlewi:
* Related to kubeflow/kfserving#486
* The tag for the kfserving image, 0.2.0, this caused a breakage.
  With the newest image there is a bug and the kfserving mutator starts
  applying itself to non kfserving pods and ends up rejecting them.

* This change rolls back to an earlier image that was used yesterday in my
  deployment.

* We remove the $(registry) parameter from kustomization.yaml
  We don't want to treat registry as a parameter; this breaks the ability
  to use kustomize's ability to override images.

Resolves #

**Checklist:**
    1. `cd manifests/tests`
    2. `make generate`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/576)

### DIFF
--- a/gcp/deployment_manager_configs/cluster-kubeflow.yaml
+++ b/gcp/deployment_manager_configs/cluster-kubeflow.yaml
@@ -31,9 +31,7 @@ resources:
     zone: SET_THE_ZONE
     # "1.X": picks the highest valid patch+gke.N patch in the 1.X version
     # https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1/projects.zones.clusters
-    # TODO(https://github.com/kubeflow/kfctl/issues/48): This is a short term
-    # work around. We should unpin once 1.14.6-gke.14 is available.
-    cluster-version: "1.14.6-gke.2"
+    cluster-version: "1.14"
     # Set this to v1beta1 to use beta features such as private clusterss
     # and the Kubernetes stackdriver agents.
     gkeApiVersion: SET_GKE_API_VERSION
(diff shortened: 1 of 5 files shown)


jlewi:
Its the jupyter test that failed. 

Test cluster is in: us-central1-a

Looks like that zone doesn't have 1.14.7 available so we ended up using 1.14.6  which has the workload identity bug.

```
gcloud container get-server-config --zone=us-central1-a
Fetching server config for us-central1-a
defaultClusterVersion: 1.13.10-gke.0
defaultImageType: COS
validImageTypes:
- COS_CONTAINERD
- UBUNTU_CONTAINERD
- COS
- UBUNTU
validMasterVersions:
- 1.14.6-gke.13
...
```

jlewi:
Cluster is kfctl-938c

The notebook-test is failing because of the issue with kfserving.

```
create Pod jupyter-test-0 in StatefulSet jupyter-test failed error: Internal error occurred: admission webhook "inferenceservice.kfserving-webhook-server.pod-mutator" denied the request: quantities must match the regular expression '^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$'
```

Need to check whether its picking up this change.

jlewi:
So the problem is that the KFDef used in the presubmit is not pulling the manifests from the PR.

Here's the repo spec of KFDef.

```
repos:
 - {name: manifests, uri: 'https://github.com/kubeflow/manifests/archive/v0.7-branch.tar.gz'}
```

jlewi:
Here's the bug
https://github.com/kubeflow/kubeflow/blob/0a7155cedeff92797cb3b03ca1ac4e16a767f881/py/kubeflow/kubeflow/ci/kfctl_go_test_utils.py#L184

We should be checking the name of the repo not the URI when deciding whether to override it.

jlewi:
I manually verified that I could spin up a notebook manually using this change.

see here
https://github.com/kubeflow/manifests/pull/577#issuecomment-546538755

I ran this in a notebook.

I'm going to manually merge this to unblock some other fixes and not wait on fixing the test definition.

PR #583, opened 2019-10-28
Add self to appropriate OWNERS files

krishnadurai:
/assign @jlewi @kkasravi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/583)

### DIFF
--- a/cert-manager/OWNERS
+++ b/cert-manager/OWNERS
@@ -1,4 +1,4 @@
 approvers:
   - kkasravi
-  - yanniszark
   - krishnadurai
+  - yanniszark
--- a/cert-manager/OWNERS
+++ b/cert-manager/OWNERS
@@ -1,0 +1,4 @@
+approvers:
+  - kkasravi
+  - yanniszark
+  - krishnadurai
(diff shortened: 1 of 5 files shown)


cert-manager/OWNERS
```diff
@@ -0,0 +1,4 @@
+approvers:
+  - kkasravi
```
jlewi: alphabetical order please

dex-auth/OWNERS
```diff
@@ -0,0 +1,5 @@
+approvers:
+  - jlewi
```
jlewi: I shouldn't be listed as an explicit approver here since I don't have specific context.
Does the same hold for @kkasravi ?

Alphabetical order please.

istio/OWNERS
```diff
@@ -0,0 +1,3 @@
+approvers:
```
jlewi: I think @lluunn do most of the ISTIo work so I would list him and you as opposed to @kkasravi as an explicit approver.

krishnadurai:
@jlewi addressed your review comments.

istio/OWNERS
krishnadurai: Made this change.

cert-manager/OWNERS
krishnadurai: Sure

PR #584, opened 2019-10-29
Fix multiple overlays as title

Bobgy:
Fix multiple overlays as title

Looks like it should be like this. Needs confirm.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/584)


zhenghuiwang:
/lgtm
/approve
thanks!

PR #589, opened 2019-10-30
update central dashboard image

gabrielwen:
Related: kubeflow/kubeflow#4415

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/589)

### DIFF
--- a/common/centraldashboard/base/kustomization.yaml
+++ b/common/centraldashboard/base/kustomization.yaml
@@ -13,7 +13,7 @@ commonLabels:
   kustomize.component: centraldashboard
 images:
 - name: gcr.io/kubeflow-images-public/centraldashboard
-  newName: gcr.io/kubeflow-images-public/centraldashboard:list-8-gf5250b01
+  newName: gcr.io/kubeflow-images-public/centraldashboard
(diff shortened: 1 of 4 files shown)


common/centraldashboard/base/kustomization.yaml
```diff
@@ -13,8 +13,8 @@ commonLabels:
   kustomize.component: centraldashboard
 images:
 - name: gcr.io/kubeflow-images-public/centraldashboard
-  newName: gcr.io/kubeflow-images-public/centraldashboard
-  digest: sha256:bcf462b61b7f39e1479c4ac6dd0eb5a1b2504f2d2e18e325c6e8404a28f09a60
+  newName: gcr.io/kubeflow-images-public/centraldashboard:list-8-gf5250b01
```
jlewi: Why are you adding the tag ":list-8-gf5250b01"; this should just get replaced by the digest. So I think its better to drop the tag. This is less confusing IMO.

gabrielwen: removed.

jlewi:
The test failure is katib: kubeflow/testing#508.

This is unrelated to this PR so I'm going to merge manually to unblock the release.

PR #592, opened 2019-10-31
update serviceaccount

asdfsx:
In `cluster-role-binding.yaml`, cluster-role `argo` is binding to serviceaccount `argo` under namespace`kubeflow`. 
So I think serviceaccount `argo` should in namespace `kubeflow`.

https://github.com/kubeflow/manifests/blob/127dacce625e5dfc7630302ea82cf8ad56f9b6dd/argo/base/cluster-role-binding.yaml#L2-L15

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/592)

### DIFF
--- a/argo/base/service-account.yaml
+++ b/argo/base/service-account.yaml
@@ -3,6 +3,7 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: argo
+  namespace: kubeflow
 ---
 apiVersion: v1
 kind: ServiceAccount


argo/base/service-account.yaml
```diff
@@ -3,6 +3,7 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: argo
+  namespace: kubeflow
```
krishnadurai: @asdfsx, the metadata/namespace field is inherited from the kustomization.go code in kfctl. When the manifests for argo are created during kfctl's `build` stage, the namespace field is added to metadata.

krishnadurai:
@asdfsx can we close this PR?

PR #594, opened 2019-10-31
Fix ml-pipeline-ui doesn't have gcp permission bug

Bobgy:
Related to https://github.com/kubeflow/pipelines/issues/2501

Mount gcp service account credentials to ml-pipeline-ui.

**Checklist:**
    1. `cd manifests/tests`
    2. `make generate`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/594)

### DIFF
--- a/pipeline/pipelines-ui/base/deployment.yaml
+++ b/pipeline/pipelines-ui/base/deployment.yaml
@@ -13,21 +13,10 @@ spec:
       labels:
         app: ml-pipeline-ui
     spec:
-      volumes:
-      - name: gcp-sa-token
-        secret:
-          secretName: user-gcp-sa
       containers:
       - name: ml-pipeline-ui
-        env:
-        - name: GOOGLE_APPLICATION_CREDENTIALS
-          value: /etc/credentials/user-gcp-sa.json
         image: gcr.io/ml-pipeline/frontend
         imagePullPolicy: IfNotPresent
         ports:
         - containerPort: 3000
-        volumeMounts:
-        - name: gcp-sa-token
-          mountPath: "/etc/credentials"
-          readOnly: true
       serviceAccountName: ml-pipeline-ui
--- a/pipeline/pipelines-ui/base/deployment.yaml
+++ b/pipeline/pipelines-ui/base/deployment.yaml
@@ -13,10 +13,21 @@ spec:
       labels:
         app: ml-pipeline-ui
     spec:
+      volumes:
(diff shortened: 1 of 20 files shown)


jlewi:
Is this intended to be a temporary fix or a long term fix?

See https://github.com/kubeflow/pipelines/issues/2501#issuecomment-548575207

pipeline/pipelines-ui/base/deployment.yaml
```diff
@@ -13,10 +13,21 @@ spec:
       labels:
         app: ml-pipeline-ui
     spec:
+      volumes:
```
jlewi: This would also need to be a GCP overlay which means you would also need to modify 
https://github.com/kubeflow/manifests/blob/127dacce625e5dfc7630302ea82cf8ad56f9b6dd/kfdef/kfctl_gcp_iap.yaml#L278

and
https://github.com/kubeflow/manifests/blob/127dacce625e5dfc7630302ea82cf8ad56f9b6dd/kfdef/kfctl_gcp_basic_auth.yaml#L270

jlewi:
@Bobgy I think relying on GCP secrets is fine as a quick fix but we should file a follow on bug to switch to workload identity before the next release.

pipeline/pipelines-ui/base/deployment.yaml
Bobgy: Thanks! Made requested changes.

Bobgy:
It seems my generated tests don't match current format. Is that a `gofmt` version mismatch?

Bobgy:
I tried to use go1.12.4, but it doesn't change anything.

Besides, the tests are failing. I couldn't really understand what the error message means. Can you help me with that?

Bobgy:
This will be a quick fix. For long term fix, tracked in https://github.com/kubeflow/pipelines/issues/1691#issuecomment-548697785

jlewi:
@Bobgy The E2E tests are failing because of an issue with GCE. See: kubeflow/kfctl/issues/69

Please note that after this fix is submitted we will need to cherry pick the PR onto the v0.7-branch.

We have a shell script to easily create cherry picks but you can also do it manually
https://github.com/kubeflow/kubeflow/blob/master/hack/cherry-picks.md

jlewi:
@Bobgy the test failed because of an issue that's been fixed on master.
Could you please rebase on the latest master?

PR #631, opened 2019-11-26
moving kfserving to 0.2.2

animeshsingh:
Moves kfserving to 0.2.2

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/631)


animeshsingh:
/assign @johnugeorge 

cc @jlewi @ellis-bigelow

yuzisun:
I think there are other changes like logger in the crd not just version bump, we need a better way to sync manifests in kfserving to here.

animeshsingh:
@yuzisun is the CRD pulled from kfserving to here? Or manually crafted?
https://github.com/kubeflow/manifests/pull/631/files#diff-7e13d671ef822b8dc0ccde821228d780

animeshsingh:
Config map needs to be updated from here once merged
https://github.com/kubeflow/kfserving/pull/579/files#diff-c8dbe0c491a054c35d4254b9dfc6a6dd

yuzisun:
> @yuzisun is the CRD pulled from kfserving to here? Or manually crafted?
> https://github.com/kubeflow/kfserving/blob/master/config/default/crds/serving_v1alpha2_inferenceservice.yaml

You can not pull from config there because it does some patch later on with kustomize. Unfortunately for now you need to manually pull the final crd out from the 0.2.2 release yaml(https://github.com/kubeflow/kfserving/blob/master/install/0.2.2/kfserving.yaml), we need a script to automate this process, manual update here is a bit error prone.

yuzisun:
@animeshsingh  most importantly we need to add the namespace selector environment variable on kfserving controller statefulset manifest @hougangliu introduced :)

animeshsingh:
Ok - seems the tests are looking good now + accounted for namespace label selector change.

yuzisun:
awesome! looks good now!
/lgtm

animeshsingh:
@krishnadurai or @kkasravi this one will need attention

kfserving/kfserving-install/base/statefulset.yaml
```diff
@@ -39,7 +39,9 @@ spec:
               fieldPath: metadata.namespace
         - name: SECRET_NAME
           value: kfserving-webhook-server-secret
-        image: $(registry)/kfserving-controller:0.2.0
+        - name: ENABLE_WEBHOOK_NAMESPACE_SELECTOR
+          value: enabled
```
krishnadurai: Do we want this value to be configured by KfDef?

animeshsingh: yeah - enables kfserving deployment in kubeflow to be independent of standalone kfserving behaviour

yuzisun: with kubeflow installation I think this should be always be on otherwise user is taking a risk. Once we migrate to object selector with kuberentes 1.15+ then this env can be retired.

PR #638, opened 2019-12-04
Allow argo to delete workflows with spec.ttlSecondsAfterFinished

consideRatio:
## Issue

Argo workflow specifications with `spec.ttlSecondsAfterFinished` will be
deleted by the workflow-controller deployment. But without the proper
permissions for its service account it won't be able to do this.

If the `workflow-controller` find a `workflow` resource with
`spec.ttlSecondsAfterFinished` set without proper permission to delete
it when time comes, something like this will instead be seen in its logs.

```
E1204 14:39:54.926727       1 ttlcontroller.go:124] error deleting 'kubeflow/redactedailyrunhwn89-5-3408726899': workflows.argoproj.io "redacteddailyrunhwn89-5-3408726899" is forbidden: User "system:serviceaccount:kubeflow:argo" cannot delete resource "workflows" in API group "argoproj.io" in the namespace "kubeflow"
```

## Solution

I added the `delete` verb to the `argo` ClusterRole's rule for working with workflow resources, as that allows the workflow resources with `spec.ttlSecondsAfterFinished` to be deleted by the `workflow-controller`. The coupling of this permissions is through the `argo` ServiceAccount used by the workflow-controller. This ServiceAccount is bound to the `argo` ClusterRole through the `argo` ClusterRoleBinding.

## Quickfix

If you find your way here and don't want to wait for a fix, you could
quickfix it by editing the argo clusterrole with:

```shell
kubectl edit clusterrole argo
```

or using a not super reliable patch that may fail in the future for
other versions that KubeFlow 0.7.0 that this worked on:

```shell
kubectl patch clusterrole argo --type=json --patch='[{"op": "add", "path": "/rules/3/verbs/-", "value": "delete"}]'
```

## PR Checklist
    1. `cd manifests/tests`
    2. `make generate`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/638)

### DIFF
--- a/argo/base/cluster-role.yaml
+++ b/argo/base/cluster-role.yaml
@@ -43,6 +43,7 @@ rules:
   - watch
   - update
   - patch
+  - delete
 ---
 apiVersion: rbac.authorization.k8s.io/v1beta1
 kind: ClusterRole
--- a/tests/argo-base_test.go
+++ b/tests/argo-base_test.go
@@ -91,6 +91,7 @@ rules:
   - watch
   - update
   - patch
+  - delete
 ---
 apiVersion: rbac.authorization.k8s.io/v1beta1
 kind: ClusterRole
--- a/tests/argo-overlays-application_test.go
+++ b/tests/argo-overlays-application_test.go
@@ -146,6 +146,7 @@ rules:
   - watch
   - update
   - patch
+  - delete
 ---
 apiVersion: rbac.authorization.k8s.io/v1beta1
 kind: ClusterRole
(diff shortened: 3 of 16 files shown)


consideRatio:
@googlebot I fixed it.

NikeNano:
@googlebot I consent.

consideRatio:
@krishnadurai i opened up a bigger PR when i learned that this was only the tip of the iceberg. Consider #642 before merge

krishnadurai:
@consideRatio At first, thanks for contributing! Nice to read a descriptive PR outline.

#642 seems interesting. I'll be taking a look at it soon. Going ahead and assigning myself.

krishnadurai:
/hold

This for #642 to be considered first

PR #653, opened 2019-12-09
Create overlay allowing the Load Balancer to be created with HTTPS wi…

### LINKED ISSUE #652: HTTPS for AWS ingress without auth
Create overlay for the **aws-alb-ingress-controller** allowing the Load Balancer to be created with HTTPS with it performing no auth. 

This is one set of changes to switch from using cognito on the ALB to using the oidc-authservice because in the current setup the users come through without any identity (anonymous). 
See https://github.com/devgrok/kubeflow-manifests/blob/aws-istio-oidc-cognito/kfdef/kfctl_aws_cognito.yaml for the desired final result.

devgrok:
Resolves #652 

* Create overlay for the aws-alb-ingress-controller allowing the Load Balancer to be created with HTTPS with it performing no auth.
* Additionally making the scheme (internet facing/internal) configurable and adding external-dns annotation for clusters with it installed. 
* Cleaned up some duplicated annotations as well between the different overlays.
I couldn't get the tests to do it, but the oidc & cognito layers could piggy back off the secure one... though that may be more complex than the code it saves.

**Checklist:**
    1. `cd manifests/tests`
    2. `make generate`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/653)


Jeffwan:
@animeshsingh Can you help approve this change?

Jeffwan:
em.. Looks like it's Junit test of katib, not related..

/test kubeflow-manifests-presubmit

jlewi:
Merging manually since test failures are unrelated and should be fixed on master.

PR #633, opened 2019-11-29
istio: oidc-authservice: add readiness probe and update image

yanniszark:
Signed-off-by: Yannis Zarkadas <yanniszark@arrikto.com>

**Which issue is resolved by this Pull Request:**
Resolves https://github.com/kubeflow/manifests/issues/632
Hopefully a fix for https://github.com/arrikto/oidc-authservice/issues/2

**Description of your changes:**
Adds a readiness check to the OIDC authservice and updates image to new version.

**Checklist:**
- [X] Unit tests have been rebuilt: 
    1. `cd manifests/tests`
    2. `make generate`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/633)

PR #622, opened 2019-11-15
Add storage overlays for IBM NFS

Tomcli:
**Which issue is resolved by this Pull Request:**
Related issue: https://github.com/kubeflow/katib/issues/725

**Description of your changes:**
The IBM Cloud NFS(Network File System) on IBM's Kubernetes service is different from regular File System storage, and we encounter several issues with some of the mysql and oidc storage setup. Therefore, we created some overlays to config to the settings that can run on IBM NFS. 

cc: @animeshsingh 

**Checklist:**
- [x] Unit tests have been rebuilt: 
    1. `cd manifests/tests`
    2. `make generate`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/622)

PR #666, opened 2019-12-12
Adding self to OWNERS

krishnadurai:
/assign @kkasravi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/666)

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -7,11 +7,11 @@ approvers:
   - IronPan
   - jlewi
   - kkasravi
+  - krishnadurai
   - kunmingg
   - lluunn
   - richardsliu
   - swiftdiaries
   - yanniszark
   - zhenghuiwang
-reviewers:
-  - krishnadurai
+


jlewi:
@krishnadurai Instead of adding yourself to root OWNERS how about adding yourself to the packages you should be an admin/owner of? 

This will help better assign reviewers.

krishnadurai:
@jlewi I'm doing this with a view that most PRs are now spanning across manifests and having a sub-directory approve right isn't helping to push PRs forward.

jlewi:
@krishnadurai Could you provide an example of PRs spanning across directories?

My expectation would be that the bulk of the PRs would be affecting individual applications (e.g Jupyter vs. KFServing). Each of those applications would have different owners and PRs should be reviewed by the respective owners.

krishnadurai:
@jlewi Here are a few PRs I faced a problem in.
#622 #633

jlewi:
Please open a new PR if this is still relevant.

PR #661, opened 2019-12-12
Adds necessary roles and roleBinding to Spark service account

krishnadurai:
**Which issue is resolved by this Pull Request:**
Spark Operator's example SparkApplication CR fails with an error failing to list or create pods with the service account ~~`spark-sa`.~~

Updates:
- Renames the service account to `spark` as mentioned in spark-operator examples.

**Description of your changes:**
This PR adds the necessary role and roleBinding to `spark-sa`

**Checklist:**
- [x] Unit tests have been rebuilt: 
    1. `cd manifests/tests`
    2. `make generate`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/661)

PR #674, opened 2019-12-15
Update istio_dex config to support Spark operator

krishnadurai:
/assign @yanniszark 

I've tested spark operator and I think we could include it in this configuration now.
#661 

/cc @holdenk

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/674)


krishnadurai:
@yanniszark could you PTAL?

krishnadurai:
@yanniszark any update here?

PR #689, opened 2020-01-02
metadata: Adding metadata-grpc config map

dushyanthsc:
This change adds the necessary config-map related to gRPC MLMD server.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/689)


tests/metadata-overlays-application_test.go
```diff
@@ -126,9 +126,12 @@ spec:
...
+          envFrom:
+          - configMapRef:
+              name: metadata-grpc-configmap
           image: gcr.io/tfx-oss-public/ml_metadata_store_server:0.15.1
           command: ["/bin/metadata_store_server"]
-          args: ["--grpc_port=8080"]
+          args: ["--grpc_port=$(METADATA_GRPC_SERVICE_PORT)"]
           ports:
             - name: grpc-backendapi
               containerPort: 8080
```
zhenghuiwang: I think you need to change `containerPort: 8080` to use `$(METADATA_GRPC_SERVICE_PORT)` as well

zhenghuiwang:
I tested this change in my KF cluster with `METADATA_GRPC_SERVICE_PORT=9090`,  I still see the service port is 8080. I think you also need to change the port in `base/metadata-service.yaml`:

```
kind: Service
apiVersion: v1
metadata:
  labels:
    app: grpc-metadata
  name: grpc-service
spec:
  selector:
    component: grpc-server
  type: ClusterIP
  ports:
    - port: 8080                         <-------------- needs to use new param
      protocol: TCP
      name: grpc-backendapi
```

tests/metadata-overlays-application_test.go
dushyanthsc: As per offline discussion, added the necessary comment.

PR #700, opened 2020-01-09
Add sidecar.istio.io/inject false label to all deployment/stateful-sets in kubeflow namespace

Bobgy:
Part of https://github.com/kubeflow/pipelines/issues/1223

Add sidecar.istio.io/inject: "false" to all deployments to unblock https://github.com/kubeflow/kfctl/pull/131
1. changed each deployment/statefulset to add annotation sidecar.istio.io/inject "false"
2. updated https://github.com/kubeflow/manifests/blob/master/kfdef/kfctl_gcp_iap.yaml to use the new overlays

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/700)

### DIFF
--- a/jupyter/jupyter-web-app/base/kustomization.yaml
+++ b/jupyter/jupyter-web-app/base/kustomization.yaml
@@ -17,7 +17,7 @@ commonLabels:
 images:
 - name: gcr.io/kubeflow-images-public/jupyter-web-app
   newName: gcr.io/kubeflow-images-public/jupyter-web-app
-  newTag: vmaster-ge2eb2540
+  newTag: vmaster-ge8c85064
 configMapGenerator:
 - env: params.env
   name: parameters
--- a/tests/jupyter-jupyter-web-app-base_test.go
+++ b/tests/jupyter-jupyter-web-app-base_test.go
@@ -370,7 +370,7 @@ commonLabels:
 images:
 - name: gcr.io/kubeflow-images-public/jupyter-web-app
   newName: gcr.io/kubeflow-images-public/jupyter-web-app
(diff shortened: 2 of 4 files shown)


jlewi:
@Bobgy Thanks.

* Can we avoid creating new overlays? It looks like you are creating an ISTIO overlay for some applications that don't already have them.

   * The downside of doing this is that means the overlay needs to be added to all KFDef specs. Your PR currently only updates the GCP spec.

* Instead of adding the annotation in the ISTIO overlay why not add it in the base package? Adding the annotation "sidecar.istio.io/inject: "false"" should be harmless in the case ISTIO isn't installed. 
  
  * The ISTIO overlays are kind of a historical artifact at this point due to the transition from non-istio to ISTIO so I'm not sure we should keep adding to them.

* Rather than iterate on all of the applications at once; it might be faster to split off one or two applications from this PR and do those first and then once we've figured everything out update all the rest to match.

Bobgy:
@jlewi Thanks, I didn't know every deployment uses istio now. I will adjust that accordingly using one component as an example. Let's move on after that change is good.

Bobgy:
@jlewi made example PR in https://github.com/kubeflow/manifests/pull/703

Bobgy:
/hold
wait for discussion in https://github.com/kubeflow/manifests/pull/703 first

jlewi:
@Bobgy looks like #703 was merged do you want to update this?

Bobgy:
Made some mistake, will open a new PR instead

PR #703, opened 2020-01-14
Two examples for disabling istio injection in kubeflow namespace

Bobgy:
Example of https://github.com/kubeflow/manifests/pull/700

Add istio disable injection annotations for two deployments.

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/703)

### DIFF
--- a/argo/base/deployment.yaml
+++ b/argo/base/deployment.yaml
@@ -23,6 +23,8 @@ spec:
       creationTimestamp: null
       labels:
         app: argo-ui
+      annotations:
+        sidecar.istio.io/inject: "false"
     spec:
       containers:
       - env:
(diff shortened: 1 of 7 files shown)


Bobgy:
One problem I realized though, is that, pods generated by all the services also need such labels. It is not an easy task to adjust every service to let them work in Kubeflow.

I'd still prefer changing cluster default to not inject pods with sidecars.

jlewi:
> One problem I realized though, is that, pods generated by all the services also need such annotations. It is not an easy task to adjust every service to let them work in Kubeflow.

What's the difficulty?

/lgtm
/approve

Bobgy:
@jlewi Not exactly difficult, but it's a lot of work to make sure each operator spawn pod has an annotation like argo. You should have more context on this. If there isn't many, it's fine.

PR #725, opened 2020-01-22
Update the command for the kfam deployment in the manifest

### LINKED ISSUE #724: kfam manifest needs to be updated to work with the new distroless image
Updating the kfam image is failing in kubeflow/manifests#707 with the error

```
"Error: failed to start container "kfam": Error response from daemon: OCI runtime create failed: container_linux.go:345: starting container process caused "exec: \"/opt/kubeflow/access-management\": stat /opt/kubeflow/access-management: no such file or directory": unknown"
```

The problem is 
kubeflow/kubeflow#4625

changed the path of the binary when we switched to a distroless image

jlewi:
* In kubeflow/kubeflow#707 the path of the binary in the docker image
  was changed so we need to update the command in the manifest.

Fix #724

Resolves #

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/725)

### DIFF
--- a/profiles/base/deployment.yaml
+++ b/profiles/base/deployment.yaml
@@ -20,7 +20,7 @@ spec:
         imagePullPolicy: Always
         name: manager
       - command:
-        - /opt/kubeflow/access-management
+        - /access-management
         args:
         - "-cluster-admin"
         - $(admin)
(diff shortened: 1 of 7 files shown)


jlewi:
I suspect the test failed because we updated the command to use "/access_management" but we didn't also update the docker image to a new image for that command.

kunmingg:
@jlewi 
Covered in https://github.com/kubeflow/manifests/pull/734

jlewi:
Thanks Kunming. 

Closing in favor of #734

PR #733, opened 2020-01-22
add cadvisor manifests

kunmingg:
Add cadvisor daemonset which collect system metrics of all pods.

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/733)


jlewi:
@kunmingg Can you explain why cadvisor is needed? On GKE aren't these metrics already collected by stackdriver?

PR #778, opened 2020-01-29
Use kustomize to make it easier to maintain the versioned KFDef specs.

jlewi:
* For each KF release we need to define a KFDef spec that overrides certain
  values (e.g. the repo of kubeflow/manifests it uses)

* Previously we just did this by modifying the KFDef specs on the release
  branch. But this was very costly to maintain; i.e. backporting changes
  on master to the release branch

* To make that easier we can generate the KFDef YAML files using kustomize;
  this allows us to use overlays to define the changes needed to customize
  the specs for a particular version

* We can keep these versioned overlays on master so that the divergence between
  master and the branches is very low

* To preserve existing behavior we still check in YAML files. a simple
  script build_kfdef_specs.py is provided to generate them.

Related to: kubeflow/kubeflow#4685

Some History:

@kkasravi originally had the idea of using kustomize to help maintain our KFDef specs. Originally the logic for generating KFDef via kustomize was baked into kfctl. We decided that was overly complicated and that following a shift left philosophy we would move the generation of the KFDef (via kustomize) to prior to invoking kfctl. That is consistent with what this PR is doing.

Furthermore, in the original use of Kustomize to generate KFDef; overlays were used to generate KFDef specs for different clouds e.g. IBM, vs. GCP, etc...

In this PR overlays are instead used to generate KFDefs for different KF versions.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/778)

### DIFF
--- a/kfdef/source/README.md
+++ b/kfdef/source/README.md
@@ -1,0 +1,3 @@
+This directory contains kustomization packages that are used to generate the YAML specs for the KFDef.
+
+The script `hack/build_kfdef_specs.py` is used to run kustomize and output the YAML files.
--- a/kfdef/source/README.md
+++ b/kfdef/source/README.md
@@ -1,3 +1,16 @@
 This directory contains kustomization packages that are used to generate the YAML specs for the KFDef.
 
 The script `hack/build_kfdef_specs.py` is used to run kustomize and output the YAML files.
+
+## Subdirectories
+
+Each sub-directory corresponds to a kustomize package corresponding to a different release
+of Kubeflow.
+
+* **master**: This is the base kustomization package
+  * In general when adding new applications or making other KFDef specs that should be carried throughout
+    future versions you would make here
+
+* **vX.Y.Z**: This is the kustomization package for Kubeflow release x.y.x. It will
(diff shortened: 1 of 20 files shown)


krishnadurai:
>     * We can keep these versioned overlays on master so that the divergence between
>       master and the branches is very low

@jlewi does this mean that we will be maintaining only a few versions of KFDefs in master ad their respective overlays?

jlewi:
@krishnadurai We can GC the overlays on master when it makes sense. Most likely this will happen when the release branch and the master branch diverge so much that we no longer want to backport changes from master to the release branch.

Does that answer your question?

kfdef/source/README.md
```diff
@@ -0,0 +1,3 @@
+This directory contains kustomization packages that are used to generate the YAML specs for the KFDef.
+
+The script `hack/build_kfdef_specs.py` is used to run kustomize and output the YAML files.
```
richardsliu: Maybe add an explanation for the new directory structure here.

krishnadurai:
@jlewi this makes sense.

I think we could follow the same approach for all configs: istio_dex and k8s_istio.

Issue #794, opened 2020-01-29
Latest tf-job-operator image doesn't work with kustomize manifest; binary is wrong


jlewi:
Here's the PR attempting to update the docker image.

https://github.com/kubeflow/manifests/pull/787#issuecomment-579608145 error is.
```
INFO|2020-01-28T21:58:34|/home/jlewi/git_kubeflow-testing/py/kubeflow/testing/troubleshoot_deployment.py|50| Found event Created container tf-job-operator
INFO|2020-01-28T21:58:34|/home/jlewi/git_kubeflow-testing/py/kubeflow/testing/troubleshoot_deployment.py|50| Found event Error: failed to start container "tf-job-operator": Error response from daemon: OCI runtime create failed: container_linux.go:345: starting container process caused "exec: \"/opt/kubeflow/tf-operator.v1\": stat /opt/kubeflow/tf-operator.v1: no such file or directory": unknown
```

jlewi:
See here
https://github.com/kubeflow/tf-operator/blob/92389064014d3e661e1a69cb9a33cb996e493e10/build/images/tf_operator/Dockerfile#L17

jlewi:
#795 will need to be cherry picked onto the release branch.

jlewi:
kubeflow/manifests#810 is updating to auto-built image.

jlewi:
#810 was merged. So we just need to verify its working on the 1.0 branch.

jlewi:
I deployed from the 1.0 branch. The operator was using image
 image: gcr.io/kubeflow-images-public/tf_operator:v1.0.0-g92389064

And it started up fine.

Logs

```
{"filename":"app/server.go:76","level":"info","msg":"EnvKubeflowNamespace not set, use default namespace","time":"2020-01-31T00:09:49Z"}
{"filename":"app/server.go:80","level":"info","msg":"Using cluster scoped operator","time":"2020-01-31T00:09:49Z"}
{"filename":"app/server.go:86","level":"info","msg":"[API Version: v1 Version: v0.1.0-alpha Git SHA: Not provided. Go Version: go1.13.5 Go OS/Arch: linux/amd64]","time":"2020-01-31T00:09:49Z"}
{"filename":"tf-operator.v1/main.go:42","level":"info","msg":"Setting up client for monitoring on port: 8443","time":"2020-01-31T00:09:49Z"}
[...]

PR #795, opened 2020-01-29
Update TFJob manifest to work with new distroless image.

jlewi:
* The path in the image changed.
* Switch to using args so the manifest is more independent of the binary
  location.

* Related to kubeflow/manifests#794

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/795)


jlewi:
@krishnadurai could you take another look please I forgot to update the tests?

PR #801, opened 2020-01-29
Fix autoprovisioning settings.

jlewi:
* kubeflow/kubeflow#4259 GKE now allows setting oauthscopes and VM service
  account for NAP pools. We need to use this otherwise private images
  won't be accessible.

* kubeflow/kubeflow#3930 Bump NAP resource constraints because these
  are global constraints not per node constraints.

  * Bump max CPU to 128 CPUs
  * Bump memory

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/801)

### DIFF
--- a/gcp/deployment_manager_configs/cluster-kubeflow.yaml
+++ b/gcp/deployment_manager_configs/cluster-kubeflow.yaml
@@ -70,11 +70,12 @@ resources:
     # This is configured by the gkeApiVersion setting.
     autoprovisioning-config:
       enabled: true
-      max-cpu: 20
-      max-memory: 200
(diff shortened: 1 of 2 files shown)


jlewi:
Test failed

```
util.py                     72 INFO     Error: failed to apply:  (kubeflow.error): Code 500 with message: kfApp Apply failed for kustomize:
 (kubeflow.error): Code 500 with message: Apply.Run  Error error when creating "/tmp/kout945535893": Internal error occurred: failed calling
 webhook "webhook.cert-manager.io": the server is currently unable to handle the request
```

Nominally looks like a flake unrelated to GCP node auto-provisioner settings.

/test all

PR #805, opened 2020-01-29
Add pdmack to OWNERS

pdmack:
Review pending OpenShift changes

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/805)

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -9,6 +9,7 @@ approvers:
   - kkasravi
   - kunmingg
   - lluunn
+  - pdmack
   - richardsliu
   - swiftdiaries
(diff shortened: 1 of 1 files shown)


jlewi:
@pdmack Are the openshift related changes in specific directories? It's generally better if we grow owners files lower down in the directory tree rather than at the root level.

This helps the auto reviewer assigner better assign reviewers. 

We might want to start creating subdirectories in
kfdef
e.g.

```
kfdef/ibm
kfdef/gcp
kfdef/aws
```
Each with their own OWNERs file

/cc @animeshsingh

PR #816, opened 2020-01-30
Address #4699 RBAC permissions for Jupyter web app

jlewi:
* Related to #4699: Notebook web server gives error unable to list events

* This error occurs because the notebook UI does a subject access review
  on the user. As a result, the user needs to be able to list events.

* To fix this we need to define appropriate view,edit,admin roles that
  will be aggregated into the Kubeflow view,edit,admin roles so that
  when a user is granted access to a namespace they will get the appropriate
  binding

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/816)

### DIFF
--- a/jupyter/jupyter-web-app/base/cluster-role.yaml
+++ b/jupyter/jupyter-web-app/base/cluster-role.yaml
@@ -52,3 +52,64 @@ rules:
   - get
   - list
   - watch
+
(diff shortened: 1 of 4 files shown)


jlewi:
This time the pytorch_job test failed.

PR #821, opened 2020-01-31
Cleanup notebook UI kubeflow user roles (related to kubeflow/kubeflow#4699)

jlewi:
* There were some labels that shouldn't have added; they were left over
  from bad copy and paste.

* Get rid of events permission; this doesn't actually fix kubeflow/kubeflow#4699
  events are cluster scoped resources. So creating a rolebinding has no effect.

* We primarily use Kubeflow roles as rolebindings not clusterrolebindings
  so including cluster scoped resources doesn't make much sense.

* Related to kubeflow/kubeflow#4699

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/821)

### DIFF
--- a/jupyter/jupyter-web-app/base/cluster-role.yaml
+++ b/jupyter/jupyter-web-app/base/cluster-role.yaml
@@ -61,10 +61,6 @@ metadata:
   name: kubeflow-notebook-ui-admin
   labels:
     rbac.authorization.kubeflow.org/aggregate-to-kubeflow-admin: "true"
-aggregationRule:
-  clusterRoleSelectors:
-  - matchLabels:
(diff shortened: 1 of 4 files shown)


jlewi:
The error is applying the seldon manifest.

```
2020-02-04T18:50:17.729517291Z  main            util.py                     72 INFO     for: "/tmp/kout664878569": unable to find api field in struct Webhook for the json field "objectSelector"]
```

I think this was fixed in #833 so I'm going to rebase to see if that fixes the problem.

[kubeflow-manifests-presubmit-kfctl-go-iap-821-2be231a-9792-5925-3747689709.log.txt](https://github.com/kubeflow/manifests/files/4155877/kubeflow-manifests-presubmit-kfctl-go-iap-821-2be231a-9792-5925-3747689709.log.txt)

PR #831, opened 2020-02-02
Regenerate tests

jbrette:
- Exclude kfdef
- Account for directory name with dot
- Rerun generation

Resolves #

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/831)

### DIFF
--- a/hack/gen-test-target.sh
+++ b/hack/gen-test-target.sh
@@ -12,7 +12,7 @@ kebab-case-2-PascalCase() {
   local a=$1 b='' array
   IFS='-' read -r -a array <<< "$a"
   for element in "${array[@]}"; do
-    part="${element}"
+    part="${element//./_}"
     part="$(tr '[:lower:]' '[:upper:]' <<< ${part:0:1})${part:1}"
     b+=$part
   done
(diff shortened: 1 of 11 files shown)


hack/gen-test-target.sh
```diff
@@ -12,7 +12,7 @@ kebab-case-2-PascalCase() {
   local a=$1 b='' array
   IFS='-' read -r -a array <<< "$a"
   for element in "${array[@]}"; do
-    part="${element}"
+    part="${element//./_}"
```
jlewi: What is this change for?

jbrette: Simply deals with directory names which are like these : components/v1.0.0/....
The generated go code should not contains `.` in the function names. The code change replaces it with _

jlewi:
/test all

Looks like the 2nd apply test failed; I think @gabrielwen is working on fixing that; you might want to talk to him if its not a flake.

/lgtm
/approve

jlewi:
See 
https://github.com/kubeflow/manifests/pull/821#issuecomment-582101210

Regarding the test failure.

PR #833, opened 2020-02-02
Remove 1.15 selectors for Seldon

### LINKED ISSUE #832: Seldon Manifests _only_ support kubernetes 1.15
The last Seldon operator [manifests](https://github.com/kubeflow/manifests/blob/c4f51b1cc0b4854b6f6180d5a3bb021dd65691a8/seldon/seldon-core-operator/base/resources.yaml#L3522) only support Kubernetes 1.15.

API reference [here](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-objectselector)

The error you'd see trying to apply these on older kubernetes versions is:

```
unable to find api field in struct Webhook for the json field "objectSelector", 
```

cliveseldon:
Related to #832

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/833)


WillBeebe:
These changes LGTM but can we add a comment or console log to where they're generated in the Makefile?

https://github.com/kubeflow/manifests/blob/master/seldon/Makefile#L5-L8

seldon/seldon-core-operator/base/resources.yaml
```diff
@@ -3515,14 +3515,6 @@ webhooks:
       path: /mutate-machinelearning-seldon-io-v1-seldondeployment
   failurePolicy: Fail
   name: mseldondeployment.kb.io
-  namespaceSelector:
```
jlewi: Should we keep the namespaceSelectors? I think those are supported in K8s 1.14.
Also should we change the namespace selector to only match the profile created namespaces.

See: kubeflow/kfserving#568 I think we want to scope it to namespaces where models should be deployed.

See: https://github.com/kubeflow/kubeflow/blob/c80316d3a646808da4e59ead71ddd1e95d1fb796/components/profile-controller/controllers/profile_controller.go#L65

We automatically add the label "serving.kubeflow.org/inferenceservice": "enabled" to profile created namespaces.
So I think it make sense to scope seldon webhooks using that label as well. This ensures problems with the seldon webhooks don't affect other namespaces .

seldon/seldon-core-operator/base/resources.yaml
```diff
@@ -3666,10 +3656,8 @@ webhooks:
     matchExpressions:
     - key: seldon.io/controller-id
       operator: DoesNotExist
-  objectSelector:
-    matchExpressions:
-    - key: seldon.io/controller-id
-      operator: DoesNotExist
+    - key: serving.kubeflow.org/inferenceservice
+      operator: Exists
```
jlewi: Could we we match specific labels? like "enabled" to match KFServing
https://github.com/kubeflow/kfserving/pull/571/files#diff-2351c76b931b04dcaa68b4d83b8cb372R149

If we can match multiple labels as opposed to just one lets do "enabled" and "true"

jlewi:
/lgtm
/approve

We will need to cherry-pick this into the 1.0 branch after it is merged.

Issue #834, opened 2020-02-03
Make sure all the application versions are set correctly


jlewi:
It looks like KFServing Application versions are currently tracking KF versions
https://github.com/kubeflow/manifests/blob/d1b3765d09c4175a8e8189ca840c3700a2b5efe5/kfserving/kfserving-install/overlays/application/application.yaml#L13

https://github.com/kubeflow/manifests/blob/6dfd70cad6e9c68496ed175b9f774f12a79da7c0/kfserving/kfserving-install/overlays/application/application.yaml#L13

That is probably because there were erroneously updated via some script
https://github.com/kubeflow/manifests/pull/784/files

richardsliu:
/cc @kkasravi 

The `app.kubernetes.io/version` label is currently being used to identify the kubeflow version associated with each application. As part of the upgrade process we look for expired resources (e.g. labelled with "v0.7.0") and delete them.

If we want to make this version represent the application-specific version (as opposed to the overall Kubeflow version), we need to rethink how upgrade is going to work.

jlewi:
@richardsliu  Is it going to be confusing to users if the application version doesn't refer to the version of the application itself rather than Kubeflow?

Here's a oneliner to show all application versions

```
kubectl -n kubeflow get applications -o=custom-columns=Name:'.metadata.name','Verions:.metadata.labels.app\.kubernetes\.io\/version' --sort-by='.metadata.name' 
Name                             Verions
api-service                      0.1.31
argo                             v2.3.0
bootstrap                        v1.0.0
centraldashboard                 v1.0.0
cloud-endpoints                  v1.0.0
gpu-driver                       v1.0.0
jupyter-web-app                  v1.0.0
katib-controller                 v1.0.0
katib-crds                       v1.0.0
kfserving                        v1.0.0
kfserving-crds                   v1.0.0
kubeflow                         v1.0.0
metadata                         v1.0.0
minio                            0.1.31
mysql                            0.1.31
notebook-controller              v1.0.0
persistent-agent                 0.1.31
pipeline-visualization-service   0.1.31
pipelines-runner                 0.1.31
pipelines-ui                     0.1.31
pipelines-viewer                 0.1.31
profiles                         v1.0.0
pytorch-job-crds                 v1.0.0
pytorch-operator                 v1.0.0
scheduledworkflow                0.1.31
seldon-core-operator             v1.0.1
spark-operator                   v1.0.0
tf-job-crds                      v1.0.0
[...]

PR #849, opened 2020-02-04
Set KFServing version to 0.2.2 in application resources; kubeflow/manifests#834

jlewi:
* The version of KFServing in 1.0 is 0.2.2; set the application resources
  accordingly

* Uses the script kubeflow/testing#596

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/849)

### DIFF
--- a/kfserving/kfserving-crds/overlays/application/application.yaml
+++ b/kfserving/kfserving-crds/overlays/application/application.yaml
@@ -3,29 +3,29 @@ kind: Application
 metadata:
   name: kfserving-crds
 spec:
-  selector:
-    matchLabels:
-      app.kubernetes.io/name: kfserving-crds
-      app.kubernetes.io/instance: kfserving-crds-v1.0.0
-      app.kubernetes.io/managed-by: kfctl
-      app.kubernetes.io/component: kfserving-crds
-      app.kubernetes.io/part-of: kubeflow
-      app.kubernetes.io/version: v1.0.0
+  addOwnerRef: true
   componentKinds:
   - group: core
     kind: ConfigMap
   - group: apps
     kind: Deployment
   descriptor:
-    type: kfserving-crds
-    version: v1beta1
-    description: ""
-    maintainers: []
-    owners: []
+    description: ''
     keywords:
-     - kfserving-crds
-     - kubeflow
+    - kfserving-crds
+    - kubeflow
     links:
     - description: About
-      url: ""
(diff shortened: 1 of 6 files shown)


kfserving/kfserving-crds/overlays/application/application.yaml
```diff
@@ -3,29 +3,29 @@ kind: Application
...
+    type: kfserving-crds
+    version: v1beta1
+  selector:
+    matchLabels:
+      app.kubernetes.io/component: kfserving-crds
+      app.kubernetes.io/instance: kfserving-crds-0.2.2
+      app.kubernetes.io/managed-by: kfctl
+      app.kubernetes.io/name: kfserving-crds
+      app.kubernetes.io/part-of: kubeflow
+      app.kubernetes.io/version: 0.2.2
```
animeshsingh: believe this is the version for application, i.e. kfserving

kfserving/kfserving-crds/overlays/application/kustomization.yaml
```diff
@@ -1,13 +1,13 @@
...
-  app.kubernetes.io/name: kfserving-crds
-  app.kubernetes.io/instance: kfserving-crds-v1.0.0
-  app.kubernetes.io/managed-by: kfctl
   app.kubernetes.io/component: kfserving-crds
+  app.kubernetes.io/instance: kfserving-crds-0.2.2
+  app.kubernetes.io/managed-by: kfctl
+  app.kubernetes.io/name: kfserving-crds
   app.kubernetes.io/part-of: kubeflow
-  app.kubernetes.io/version: v1.0.0
+  app.kubernetes.io/version: 0.2.2
```
animeshsingh: same here?

kfserving/kfserving-crds/overlays/application/application.yaml
jlewi: That is correct. The application in this case is KFServing so the version should be KFServing version not Kubeflow version.

Ideally "Kubeflow" is its own application which should be the parent of many applications.
There might be some ongoing work to clean up /fix labels to make that work.

kfserving/kfserving-crds/overlays/application/kustomization.yaml
jlewi: Yup. This should be the KFServing version.

PR #853, opened 2020-02-04
Update Katib manifests with the latest changes

andreyvelich:
This PR updates Katib manifests with the latest changes and also changes some parameters according to Katib repository to be consistent.

@richardsliu Do we need to update port to 8080 [here](https://github.com/kubeflow/manifests/blob/master/katib/katib-controller/overlays/istio/katib-ui-virtual-service.yaml#L20), since we are using `targetPort: 8080` for Katib UI or we don't need to make this change?

Please, take a look.
/assign @richardsliu @johnugeorge 
/cc @hougangliu @gaocegege 

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/853)


katib/katib-controller/base/katib-controller-rbac.yaml
```diff
@@ -54,13 +54,10 @@ rules:
   resources:
   - experiments
   - experiments/status
-  - experiments/finalizers
```
richardsliu: Why are these removed?

katib/katib-controller/base/katib-mysql-pv.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
```
richardsliu: Is this needed?

johnugeorge: PV and pvc changes are not required

andreyvelich: I think it is useful when you are not using GCP and you don't have StorageClass on your cluster. We have this PV in Katib manifest.
What do you think @richardsliu @johnugeorge ?

katib/katib-controller/base/katib-controller-rbac.yaml
andreyvelich: We don't have these roles in controller rbac: https://github.com/kubeflow/katib/blob/master/manifests/v1alpha3/katib-controller/rbac.yaml#L54. Do we need it?

katib/katib-controller/base/katib-mysql-pv.yaml
johnugeorge: But then, need not be in base in that case.

andreyvelich: @johnugeorge @richardsliu Ok, I removed it from base.

PR #860, opened 2020-02-05
Set the correct version for Katib,Seldon,metadata kubeflow/manifests#834

jlewi:
* The version in the application CR should be the version of the application
  not Kubeflow

* Since Katib and metadata aren't going 1.0 we should bump the version to
  "0.8.0"

* Seldon's version is 1.15

* Note: I'm not including the "v" prefix in the version number. I'm not sure
  where that started or why we did it but it looks like a bug. We are
  following semantic versioning. So the version is X.Y.Z.
  Most likely the v was included as some sort of prefix as part of some larger
  tag e.g. to ensure it was interpreted as a string and not an integer.

* Related to kubeflow/manifests#834

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/860)

### DIFF
--- a/metadata/overlays/application/application.yaml
+++ b/metadata/overlays/application/application.yaml
@@ -3,14 +3,7 @@ kind: Application
 metadata:
   name: metadata
 spec:
-  selector:
-    matchLabels:
-      app.kubernetes.io/name: metadata
-      app.kubernetes.io/instance: metadata-v0.7.0
-      app.kubernetes.io/managed-by: kfctl
-      app.kubernetes.io/component: metadata
-      app.kubernetes.io/part-of: kubeflow
-      app.kubernetes.io/version: v0.7.0
+  addOwnerRef: true
   componentKinds:
   - group: core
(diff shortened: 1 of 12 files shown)


metadata/overlays/application/application.yaml
```diff
@@ -21,20 +14,27 @@ spec:
...
+    type: metadata
+    version: alpha
+  selector:
+    matchLabels:
+      app.kubernetes.io/component: metadata
+      app.kubernetes.io/instance: metadata-0.8.0
+      app.kubernetes.io/managed-by: kfctl
+      app.kubernetes.io/name: metadata
+      app.kubernetes.io/part-of: kubeflow
+      app.kubernetes.io/version: 0.8.0
```
richardsliu: @zhenghuiwang Is this the correct version for metadata?

zhenghuiwang: The correct version is 0.2.1.

Issue #774, opened 2020-01-28
Get rid of KFDef kustomize magic?


jlewi:
Right now KFDef provides the following functionality

1. The ability to list a set of kustomize packages to be deployed
1. The ability to set parameters for each of those applications
1. The ability to combine multiple overlays like mixins (see [comment](https://github.com/kubeflow/kubeflow/issues/3490#issuecomment-579395304))

Could we get rid of this magic and rely more on straight kustomize.

* Per kubernetes-sigs/kustomize#171 I believe kustomize supports composition when each base package describes a distinct set of resources. So instead of listing kustomize packages in KFDef I think we could just define a new kustomize package with a kustomization.yaml that looked like

```...
bases:
- jupyter/notebook-controller/overlays/application
- jupyter/jupyter-web-app/overlays/application
```

A user could then just do

```
kustomize build to see all resources
```

Right now our overlays are structured such that we have multiple overlays (e.g. istio and application) which depend on base and then kfctl combines them. 

But we don't really need that magic. We could restructure our overlays so that they all use
what kustomize supports.

Along these lines we might want to rethink how we do parameter substitution to better leverage
https://github.com/kubernetes-sigs/kustomize/blob/master/examples/configGeneration.md

Can we structure it so that overriding parameters) would just be generating config map patches overlays in this package?

Related issues and threads:
kubeflow/kfctl#156
kubeflow/kubeflow#3490

jlewi:
From a release stand point I think it could be very useful if we had overlays for each version e.g.
if we structured our kustomize apps 

```
bases
  ...
overlays
   master
     ...
    v1.0.0
    v1.1.0
```

v1.1.0 could  potentially use v1.0.0 as its base.

One of the motivations here is to reduce the cost of maintaining release branches.

Right now there is a bunch of work that needs to be done to update the release branch for each application e.g.
  1. Bump the version in the Application CR
  1. Update the docker image

Right now we make these changes in the base package.

That means we can't make these changes until we cut a release branch. This means its hard to simultaneously 

* Start preparing the application for the release
* Continue to fix bugs

If we use overlays for the version then we can start checking that into master and easily keep it in sync as fixes to the underlying base deployment go in.

SebBlin:
The magic feature of having multiple overlays and combining them is a **"really great feature"** and should be kept.
I use it to build my own repo that contain all my personnal customization to apply during the deployment process.

Here is an exemple below for Dex : 
[...]

PR #873, opened 2020-02-06
WIP get rid of merging overlays

kunmingg:
Related https://github.com/kubeflow/manifests/issues/774

* flatten overlays; 
* create kustomization file as signal entry point for `apply`.
* make kustomize package deployable on its own (through `kustomize` CLI)

For example for `jupyter-web-app`, before the change we need to merge overlays `application` and `istio` with `kfctl` logic since user need both of them:
```
bases:
...
overlays:
- application
- istio
```
However merge overlays is not supported by `kustomize`, maintain this feature in `kfctl` is expensive and will block user deploying our pkg with `kustomize` CLI directly.

After this PR merge, user can just `kustomize build base` directly:
```
bases:
- application
- base
- istio
- kustomization.yaml
```

`overlays` are still supported but only one of them will be used during deployment: `kustomize build overlays/name`

update test generation to allow nested `kustomization.yaml` under `base`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/873)


kunmingg:
@jlewi 
Can you review the changes on `jupyter-web-app`? 
If looks good will apply same change to other components.

jlewi:
@kunmingg This is really elegant. So If I understand correctly if someone just wanted
to install the jupyter web app without the application CR or ISTIO they could do

```
kustomize build jupyter-web-app/bases 
```

If they want to install the jupyter web-app with the application resource they would do

```
kustomize build jupyter-web-app
```

If we wanted to define some overlays for GCP; e.g. if we wanted to override the jupyter-web-app configmap we would define an overlay whose base was jupyter-web-app?

How about including a markdown doc in this PR that describes the layout and changes? e.g. "manifests_v2.md"? We could mark it as a proposal and then discuss and circulate on the mailing list.

Related to that it might be useful to demonstrate the use of composition of applications. So define a Kubeflow manifest that combines the TFJob and JupyterWebApp manifests? This would essentially eliminate the use of KFDef to list all the KF applications. 

Would it be useful to put these new applications in a different subdirectory? e.g.
experimental/jupyter
experimental/tfjob

This would potentially allow us to check in the code while we experiment and socialize the idea?

jlewi:
@kunmingg WDYT the next steps are here?

I think I've changed course a bit from doing this in experimental. I think I'm leaning towards doing this in a more incremental way. In particular, I think we should consider moving forward refactoring each application manifest and updating the KFDefs to remove the dependency on applying multiple overlays. 

/cc @kimwnasptd

kunmingg:
@jlewi Sounds good, let me update this PR with docs explaining new behaviors.

jlewi:
Actually if we look it at what kfctl does it just ends up generating the following kustomization file

```
bases:
- base
- overlays/istio
commonLabels:
  app.kubernetes.io/component: jupyter-web-app
  app.kubernetes.io/instance: jupyter-web-app-v1.0.0
  app.kubernetes.io/managed-by: kfctl
  app.kubernetes.io/name: jupyter-web-app
  app.kubernetes.io/version: v1.0.0
kind: Kustomization
namespace: kubeflow
resources:
- overlays/application/application.yaml
```

I can't think of any reason to autogenerate that file as opposed to just checking it in. All though bases is deprecated so we could just change it to

```
commonLabels:
  app.kubernetes.io/component: jupyter-web-app
  app.kubernetes.io/instance: jupyter-web-app-v1.0.0
  app.kubernetes.io/managed-by: kfctl
  app.kubernetes.io/name: jupyter-web-app
  app.kubernetes.io/version: v1.0.0
kind: Kustomization
namespace: kubeflow
resources:
- base
- overlays/istio
- overlays/application/application.yaml
```

I think we might need to add the configurations to this in order to handle substituting namespace and cluster prefix into the virtual service.

I don't think we really want the files
overlays/istio/kustomization.yaml
overlays/application/kustomization.yaml

That implies you can run kustomize build in that directory but that doesn't really make sense. The resources defined there only make sense as additions to the base package.

That said it might be worth keeping them for some sort of backwards compatibility.

jlewi:
@kunmingg I was doing a bunch of experiments; see https://github.com/kubeflow/manifests/pull/962

TL;DR lets chat before you invest more time in this PR.

PR #885, opened 2020-02-09
EnvoyFilter to add anonymous@kubeflow.org user to all gateway requests

krishnadurai:
Related to kubeflow/kubeflow#4731

Anonymous user anonymous@kubeflow.org added through this manifest to Istio gateway based requests.

Note that this implementation spoofs a user to enable non-auth setups with an anonymous user.
**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

/assign @jlewi @swiftdiaries

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/885)

### DIFF
--- a/istio/add-anonymous-user-filter/base/envoy-filter.yaml
+++ b/istio/add-anonymous-user-filter/base/envoy-filter.yaml
@@ -1,0 +1,19 @@
+apiVersion: networking.istio.io/v1alpha3
+kind: EnvoyFilter
+metadata:
+  name: add-user-filter
+spec:
+  workloadLabels:
+    app: istio-ingressgateway
+  filters:
+  - listenerMatch:
+      listenerType: GATEWAY
+    filterName: envoy.lua
+    filterType: HTTP
+    insertPosition:
+      index: FIRST
+    filterConfig:
+      inlineCode: |
+        function envoy_on_request(request_handle)
+            request_handle:headers():add("kubeflow-userid","anonymous@kubeflow.org")
+        end
--- a/istio/add-anonymous-user-filter/base/kustomization.yaml
+++ b/istio/add-anonymous-user-filter/base/kustomization.yaml
@@ -1,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: istio-system
+resources:
+- envoy-filter.yaml
--- a/kfdef/kfctl_k8s_istio.yaml
+++ b/kfdef/kfctl_k8s_istio.yaml
@@ -29,6 +29,14 @@ spec:
         path: istio/istio
     name: istio
   - kustomizeConfig:
+      parameters:
+      - name: namespace
+        value: istio-system
+      repoRef:
+        name: manifests
+        path: istio/add-anonymous-user-filter
+    name: add-anonymous-user-filter
+  - kustomizeConfig:
       repoRef:
         name: manifests
         path: application/application-crds
--- a/kfdef/kfctl_k8s_istio.yaml
+++ b/kfdef/kfctl_k8s_istio.yaml
@@ -119,6 +119,9 @@ spec:
       overlays:
       - istio
       - application
+      parameters:
+      - name: userid-header
+        value: kubeflow-userid
       repoRef:
         name: manifests
         path: jupyter/jupyter-web-app
--- a/tests/istio-add-anonymous-user-filter-base_test.go
+++ b/tests/istio-add-anonymous-user-filter-base_test.go
@@ -1,0 +1,76 @@
+package tests_test
+
+import (
+	"sigs.k8s.io/kustomize/v3/k8sdeps/kunstruct"
+	"sigs.k8s.io/kustomize/v3/k8sdeps/transformer"
+	"sigs.k8s.io/kustomize/v3/pkg/fs"
+	"sigs.k8s.io/kustomize/v3/pkg/loader"
+	"sigs.k8s.io/kustomize/v3/pkg/plugins"
+	"sigs.k8s.io/kustomize/v3/pkg/resmap"
+	"sigs.k8s.io/kustomize/v3/pkg/resource"
+	"sigs.k8s.io/kustomize/v3/pkg/target"
+	"sigs.k8s.io/kustomize/v3/pkg/validators"
+	"testing"
+)
+
+func writeAddAnonymousUserFilterBase(th *KustTestHarness) {
+	th.writeF("/manifests/istio/add-anonymous-user-filter/base/envoy-filter.yaml", `
+apiVersion: networking.istio.io/v1alpha3
+kind: EnvoyFilter
+metadata:
+  name: add-user-filter
+spec:
+  workloadLabels:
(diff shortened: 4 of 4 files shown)


krishnadurai:
/cc @Tomcli @animeshsingh @Jeffwan 

Does kubeflow/kubeflow#4731 affect your non-auth based setups as well? I will add this filter to your configs.

/cc @yanniszark @johnugeorge PTAL

jlewi:
Thanks @krishnadurai !

nit: Lets not close kubeflow/kubeflow#4731 until we've cherry-picked this onto the 1.0 branch and verified it works.

I did a quick sanity check on a v1 cluster and it looks like this will work.

```
kubectl auth can-i list notebooks --namespace kubeflow-kf-ci-v1-user --as anonymous@kubeflow.org
no

kubectl create rolebinding --user=anonymous@kubeflow.org --clusterrole=kubeflow-edit anonymous --namespace=kubeflow-kf-ci-v1-user
rolebinding.rbac.authorization.k8s.io/anonymous created

kubectl auth can-i list notebooks --namespace kubeflow-kf-ci-v1-user --as anonymous@kubeflow.org
yes
```

In particular it doesn't look like K8s does any validation to ensure the email is a valid id.

It looks like we would still need to create appropriate rolebindings for the target namespaces. I assume

1. Users could create this manually using kubectl
1. If they specify the appropriate role in the profile it will be created?

So unless I'm missing something; LGTM.

krishnadurai:
@jlewi I've verifies that Istio and K8s rolebindings have been created for anonymous@kubeflow.org in anonymous namespace by the profile controller.
Am I missing checking for any other role binding?

krishnadurai:
@jlewi thanks for verifying, I'll check this behavior once and see if there's an appropriate fix.

krishnadurai:
@jlewi Here's my verification. 

I've created a profile called 'anonymous' from the landing page (profile creation page) after the cluster setup is complete and Kubeflow central dashboard is available.

```
kubectl get rolebindings -n anonymous
NAME             AGE
default-editor   21h
default-viewer   21h
namespaceAdmin   21h
```

```
kubectl get rolebindings -n anonymous -o yaml namespaceAdmin
kind: RoleBinding
metadata:
  annotations:
    role: admin
    user: anonymous@kubeflow.org
  creationTimestamp: "2020-02-09T11:38:29Z"
  name: namespaceAdmin
  namespace: anonymous
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: kubeflow-admin
subjects:
- apiGroup: rbac.authorization.k8s.io
  kind: User
  name: anonymous@kubeflow.org
```

```
kubectl auth can-i list notebooks --namespace anonymous --as anonymous@kubeflow.org
yes
```

> In particular it doesn't look like K8s does any validation to ensure the email is a valid id.

As we are looking at the Kubernetes Auth system here, the user in question is the Kubernetes user. One alternative is that we borrow the current Kubernetes user's email and username while setting up the no-auth user's profile, instead of using 'anonymous' and 'anonymous@kubeflow.org' as user-id and email respectively. This setup wouldn't work if multiple users would use this no-auth setup.

The best way to go about this would be to recommend cluster admins to manually associate users with Kubeflow cluster-roles in the no-auth setup to operate with Kubernetes resources.

jlewi:
> As we are looking at the Kubernetes Auth system here, the user in question is the Kubernetes user, which will be different from 'anonymous@kubeflow.org'. 

I think you might be pointing out the following. When the user authenticates to the Kubernetes API (e.g. using kubectl) they won't be authenticating as anonymous@kubeflow.org; they would be authenticating as  user@acme.com.

So if we use anonymous@kubeflow.org we would need two potential RBAC bindings

1. One rule for anonymous@kubeflow.org
1. One rule for user@acme.com

The profile resource can only create one of these rules depending on which user is entered into the profile. 

One possible solution would be using user@acme.com in the envoy filter but that would only work for one user.

What if we did the following

1. Create a ClusterRoleBinding for anonymous@kubeflow.org to kubeflow-edit
    * This would grant the anonymous@kubeflow.org permission in all namespaces
1. Tell users to put their email (e.g. anonymous@kubeflow.org) in the profile

This would grant anonymous@kubeflow.org access to all namespaces. I think that's OK though. If someone is really concerned about security or a production cluster; my expectation would be that they setup a proper solution for AuthN.

krishnadurai:
I agree with your thoughts on defaulting to 'anonymous@kubeflow.org' to profiles being created. CentralDashboard already creates the default profile with 'anonymous@kubeflow.org'.

https://github.com/kubeflow/kubeflow/blob/06b4cb4ba8719893de3ee5cac92ff9de3f2be75f/components/centraldashboard/app/attach_user_middleware.ts#L10-L15

There are two ways to access resources created by Kubeflow:
1. Through the notebook, which has mounted service accounts with appropriate privileges.
1. Through kubectl

While accessing resources through the notebook:

> What if we did the following
> 
>     1. Create a ClusterRoleBinding for [anonymous@kubeflow.org](mailto:anonymous@kubeflow.org) to kubeflow-edit
>
>        * This would grant the [anonymous@kubeflow.org](mailto:anonymous@kubeflow.org) permission in all namespaces      
>     2. Tell users to put their email (e.g. [anonymous@kubeflow.org](mailto:anonymous@kubeflow.org)) in the profile

We would need the users to enter the email 'anonymous@kubeflow.org' while creating a profile through its spec.
We wouldn't require the cluster role binding as the notebook would mount service accounts which would take care of the access - with namespaceAdmin binding.

```
kubectl get rolebindings -n anonymous -o yaml namespaceAdmin
kind: RoleBinding
metadata:
  annotations:
    role: admin
    user: anonymous@kubeflow.org
  creationTimestamp: "2020-02-09T11:38:29Z"
  name: namespaceAdmin
  namespace: anonymous
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: kubeflow-admin
subjects:
- apiGroup: rbac.authorization.k8s.io
  kind: User
  name: anonymous@kubeflow.org
```

Though this would not work for access through `kubectl` as Kubernetes Auth is going to receive the original email of the user: user@acme.com.

krishnadurai:
@jlewi I'm clarifying.
Are you suggesting that users add their email as 'anonymous@kubeflow.org' in their kubeconfig as well?

jlewi:
@krishnadurai Thanks I think what you are saying is that everything should just work.

IIUC the CUJ would be

1. User would access central dashboard
1. This would create a profile with the email `anonymous@kubeflow.org`
     * This would create a role binding for `anonymous@kubeflow.org`
     * So notebooks and other services which use Kubernetes service accounts would work
1. Using the [profiles UI](https://master.kubeflow.org/docs/other-guides/multi-user-overview/#managing-contributors-through-ui) they could add their K8s identity (e.g. jlewi@google.com) as a contributor to the namespace
    * This would grant them kubectl access to the newly created namespace
    * They wouldn't need this if they were already a cluster admin

Is that right

/lgtm
/approve
/hold

krishnadurai:
Yes, you are right.

/hold cancel

PR #909, opened 2020-02-13
Add EnvoyFilter to add anonymous to all gateway requests in aws no-auth manifest

Jeffwan:
Resolves https://github.com/kubeflow/kubeflow/issues/4731

Apply the fix https://github.com/kubeflow/manifests/pull/885 on aws manifest without auth. 

This change needs to be cherry-pick back to v1.0-branch

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/909)

### DIFF
--- a/kfdef/kfctl_aws.v1.0.0.yaml
+++ b/kfdef/kfctl_aws.v1.0.0.yaml
@@ -29,6 +29,14 @@ spec:
         path: istio/istio
     name: istio
   - kustomizeConfig:
+      parameters:
+      - name: namespace
+        value: istio-system
+      repoRef:
+        name: manifests
+        path: istio/add-anonymous-user-filter
+    name: add-anonymous-user-filter
+  - kustomizeConfig:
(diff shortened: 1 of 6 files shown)


kfdef/kfctl_aws.v1.0.0.yaml
```diff
@@ -28,6 +28,14 @@ spec:
         name: manifests
         path: istio/istio
     name: istio
+  - kustomizeConfig:
+      parameters:
```
jlewi: I think you want to modify the files in
https://github.com/kubeflow/manifests/tree/master/kfdef/source/master

So that when the script is to regenerate the files you pull in the changes.

Jeffwan: Ah. Forget these manifests are generate ones after the latest PR. I will make the change

PR #929, opened 2020-02-18
Add option to generate kfdef for KubeFlow Operator

Tomcli:
Right now the generated kfdef is not a valid Kubernetes resource since it's missing the metadata.name field. We want to add an option to the build_kfdef_specs.py to generate kfdef with metadata.name that can be used on KubeFlow Operator.

Also fixing a yaml formatting bug when generating kfdef. 

Resolves #

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/929)

### DIFF
--- a/kfdef/source/README.md
+++ b/kfdef/source/README.md
@@ -1,6 +1,8 @@
 This directory contains kustomization packages that are used to generate the YAML specs for the KFDef.
 
 The script `hack/build_kfdef_specs.py` is used to run kustomize and output the YAML files.
+By default *metadata.name* is not generated. However, `OPERATOR_KFDEF` environment variable can be
+set to `true` for building the KubeFlow operator version of kfdefs that contain *metadata.name*.
 
 ## Subdirectories
 
--- a/hack/build_kfdef_specs.py
+++ b/hack/build_kfdef_specs.py
@@ -59,10 +59,11 @@ class KFDefBuilder:
 
           # Remove the name. Kustomize requires a name but we don't want
           # a name so that kfctl will fill it in based on the app directory
-          del spec["metadata"]["name"]
+          if os.environ.get("OPERATOR_KFDEF", "").lower() != "true":
+            del spec["metadata"]["name"]
 
           with open(new_file, "w") as hf:
-            yaml.safe_dump(spec, hf)
+            yaml.safe_dump(spec, hf, default_flow_style = False)
 
 if __name__ == "__main__":
 


kfdef/source/README.md
```diff
@@ -1,6 +1,8 @@
 This directory contains kustomization packages that are used to generate the YAML specs for the KFDef.
 
 The script `hack/build_kfdef_specs.py` is used to run kustomize and output the YAML files.
+By default *metadata.name* is not generated. However, `OPERATOR_KFDEF` environment variable can be
+set to `true` for building the KubeFlow operator version of kfdefs that contain *metadata.name*.
```
animeshsingh: Would it cause any harm if *metadata.name* is generated for every kfdef?

animeshsingh:
Also vis a vis the new instructions - do the kfdef need to be *generated* each time before you can use them? Are there pre-generated ones? Do we need to add more details in operator readme to account for that?

kfdef/source/README.md
Tomcli: https://github.com/kubeflow/manifests/blob/15ac5f96173375b7df04243ab6424c840fc1be2f/hack/build_kfdef_specs.py#L60-L62
Based on the comments it preferred to let kfctl to fill in the name instead of generating it with the script.

Tomcli:
Right now the generated ones don't have the names, so we can't apply it directly to Kubernetes as a custom resource for the KubeFlow operator. I can extend the instructions to show how it can regenerate the kfdef for KubeFlow Operator.

kfdef/source/README.md
animeshsingh: @jlewi @kunmingg 

what's the rationale to omit the *metadata.name* from kfdef? Previously the operator could directly use them, and now we have to tell user to do manipulation in order to be able to use them

jlewi:
The name isn't in the KFDef by design.  The expectation is that 

1. The user will set the name or
1. kfctl will select an appropriate default name
    * kfctl selects the name based on the folder in which it is generated

Various resources are named based on the name so we want it to be unique for deployments. So hardcoding a name like "kubeflow" doesn't make sense.

kfdef/source/README.md
jlewi: Yes; generating a name for every KFDef will be a significant behavior change.

jlewi:
@animeshsingh and @Tomcli is it possible to set the name and other defaults in an admission hook?

I don't think we want to have an environment variable or other mechanism of generating different versions; e.g. with name or without name. My prediction is that will lead to problems because people will run things with the incorrect flag.

Perhaps instead we should let the different config files (e.g. IBM vs. GCP) behave differently. So if you want a name in your configs then change the logic to allow it.

animeshsingh:
> Various resources are named based on the name so we want it to be unique for deployments. So hardcoding a name like "kubeflow" doesn't make sense.

@jlewi Can we have "kubeflow" by default, and users have an option to make it unique?

animeshsingh:
>  is it possible to set the name and other defaults in an admission hook?

@jlewi will make the solution unnecessarily complex, and we are not fan of web-hooks for all the other side effects they bring.  In my opinion, having a default name and giving users an option to override it through kfctl would be best - keeps the default kfdefs there as compliant Kubernetes resources by default

animeshsingh:
@jlewi do we agree with this minimal change - there is a default name in kfdefs and kfctl overrides it with user supplied name?

jlewi:
> @jlewi do we agree with this minimal change - there is a default name in kfdefs and kfctl overrides it with user supplied name?

I don't think I understand why that change is necessary.

If you want to publish KFDef's with the name field set; why not just publish KFDefs with the name field set?

If `build_kfdef_specs.py` doesn't do what you want; then maybe just don't use it?

It sounds like you are trying to create and curate KFDefs intended to be used with the KFDef operator and I suspect OpenHub.

The patterns/assumptions you need are different from other use cases. My suggestion would be to curate your KFDefs separately and build out your toolchain for doing so as needed.

PR #944, opened 2020-02-22
Remove envoyfilter in identity aware manifest and fix bugs

Jeffwan:
Resolves https://github.com/kubeflow/kubeflow/issues/4792

- Remove envoy filter in AWS identity aware manifest
- Add  `name: namespace value: istio-system` to make sure mix adaptor are installed inside istio-system namespace
- optimize parameters

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/944)

### DIFF
--- a/kfdef/source/master/kfctl_aws_cognito.yaml
+++ b/kfdef/source/master/kfctl_aws_cognito.yaml
@@ -37,14 +37,6 @@ spec:
         path: istio/istio
     name: istio
   - kustomizeConfig:
-      parameters:
-      - name: namespace
-        value: istio-system
-      repoRef:
-        name: manifests
-        path: istio/add-anonymous-user-filter
-    name: add-anonymous-user-filter
-  - kustomizeConfig:
       repoRef:
         name: manifests
         path: application/application-crds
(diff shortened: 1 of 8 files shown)


kfdef/source/master/kfctl_aws_cognito.yaml
```diff
@@ -339,12 +331,12 @@ spec:
       overlays:
       - application
       parameters:
+      - name: namespace
```
krishnadurai: Shouldn't this be 'istio-namespace' instead of 'namespace' since the line:
https://github.com/kubeflow/manifests/pull/944/files#diff-e7eae0bee4eed3f89046e8546a08bce1R24

Refers to this as 'istio-namespace'?

Am I missing something?

Jeffwan: `namespace` is used to override customize resource files namespace. I notice kfctl will by default use `kubeflow` even I have fix value `istio-system` in the manifest.https://github.com/kubeflow/manifests/blob/a235b9825390da5c8b7cf8f479c5802509d4e936/aws/aws-istio-authz-adaptor/base/kustomization.yaml#L3

It still has `istio-namespace` https://github.com/kubeflow/manifests/blob/a235b9825390da5c8b7cf8f479c5802509d4e936/aws/aws-istio-authz-adaptor/base/params.env#L3  and I just leave a default value and don't explicitly override it in the manifest.

krishnadurai: Ok thanks.

Jeffwan:
@jlewi There's a typo. could you comment it again?

Issue #924, opened 2020-02-17
Create a separate IngressGateway for KFServing to disable Auth on gateway requests


krishnadurai:
KNative probes currently are blocked because of authentication in Auth based Kubeflow setups as described in the following issues:
https://github.com/knative/serving/issues/6829
https://github.com/kubeflow/kfserving/issues/668

A viable temporary solution is to ensure KFServing has its own Ingress Gateway setup without auth.

/cc @yuzisun

yuzisun:
thanks @krishnadurai !

yuzisun:
FYI this is the guide you can follow to create a separate istio ingress gateway which works with Knative 
https://knative.dev/docs/serving/setting-up-custom-ingress-gateway/

Jeffwan:
@krishnadurai What's the error message? I only see `RBAC: access defnied` in activator pods.

```
{"level":"error","ts":"2020-04-03T23:14:35.266Z","logger":"activator","caller":"net/revision_backends.go:285","msg":"Failed to probe clusterIP 10.100.94.110:80","commit":"6b0e5c6","knative.dev/controller":"activator","knative.dev/pod":"activator-6cf5b994c5-vcpp9","error":"unexpected body: want \"queue\", got \"RBAC: access denied\"","stacktrace":"knative.dev/serving/pkg/activator/net.(*revisionWatcher).checkDests\n\t/home/prow/go/src/knative.dev/serving/pkg/activator/net/revision_backends.go:285\nknative.dev/serving/pkg/activator/net.(*revisionWatcher).run\n\t/home/prow/go/src/knative.dev/serving/pkg/activator/net/revision_backends.go:326"}
```
the cluster IP `10.100.94.110` is the svc object created from inferenceservice.  In my example, `flowers-sample-predictor-default-g9875-private`

Can you help verify it's the same issue?

yuzisun:
@Jeffwan i think this is the issue when istio RBAC is turned on, can you check if that’s the case ?

PR #949, opened 2020-02-25
Adds kfserving ingressgateway for Istio 1.1.6

krishnadurai:
Modifies KFServing and KNative Serving config-maps to use this gateway

Addresses: kubeflow/manifests#924

This implementation introduces an alternative Kubeflow ingress-gateway for KFServing. This route was chosen because the existing Kubeflow gateway is protected by Auth in several configurations. The Knative service readiness prober requests fail when they try to probe a KFServing InferenceService since those prober requests are not backed with Authentication credentials.

This implementation is a temporary solution until KNative introduces a solution for probing services behind an authenticated gateway as mentioned in this issue: knative/serving#6829

Related issues:
kubeflow/kfserving#668

/cc @yuzisun @animeshsingh 

/hold for testing

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/949)

### DIFF
--- a/istio/kfserving-gateway/base/service.yaml
+++ b/istio/kfserving-gateway/base/service.yaml
@@ -1,0 +1,38 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: kfserving-ingressgateway
+  labels:
+    app: kfserving-ingressgateway
+    kfserving: ingressgateway
+spec:
+  type: LoadBalancer
+  selector:
+    app: kfserving-ingressgateway
+    kfserving: ingressgateway
+  ports:
+    - name: http2
+      nodePort: 32380
+      port: 80
+      targetPort: 80
+    - name: https
+      nodePort: 32390
+      port: 443
+    - name: tcp
+      nodePort: 32400
+      port: 31400
+    - name: tcp-pilot-grpc-tls
+      port: 15011
+      targetPort: 15011
+    - name: tcp-citadel-grpc-tls
+      port: 8060
+      targetPort: 8060
+    - name: tcp-dns-tls
+      port: 853
+      targetPort: 853
+    - name: http2-prometheus
+      port: 15030
+      targetPort: 15030
+    - name: http2-grafana
+      port: 15031
+      targetPort: 15031
--- a/istio/kfserving-gateway/base/service.yaml
+++ b/istio/kfserving-gateway/base/service.yaml
@@ -11,6 +11,9 @@ spec:
     app: kfserving-ingressgateway
     kfserving: ingressgateway
   ports:
+    - name: status-port
+      port: 15020
+      targetPort: 15020
     - name: http2
       nodePort: 32380
       port: 80
@@ -30,9 +33,18 @@ spec:
     - name: tcp-dns-tls
       port: 853
       targetPort: 853
+    - name: https-kiali
+      port: 15029
+      targetPort: 15029
     - name: http2-prometheus
       port: 15030
       targetPort: 15030
     - name: http2-grafana
       port: 15031
       targetPort: 15031
+    - name: https-tracing
+      port: 15032
+      targetPort: 15032
+    - name: tls
+      port: 15443
+      targetPort: 15443
--- a/istio/kfserving-gateway/base/deployment.yaml
+++ b/istio/kfserving-gateway/base/deployment.yaml
@@ -1,4 +1,4 @@
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
 kind: Deployment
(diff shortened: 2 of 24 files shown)


krishnadurai:
Note that PR for Istio 1.3.1 will be raised separately. This has been verfied in kfctl_k8s_istio and kfctl_gcp_iap configurations.

istio/kfserving-gateway/base/service.yaml
```diff
@@ -0,0 +1,50 @@
...
+      port: 853
+      targetPort: 853
+    - name: https-kiali
+      port: 15029
+      targetPort: 15029
+    - name: http2-prometheus
+      port: 15030
+      targetPort: 15030
+    - name: http2-grafana
+      port: 15031
```
animeshsingh: are these ports for kiali/grafana/prometheus always constant? do we deploy them with kubeflow?

istio/kfserving-gateway/base/service.yaml
```diff
@@ -0,0 +1,50 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: kfserving-ingressgateway
+  labels:
+    app: kfserving-ingressgateway
+    kfserving: ingressgateway
+spec:
+  type: LoadBalancer
```
yuzisun: Looks the kubeflow istio ingressgateway is using NodePort https://github.com/kubeflow/manifests/blob/master/istio/istio-install/base/istio-noauth.yaml#L14047
Would this work for on prem setup?

krishnadurai: @yuzisun The reason for istio-ingressgateway being NodePort is because of Google's IAP needing it to be NodePort.

> Would this work for on prem setup?

If an on-premise setup has some component to provision a LoadBalancer, then LoadBalancer will get its IP. If not, the NodePorts are still accessible though the IP field for LoadBalancer will be in <pending> state.

```
NAME                   TYPE           CLUSTER-IP      EXTERNAL-IP   PORT(S)                                                                                                                                      AGE
istio-ingressgateway   LoadBalancer   10.103.151.45   <pending>     15020:30047/TCP,80:31380/TCP,443:31390/TCP,31400:31400/TCP,15029:30370/TCP,15030:31766/TCP,15031:32168/TCP,15032:31581/TCP,15443:30268/TCP   2d12h
```

IMO, This should be acceptable. WDYT?

yuzisun: we are using this ingress gateway in GCP IAP setup, so that's ok ?

krishnadurai: In istio-ingressgateway for GCP IAP, the ingress gateway service is exposed to an Identity Aware Proxy (which has auth checking) by the NodePort and the intention is to disallow access through any other mechanism other than this proxy.
Since we don't want to have auth checks, LoadBalancer should be alright.

yuzisun: got it, thanks for the explanation! btw we also need to also add the HPA for this ingress gateway?

krishnadurai: Good suggestion @yuzisun. I'll include the definition for HPA.

krishnadurai: @yuzisun I've included an HPA.

yuzisun:
Thanks @krishnadurai! This is awesome and removes the biggest road block for kubeflow integration!

/lgtm

krishnadurai:
Thanks @yuzisun! You have been super helpful throughout.

@animeshsingh could you please take another look to get this in soon?

istio/kfserving-gateway/base/service.yaml
animeshsingh: Thaks - great to know. What's the total infra cost of these 3 services - believe a container each? Are any customized Kubeflow metrics configured to flow to them

krishnadurai: @animeshsingh These require a container each. Prometheus, Graphana etc do come with the default istio install in Kubeflow.

Kubeflow does report Prometheus metrics through various operators or services like tfjob-operator or profile-controller. Likewise, the gateway container istio-proxy also reports metrics.

These ports in question are ports which the istio-proxy container exposes. The container istio-proxy is a part of the deployment servicing kfserving-gateway. These ports host https end-points for each of these services like Prometheus to poll these (gateway) services for metrics.

krishnadurai:
@animeshsingh can you PTAL at this again?

PR #950, opened 2020-02-25
pipeline/api-service: Set minio secure=false

discordianfish:

Currently the apiserver uses secure=false by default. That means people
using S3 for instance will default to not using TLS which is a security
issue.

Since the bundled minio doesn't support TLS, we should explicitly set
secure=false here and change the apiserver's defaults to secure=true.

**Checklist:**
- [z] Unit tests have been rebuilt: 
    1. `cd manifests/tests`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/950)

### DIFF
--- a/pipeline/api-service/base/config-map.yaml
+++ b/pipeline/api-service/base/config-map.yaml
@@ -13,7 +13,8 @@ data:
       "ObjectStoreConfig":{
         "AccessKey": "minio",
         "SecretAccessKey": "minio123",
-        "BucketName": "mlpipeline"
+        "BucketName": "mlpipeline",
+        "Secure": false
       },
       "InitConnectionTimeout": "6m",
       "DefaultPipelineRunnerServiceAccount": "pipeline-runner",
--- a/tests/pipeline/api-service/base/test_data/expected/~g_v1_configmap_ml-pipeline-config.yaml
+++ b/tests/pipeline/api-service/base/test_data/expected/~g_v1_configmap_ml-pipeline-config.yaml
@@ -10,7 +10,8 @@ data:
       "ObjectStoreConfig":{
         "AccessKey": "minio",
         "SecretAccessKey": "minio123",
-        "BucketName": "mlpipeline"
+        "BucketName": "mlpipeline",
+        "Secure": false
       },
       "InitConnectionTimeout": "6m",
       "DefaultPipelineRunnerServiceAccount": "pipeline-runner",
--- a/tests/pipeline/api-service/overlays/application/test_data/expected/~g_v1_configmap_ml-pipeline-config.yaml
+++ b/tests/pipeline/api-service/overlays/application/test_data/expected/~g_v1_configmap_ml-pipeline-config.yaml
@@ -10,7 +10,8 @@ data:
       "ObjectStoreConfig":{
         "AccessKey": "minio",
         "SecretAccessKey": "minio123",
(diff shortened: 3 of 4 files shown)


Bobgy:
FYI, there are two more manifests you need to update:
* https://github.com/kubeflow/pipelines/blob/master/manifests/gcp_marketplace/chart/kubeflow-pipelines/templates/pipeline.yaml#L643
* https://github.com/kubeflow/pipelines/blob/master/manifests/kustomize/base/pipeline/ml-pipeline-apiserver-deployment.yaml

discordianfish:
@Bobgy Hrm.. for what are these manifests used? kfctl uses the manifest repo. Wouldn't it make sense to remove the manifests here?
Either way, for GCP I'd hope secure=true would work and the other one I don't see where these things are set in general.

Bobgy:
https://www.kubeflow.org/docs/pipelines/installation/overview/
They are for standalone and GCP hosted manifests, there are other users who prefer not to use Kubeflow as a whole.

discordianfish:
@Bobgy Hrm honestly I don't really understand these manifests.

Re/ GCP: Why is it deploying minio even when managed storage is used? Should I set secure=false in both cases?

Re/ standalone: This doesn't seem to specify any config and it's pinned to a rather old apiserver version (aka shouldn't be affected by the change of defaults in apiserver). Should I add a env var to disable it there too?

Jeffwan:
Hi, I have a PR to add more options here to support S3. I will clean it up and send it out. By default,  we don't have to configure this field

Bobgy:
> Re/ GCP: Why is it deploying minio even when managed storage is used? Should I set secure=false in both cases?

It uses minio as a proxy to managed storage. Yes, I think you should set secure=false.

> Re/ standalone: This doesn't seem to specify any config and it's pinned to a rather old apiserver version (aka shouldn't be affected by the change of defaults in apiserver). Should I add a env var to disable it there too?

Yeah, it feels confusing, but version is actually centralized in kustomization.yaml, version here doesn't take effect. Yes, you should add a env var here too.

Bobgy:
@Jeffwan Can you elaborate why we don't have to configure this field?

discordianfish:
@Bobgy Okay, I've added the env vars to the manifests in https://github.com/kubeflow/pipelines/pull/3168

discordianfish:
Tests over in pipelines pass as well, so should be good to merge
/assign @jlewi

jlewi:
/assign @Bobgy 

@Bobgy please lgtm and then ping me to approve it. Separately we should create an OWNERs file in /pipelines

Bobgy:
Asking for ownership in https://github.com/kubeflow/manifests/pull/967

This change is backward compatible, we can merge it first.
/lgtm

@jlewi for approval

Bobgy:
@Jeffwan Can you refer to your new PR? Should we continue this or close?

discordianfish:
I've rebased this, can we make sure this gets merged before it outdates again? @Bobgy @IronPan @rmgogogo

Bobgy:
@discordianfish thanks for the reminder.

I have permission to approve now.
/lgtm
/approve

Bobgy:
Hmmm, looks like I need tests folder ownership too
@jlewi for approval

Bobgy:
OK, I got the permission now.
/approve

PR #962, opened 2020-02-28
Initial "stack" and v3 patterns for GCP

jlewi:
Initial "stack" for deploying Kubeflow on GCP

* As discussed in http://bit.ly/kf_kustomize_v3 we want to use better
  patterns in our kustomize manifests.

* This PR uses kustomize to compose Kubeflow applications into a "stack" of
  the applications to be installed on GCP.

  * Note: This is only an initial set of applications not all applications
    are installed.

* Define a "stack" a stack is an oppionated way of combining applications
  * a stack is oppinionated about which applications to include and how
    to configure them
  * So we define a GCP stack that would contain the apps to be
    included in GCP deployments.

  * Stacks can be used to replace much of the functionality currently
    achieved by the list of applications in KFDef

    * Instead of listing applications in KFDef we could just point to a
      kustomization file containing the list of applications.

* Define an example illustrating how alice would define an overlay
  to add her kustomizations.

  * Show how she can define a patch to modify the additional configuration
    parameters for her deployment.

Start defining a KFDef that uses the new v3 manifests and the stacks

  * We still need to have multiple applications (and not a single kustomize) application because the namespace for some applications are different
    * e.g. ISTIO is installed in istio-system not kubeflow namespace so
      we keep it as a separate kustomize package
  
* Create v3 version of a couple applications

  * meta controller
  * notebook controller
  * certmanager

Fix some bugs in some v3 packages

   * jupyter_web_app - base/deployment.yaml had some changes that needed to be removed because these are now in a patch because we don't want the changes to be included in the v3 version

Fix centraldashboard v3

  * Move the resources back to base and make base_v3 depend on base rather than
  the other way around.

  * It also looks like we ended up duplicating resources between base_v3 and
    base; probably because of bad merges and rebase

  * In rolebindings don't use vars to substitute in the namespace just
    hardcode the values as needed.

* Upgrade kustomize to v3.2.0 otherwise the tests don't pass
  * In particular using v3.1.0 the jupyter-web-app doesn't have the unique
    name of the kubeflow-config configmap (i.e. with the content hash)
    substituted into the environment variable configmap ref names.

Related to: #1063
Related to: #774

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/962)


krishnadurai:
/cc

Following this work to see how it affects test generation. I would like to review it once ready.

jlewi:
Split off #1050 for profiles and application CR.

jlewi:
I think once the following PRs are merged this should be ready to go:
#1051 jupyter web app manifest v3
#1061 v3 manifests for ISTIO and GCP

jlewi:
Woo hoo. Tests passed
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubeflow_manifests/962/kubeflow-manifests-presubmit/1246863400240156672/

I'm going to rebase to squash the commits and then this PR is ready to go.

jlewi:
/assign @kunmingg 
/assign @krishnadurai 
/cc @yanniszark 

I think this PR is ready for review.

kunmingg:
/lgtm

Shall we add test coverage to make sure param names are unique?

jlewi:
> Shall we add test coverage to make sure param names are unique?

@kunmingg I think we already test that. We test that the kubeflow GCP stack can generate valid and expected kustomize output. So if we have multiple variables with the same name in the included packages that will fail.


/approve

Issue #923, opened 2020-02-17
istio 1.3.1 configuration not working properly in GKE node pools with GPUs


rafaelbarreto87:
/kind bug

**What steps did you take and what happened:**

GKE node pools with GPUs has the following taint: `nvidia.com/gpu=present:NoSchedule`. This prevents `istio-nodeagent` from running in the nodes, which in turns break `istio-proxy`.

Steps to reproduce:

1. Deploy Kubeflow to a GKE cluster with GPU node pool.
2. Try to create a notebook instance requesting a GPU.
3. Observe `istio_proxy` hang waiting for `/var/run/sds/uds_path`:

```
$ kubectl -n admin logs notebook-dev-gpu-0 -c istio-proxy -f
2020-02-17T01:41:52.328901Z     info    FLAG: --applicationPorts="[8888]"
2020-02-17T01:41:52.328955Z     info    FLAG: --binaryPath="/usr/local/bin/envoy"
2020-02-17T01:41:52.328965Z     info    FLAG: --concurrency="2"
2020-02-17T01:41:52.328971Z     info    FLAG: --configPath="/etc/istio/proxy"
2020-02-17T01:41:52.328978Z     info    FLAG: --connectTimeout="10s"
2020-02-17T01:41:52.328985Z     info    FLAG: --controlPlaneAuthPolicy="MUTUAL_TLS"
2020-02-17T01:41:52.328992Z     info    FLAG: --controlPlaneBootstrap="true"
2020-02-17T01:41:52.328997Z     info    FLAG: --customConfigFile=""
2020-02-17T01:41:52.329002Z     info    FLAG: --datadogAgentAddress=""
2020-02-17T01:41:52.329008Z     info    FLAG: --disableInternalTelemetry="false"
2020-02-17T01:41:52.329015Z     info    FLAG: --discoveryAddress="istio-pilot.istio-system:15011"
2020-02-17T01:41:52.329020Z     info    FLAG: --dnsRefreshRate="300s"
2020-02-17T01:41:52.329027Z     info    FLAG: --domain="admin.svc.cluster.local"
2020-02-17T01:41:52.329033Z     info    FLAG: --drainDuration="45s"
2020-02-17T01:41:52.329039Z     info    FLAG: --envoyAccessLogService=""
2020-02-17T01:41:52.329044Z     info    FLAG: --envoyMetricsServiceAddress=""
2020-02-17T01:41:52.329050Z     info    FLAG: --help="false"
[...]

PR #964, opened 2020-02-28
Automated cherry pick of #941: Configure tolerations for istio-nodeagent so that it gets Cherry pick of #941 on v1.0-branch. #941: Configure tolerations for istio-nodeagent so that it gets

krishnadurai:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/964)

### DIFF
--- a/istio-1-3-1/istio-install-1-3-1/base/daemon-set.yaml
+++ b/istio-1-3-1/istio-install-1-3-1/base/daemon-set.yaml
@@ -17,6 +17,11 @@ spec:
         app: nodeagent
         istio: nodeagent
     spec:
+      tolerations:
+      - effect: NoExecute
+        operator: Exists
+      - effect: NoSchedule
+        operator: Exists
       affinity:
         nodeAffinity:
           preferredDuringSchedulingIgnoredDuringExecution:
--- a/tests/istio-1-3-1-istio-install-1-3-1-base_test.go
+++ b/tests/istio-1-3-1-istio-install-1-3-1-base_test.go
@@ -1797,6 +1797,11 @@ spec:
         app: nodeagent
         istio: nodeagent
     spec:
+      tolerations:
+      - effect: NoExecute
+        operator: Exists
+      - effect: NoSchedule
+        operator: Exists
       affinity:
         nodeAffinity:
           preferredDuringSchedulingIgnoredDuringExecution:


krishnadurai:
@rafaelbarreto87 would you please consent to this PR as I'm looking to cherry-pick your changes to 1.0 release branch.

You just need to comment:
@googlebot I consent.

jlewi:
@krishnadurai looks like tests are failing.

jlewi:
Original issue: kubeflow/manifests#923

Do we need to cherry-pick this onto the 1.0 branch? It looks like the issue was about GPUs and ISTIO 1.3. But I think on GCP we are still using the older version of ISTIO?

rafaelbarreto87:
@jlewi We're currently using the Dex deployment in a GKE cluster because it works better for our use case (it's easier to setup in a private cluster). Also, the fix, while works for GCP, is not GCP specific. The core problem is that istio-nodeagent does not get scheduled in all cluster nodes due to the lack of tolerations. This happens to apply to GPU node pools, but is generally applicable.

jlewi:
@rafaelbarreto87 This PR is changing the 1.3 manifests of ISTIO on the 1.0 branch. It only makes sense to cherry-pick this onto the 1.0 branch if some of the 1.0 configs are using ISTIO 1.3. 

The GCP configs are still using ISTIO 1.2 so cherry-picking this onto the 1.0 branch is kind of a null op as far as GCP is concerned.

How would 1.0 configs pick up this change? Do we have KFDef configs on 1.0 that are using the 1.3 manifests?

I'd like to make sure that anything on the 1.0 branch that might be impacted by this change is reviewed by the appropriate reviewers.

It looks like the kfctl_istio_dex.yaml files are using the 1.3 install.
https://github.com/kubeflow/manifests/blob/a0f129168c2b58a6ee6700b6914ef8448769a963/kfdef/kfctl_istio_dex.v1.0.0.yaml#L25

So someone who owns that config should approve this CP.

/assign @yanniszark 

@krishnadurai  I think you can fix the CLA by rebasing the change and squashing commits. That will change the author of the commit to be you and the CLA check will pass.

rafaelbarreto87:
@jlewi I read what you wrote too quickly and I completely understood it. Thought you're saying this should only be applied to the GCP manifests. Never mind that.

@googlebot I consent

rafaelbarreto87:
@googlebot I consent

krishnadurai:
@richardsliu @jlewi can this be included?

krishnadurai:
@yanniszark could you PTAL?

yanniszark:
If I understand correctly, the issue here is that we want the istio-nodeagent daemonset to run on all Nodes.
Some users have Nodes with special hardware (GPUs), on which the DaemonSet is not scheduled.
The current PR tolerates NoSchedule and NoExecute for all keys.

IMO, that's a bit too much.
If you are adding tolerations, you should know exactly what you're allowing.
For example, with the proposed configuration, nodeagent Pods will also run on the master Node.

I would propose the following:
* Add tolerations for common keys only (e.g., `nvidia.com/gpu`). I think the NVidia GPU one will solve 90% of the use-cases and the rest will need to patch the configuration with the values they need.
* Put any changes in an overlay.

What do you think?

jlewi:
@yanniszark IIUC your comments would apply to the changes on master?

To be clear, I consider myself to be a bystander because for GCP we aren't using the ISTIO 1-3-1 configs and our plan is to move to Anthos Service Mesh and get of the business of maintaining ISTIO configs.

krishnadurai:
Note: #923 was the original issue.

@yanniszark it does make sense to limit the tolerations to the specific key `nvidia.com/gpu=present`. You have mentioned the unintended consequences of not limiting the tolerations.

@jlewi this should apply to master as well.

I'll make the changes required in master.

cc: @rafaelbarreto87

PR #969, opened 2020-03-02
Fix pipeline default service account

Bobgy:
Resolves https://github.com/kubeflow/kubeflow/issues/4803

1. Use `kf-user` as pipeline default sa.
2. Bind `pipeline-runner` clusterrole to `kf-user` to avoid breaking existing usages.

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/969)

### DIFF
--- a/pipeline/pipelines-runner/base/cluster-role-binding.yaml
+++ b/pipeline/pipelines-runner/base/cluster-role-binding.yaml
@@ -12,3 +12,4 @@ subjects:
 # temporarily switched to kf-user, because pipeline-runner isn't bound to workload identity by default
 - kind: ServiceAccount
   name: kf-user
+  namespace: kubeflow
--- a/pipeline/pipelines-runner/base/cluster-role-binding.yaml
+++ b/pipeline/pipelines-runner/base/cluster-role-binding.yaml
@@ -9,3 +9,6 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: pipeline-runner
+# temporarily switched to kf-user, because pipeline-runner isn't bound to workload identity by default
+- kind: ServiceAccount
+  name: kf-user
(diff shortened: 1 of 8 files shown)


Bobgy:
/retest
The error doesn't seem related

Bobgy:
Strange, I got error: ```
util.py                     72 INFO     failed to apply:  (kubeflow.error): Code 500 with message: kfApp Apply failed for kustomize:  (kubeflow.error): Code 500 w
ith message: Apply.Run  Error error when creating "/tmp/kout809907334": ClusterRoleBinding.rbac.authorization.k8s.io "pipeline-runner" is invalid: subjects[1].nam
espace: Required value
```

I guess when referring to a service account not defined in the same kustomize package, it must have a namespace.

**Comment by Bobgy** [comment]
I'm not sure this is the right fix, because it seems https://github.com/kubeflow/kfctl/search?q=kf-user&unscoped_q=kf-user `kf-user` sa only exists in gcp and aws clusters.

I think we'd need a solution to let each application flexibly define what service account needs what binding in their yamls. KCC can support that for GCP, we'd need something similar.

**Comment by Bobgy** [comment]
The error message seems irrelevant, will retry again.
/retest

File: pipeline/pipelines-runner/base/cluster-role-binding.yaml [hunk 1]
```diff
@@ -9,3 +9,7 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: pipeline-runner
+# temporarily switched to kf-user, because pipeline-runner isn't bound to workload identity by default
+- kind: ServiceAccount
+  name: kf-user
+  namespace: kubeflow
```
Bobgy: @jlewi  Should I use var replace here?

Bobgy:
The failed step cannot pull image: gcr.io/kubeflow-ci/test-worker-py3:789005d.
I checked on gcr.io, the image doesn't exist, seems like we are pinning to an invalid image.

gabrielwen:
@Bobgy the test is fixed :)

Bobgy:
Thanks! @jlewi @gabrielwen

PR #971, opened 2020-03-02
Experimental GCP config for kubeflow setup

kunmingg:
* This PR contains ASM config entry for kubeflow.
  * IstioControlPlane CR from ASM.
  *  CNRM resources mapped from current kuebflow DM entry
* update istio clusterRbacConfig mode to `ON_WITH_EXCLUSION `:
   * Now ASM create namespace "isitio-system", it turns on side car for this namespace which requires user having "istio policy allow" to access central UI. With turning rbac off for "isitio-system", we can continue to make shared UIs accessible by default 

Related to: kubeflow/kubeflow#3650 ISTIO on Private GKE

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/971)


jlewi:
What are the new DM configs for?

Can we organize all of the new GCP CNRM configs under the /gcp directory?

jlewi:
@kunmingg could you please update the PR explaining what this setup is for?

I'm still unclear about what the new DM configs under experimental/gcp are for?

Are we going to continue to use DM or switch to CNRM style resources for everything?

kunmingg:
@jlewi 
description updated

jlewi:
Thanks for the clarifiction.

Should we rename "gcp/asm" to something like "gcp/cnrm" or "gcp/kcc"? It looks like the directory contains CNRM resources for deploying Kubeflow. 

Could you add a README to this directory and explain that the config is experimental?

PR #980, opened 2020-03-04
adds grpc server as part of Dex config for programmatic access

swiftdiaries:
Related to https://github.com/kubeflow/kfctl/issues/140

Enables gRPC Server on Dex for programmatic access. This will go into a potential Kubeflow Auth library.

Dex documentation for using the Server and downstream clients:
https://github.com/dexidp/dex/blob/master/Documentation/api.md

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/980)

### DIFF
--- a/dex-auth/dex-crds/base/config-map.yaml
+++ b/dex-auth/dex-crds/base/config-map.yaml
@@ -18,6 +18,10 @@ data:
     oauth2:
       skipApprovalScreen: true
     enablePasswordDB: true
+    grpc:
+    # Cannot be the same address as an HTTP(S) service.
+    addr: 127.0.0.1:5557
+    reflection: true
     staticPasswords:
     - email: $(static_email)
       hash: $(static_password_hash)
--- a/dex-auth/dex-crds/base/config-map.yaml
+++ b/dex-auth/dex-crds/base/config-map.yaml
@@ -19,9 +19,9 @@ data:
       skipApprovalScreen: true
     enablePasswordDB: true
     grpc:
-    # Cannot be the same address as an HTTP(S) service.
-    addr: 127.0.0.1:5557
-    reflection: true
+      # Cannot be the same address as an HTTP(S) service.
+      addr: 127.0.0.1:5557
+      reflection: true
     staticPasswords:
     - email: $(static_email)
       hash: $(static_password_hash)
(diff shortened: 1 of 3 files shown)


dex-auth/dex-crds/base/config-map.yaml
```diff
@@ -18,6 +18,10 @@ data:
     oauth2:
       skipApprovalScreen: true
     enablePasswordDB: true
+    grpc:
+      # Cannot be the same address as an HTTP(S) service.
+      addr: 127.0.0.1:5557
```
krishnadurai: @swiftdiaries would it be better to expose this with `0.0.0.0:5557`?

The client service might need to connect to this dex service through this port and binding of `0.0.0.0` would allow the client service to connect across pods.

yanniszark:
@swiftdiaries thanks for this effort!
I'm a bit worried about the security implications of this.
This is a management interface and we should be careful about how we expose it.
Perhaps it can be secured by an Istio sidecar and Istio RBAC?

dex-auth/dex-crds/base/config-map.yaml
swiftdiaries: Changed service to `0.0.0.0:5557`

swiftdiaries:
@yanniszark 
This is currently experimental and the only way to access is to port-forward to the Dex pod. I have not added the gRPC service port to the Dex service definitions. We can expose the gRPC port and add Istio RBAC when we have a proper auth library. 
Does this address your concern?

yanniszark:
@swiftdiaries thanks!
Since this is experimental, what is the use-case around this case?
I'm asking because for a security feature like that, it's better to leave it off if it's not needed.
Also, can we make sure it is only listening to localhost and not 0.0.0.0? Will it work then?

swiftdiaries:
The use case is around building a common auth library for Kubeflow as is detailed in the related issue. As you can see from the related issue that there is a lot of interest around this :)

Are there any specific concerns around making 0.0.0.0?
Because if not, I think it's better to keep that going forward for exposing it in the future.

yanniszark:
@swiftdiaries thanks for the explanation. 
For 0.0.0.0, my concern is that it will be exposed in the Pod's IP.
Currently, there is no mechanism in place to prevent someone from accessing it (eg frkm their notebook).

As for the issue, can you describe how this interface will be used to solve programmatic authentication?
AFAIK Dex doesn't support a programmayic oidc flow, so how is this going to solve that?

yanniszark:
@swiftdiaries I hope I didn't deter you from this effort :sweat_smile: 
Is this no longer needed?

PR #910, opened 2020-02-13
Katib move db to overlay

discordianfish:
**Which issue is resolved by this Pull Request:**
Resolves #904 

**Description of your changes:**
similar to #673, this allows users to create their own overlay to use
external databases.

**Checklist:**
- [x] Unit tests have been rebuilt: 
    1. `cd manifests/tests`
    2. `make generate-changed-only`
    3. `make test`

...but to make the ibm-storage-config overlay working, I had to change it's base to the db overlay which seems to break the tests:
```
=== RUN   TestKatibControllerOverlaysIbmStorageConfig
2020/02/13 18:14:01 got file 'db', but '/manifests/katib/katib-controller/overlays/db' must be a directory to be a root
--- FAIL: TestKatibControllerOverlaysIbmStorageConfig (0.00s)
    katib-katib-controller-overlays-ibm-storage-config_test.go:572: Err: accumulating resources: accumulating resources from '../../overlays/db': security; file '/manifests/katib/katib-controller/overlays/db' is not in or below '/manifests/katib/katib-controller/overlays/ibm-storage-config'
```

Running `kubectl kustomize .` (v1.17.2) builds the ibm overlay properly though, including the mysql components from the db overlay. Looks like the error is caused by a security feature that was added: https://github.com/kubernetes-sigs/kustomize/blob/0075d0a88c380f22ed27c324cbeda05c0640e885/docs/FAQ.md#security-file-foo-is-not-in-or-below-bar

The FAQ recommends to share only bases, but - afaik - that would mean we can't have a overlay disabling the db. Unless I'm missing something, the only way to make this work is disabling the security check. If that's desired, I can look into doing that for kfctl and the tests.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/910)

PR #986, opened 2020-03-06
Katib move db to overlay disable loadrestrictions

### LINKED ISSUE #904: Move database from katib-controller base to overlay
Similar to #673, the database shouldn't be included in katib-controller to allow using an external database.

discordianfish:
Resolves #904

This is an **alternative** to #910 but I can't get the tests to pass because some assumptions on how the tests are built. If this is the preferred way over #910, please advise on how to change the test generation etc to make it work.

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/986)


discordianfish:
/assign @yanniszark

Since this disables the load restrictions, it probably makes sense for @jlewi to approve this. Hope doing this here is fine, given that it seems to be the only solution and http://bit.ly/kf_kustomize_v3 suggestes doing this in the future anyways

jlewi:
@discordianfish Where are you disabling the load restrictor? In the tests? That seems fine to me.

Someone on the Katib team should review overall layout of the katib package.

discordianfish:
btw, this is based upon https://github.com/kubeflow/manifests/pull/1055 - so we can either merge this as is or that one first and I'll rebase, up to you.

jlewi:
Thanks. I LGTM'd #1055.
You should also be able to do `make generate-changed-only` to only regenerate tests based on what changed but its imperfect.

discordianfish:
@jlewi Yes, that created a mess on this PR that failed tests.
Rebased this, let see what the tests say.

discordianfish:
No tests? Hrm..

/retest

discordianfish:
Ok, tests passed. @yanniszark can you review?

jlewi:
/assign @gaocegege 

@gaocegege Does someone from Katib want to take a look at this?

katib/katib-controller/base/katib-db-manager-deployment.yaml
```diff
@@ -24,14 +24,6 @@ spec:
       - name: katib-db-manager
         image: gcr.io/kubeflow-images-public/katib/v1alpha3/katib-db-manager
         imagePullPolicy: IfNotPresent
-        env:
```
andreyvelich, line 27: I am not sure that we can move Katib mysql DB to the overlay currently.
Katib db manager will not work without running Katib mysql deployment.

In that case, if user wants to deploy Katib using the base from manifests repository without other Kubeflow components and overlays, how he can do it?

What do you think @gaocegege @johnugeorge @richardsliu ?

discordianfish: @andreyvelich I'm following the same pattern here as in https://github.com/kubeflow/manifests/pull/673 - if this isn't a problem there, why is is a problem here?

Yes, the/some db overlay is required but this is required in many other manifests as well. What is the problem with that?

andreyvelich: I am no sure how metadata connects with the DB, but Katib DB Manager can't work without mysql DB.
If users want to use their own DB (not mysql), they must create custom DB manager.

@johnugeorge @jlewi Do we have use-case when user deploys Katib from Manifest repository from the base folder, or in that case user will use manifest folder from the Katib repository?

discordianfish: @andreyvelich Yes, that's true. A user either needs to use the db overlay here, to get the bundled database, or a custom overlay for their custom database setup.

Maybe it helps to clarify by providing our overlay as example (it uses RDS but that shouldn't matter):

*kustomization.yaml:*
```
kind: Kustomization
bases:
- ../../base
patchesStrategicMerge:
- katib-db-manager-deployment.yaml
```

*katib-db-manager-deployment.yaml:*
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: katib-db-manager
spec:
  template:
    spec:
      containers:
      - name: katib-db-manager
        env:
          - name: DB_NAME
            value: mysql
          - name: KATIB_MYSQL_DB_DATABASE
            value: katib_production
          - name: KATIB_MYSQL_DB_HOST
            valueFrom:
              secretKeyRef:
                name: db-secrets
                key: MYSQL_HOST
          - name: DB_USER
            valueFrom:
              secretKeyRef:
                name: db-secrets
                key: MYSQL_USER_NAME
          - name: DB_PASSWORD
            valueFrom:
              secretKeyRef:
                name: db-secrets
                key: MYSQL_ROOT_PASSWORD
```

andreyvelich: In that case, should we also move Katib DB Manager to the overlay?
For example, if user wants to use not mysql DB ? How he can do it?

discordianfish: @andreyvelich Why do you think the db-manager should be moved to the overlay? As it is right now, a user can use a similar patch like the one above but configure a different database.

andreyvelich: @discordianfish As I said, currently Katib DB Manager can't work with not mysql database.
For example, what if user wants to use PostgreSQL? How he can patch the base directory ?

johnugeorge: @andreyvelich Why do you need db-manager in the overlay?

johnugeorge: I am not sure of the best practice here. With this patch, we cannot apply from base folder as database configs are available only in overlay.

discordianfish: @andreyvelich Got it. Well, then they can't? Or do we want to support people writing custom katib db manager? But now that the loadrestrictions are disabled, I could try to also move the db-manager to a deployment and have the db overlay depend on it.

@johnugeorge What is the problem with this? the same applies to other components, like the metadata deployment where the db was moved to an overlay as well.

@jlewi Can you suggest an direction here?

andreyvelich: @discordianfish I think we want to support different DB backend in the future. That is why we created Katib DB manager.
Do other components, like metadata, can work without this DB?

As I said before, Katib DB Manager and DB, currently, is the core component for Katib. Users can use Katib without it, but they have to create their own DB backend.

/cc @gaocegege

Jeffwan: To install katib standalone,  people need to use `kustomize build  overlays/db` after this change.

jlewi: @discordianfish I'm having trouble understanding the thread. 

Ultimately what is the goal of the PR? I.e. what is the reason to move the db to an overlay? My inference from #904 is that the goal is to allow external DBs to be configured.

@andreyvelich and @johnugeorge what are the configurations we want to support? i.e.

* Katib backed Katibe specific MySQL in cluster
* Katib backed by external DB (MySQL) etc...?
* Others?

Can we list out the configurations we want to support and then figure out how to arrange the package to maximize code sharing?

discordianfish: @jlewi 
> My inference from #904 is that the goal is to allow external DBs to be configured.

That is correct, IMO it would make sense to not broaden the scope to support possible yet-to-be implemented databases.

andreyvelich: > 
> @andreyvelich and @johnugeorge what are the configurations we want to support? i.e.
> 
> * Katib backed Katibe specific MySQL in cluster
> * Katib backed by external DB (MySQL) etc...?
> * Others?
> 

I think, currently, we want to support these configurations:

1. Installing Katib from manifest folder with all components (Controller, DB Manager, MySQL DB)
2. Installing Katib with custom settings for MySQL DB. In that case, user can use Katib DB Manager, but create own yaml for MySQL DB.
3. Installing Katib with custom DB (e.g. PostgreSQL, ModelDB). In that case, user has to set configuration for Katib Controller how to call DB Manager. Here is the tutorial what needs to be changed: https://www.kubeflow.org/docs/components/hyperparameter-tuning/env-variables/#katib-controller.

/cc @johnugeorge

jlewi: @andreyvelich and @discordianfish sounds like there is agreement on support for custom DB. 

So is the issue just the question about how to organize the kustomize manifests to support all 3 options?

discordianfish: If everyone is happy with moving the (mysql-)katib-db-manager to it's own overlay, which would depend on the db overlay, then I'm happy to change the PR that way.
Does that sound good?

andreyvelich: @discordianfish You want to create overlay with `katib-db-manager` which would depend on overlay with appropriate DB, for example mysql DB, right?

discordianfish: @andreyvelich I've moved the katib-db-manager to an overlay. I can't make it depend on the db overlay though, since that would make it impossible to use the db-manager overlay with a custom db overlay (that might be named different).

To address the use cases from above:

1. Install from provided kfdef
2. Remove db overlay from kfdef and add custom overlay to deploy custom mysql or patch db-manager to use external mysql
3. Remove db overlay and db-manager overlay and add custom overlay to patch katib deployment and/or deploy custom db-manager

discordianfish:
Now the tests fail. Hopefully #1104 will fix this.

Can we avoid broaden the scope of a two month old PR every time? This doesn't seem to be a good strategy, given how much in flux everything else it.

katib/katib-controller/base/katib-db-manager-deployment.yaml
andreyvelich: SGTM. Should we rename overlay `db` to `db-mysql` what do you think @discordianfish?

/cc @gaocegege @johnugeorge

discordianfish: @andreyvelich done

johnugeorge: @jlewi Yes you are right. lets discuss this in community meeting and get this to closure.

jlewi: Would a composition pattern work better here? It looks like we have three sub-applications

* Katib controller (+ associated resources) e.g. RBAC
* DB manager 
* A MySQL itself.

What if each of those components was their own Kustomize package and then we used Kustomize to compose those applications into different configurations so we'd have

* One for standalone - Which would combine all three and add an overlays to configure DB Manager to work with MySQL
* One for an external DB which would combine the first two
   * I'm guessing user would have to add in their own overlays to specify custom information

To put this another way, with these changes why is the MySQL DB an overlay on the "base" package? With the changes in this PR the base package is defining the controller and some other core resources. It doesn't look like the MySQL DB is actually modifying any of those resources.

Should we do the composition in KFDefs or should we define a new kustomize package e.g. "standalone" which combines istio, application, db-manager, and db-mysql packages?

andreyvelich: @jlewi If we create three sub-application, each of them will have base, right? 
Like we did with crd: https://github.com/kubeflow/manifests/tree/master/katib/katib-crds.
In that case, how user can install Katib standalone?

Run `kustomize build` in each folder:
1. katib-controller
2. db-manager
3. db-mysql
4. katib-crds 

Right?

discordianfish: Let me know what you decide, until then I'll put this on hold and and we'll change our strategy to maintaining a fork of manifests instead of just adding our overlays.

jlewi: @andreyvelich You can use kustomize to compose kustomize packages. A resource can be a path to a folder containing a kustomization.yaml file. You can have multiple such folders listed in resources provided each unique resource is defined only in folder. (You would use patches to modify a resource).

So you can define a standalone package that would incorporate the other three.

The kustomize would look like

```
resources:
- <path>/<to>/katib-controller
- <path>/<to>/db-manager
- <path>/<to>/db-mysql
```

andreyvelich: @jlewi I am fine with this approach. Then as @jlewi mentioned before, we can create overlays to patch DB Manager deployment with appropriate DB.
What do you think @johnugeorge @richardsliu @gaocegege ?

discordianfish:
Replaced by #1113 1141

PR #989, opened 2020-03-08
Update DBConfig.GroupConcatMaxLen in ml-pipeline apiserver config map

### LINKED ISSUE #988: Update ml-pipeline-config ConfigMap to include GroupConcatMaxLen
Some time ago, https://github.com/kubeflow/pipelines/issues/2310 (in which metrics were not showing up properly in KFP UI) has been fixed by https://github.com/kubeflow/pipelines/pull/2497. However, the needed configuration update didn't make to the manifests. Therefore, as of now, the issue still persists in Kubeflow deployment. We need to include DBConfig.GroupConcatMaxLen in the ml-pipeline apiserver ConfigMap.

rafaelbarreto87:

Resolves #988.

Some time ago https://github.com/kubeflow/pipelines/issues/2310
has been reported on pipelines related to a problem in which metrics
were not being properly loaded in the UI. The problem has been fixed by
https://github.com/kubeflow/pipelines/pull/2497. However, the manifests
didn't get updated and, therefore, the issue continues in the standard
deployment of Kubeflow. This PR just updates the corresponding ConfigMap
manifest.

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/989)


Bobgy:
@rafaelbarreto87 Can you please update merge conflict
/assign @Bobgy

rafaelbarreto87:
@Bobgy done!

Bobgy:
Thanks for the fix!
/approve
/lgtm

ConverJens:
@Bobgy @rafaelbarreto87 
I'm using KubeFlow istio dex 1.1.0 and run into an issue related to gRPC limit. Unfortunately, it seems as this config map is no longer used and hence this limit can no longer be changed.

Bobgy:
@ConverJens it's also configurable via env var DBCONFIG_GROUPCONCATMACLEN

ConverJens:
@Bobgy I tested that as well yesterday, adding:
-name: DBCONFIG_GROUPCONCATMAXLEN
  value: "8388608"
under upstream/base/pipeline/ml-pipeline-apiserver-deployment.yaml and I can exec into the pod and verify that the value is indeed set. The problem is that I still get:
```{"error":"grpc: received message larger than max (5429435 vs. 4194304)","message":"grpc: received message larger than max (5429435 vs. 4194304)","code":8}```
in the UI so the limit doesn't change.

I removed and redeployed all pipeline components as well as the mysql DB but this didn't change anything.

To be on the safe side I exec:ed into the api-server and ran:
`sed -i 's/4194304/8388608/g' config.json`
which updated config.json and viper should then repopulate this field but still no difference.

Following a comment in this issue https://github.com/kubeflow/pipelines/issues/2310, I also tested running:
`set GLOBAL group_concat_max_len=8388608;`
but this had no effect either. 

Do I need to set this value somewhere else or is something else needed to ensure this limit is picked up? Where does this limit actually originate from, i.e. where is the actual limit set?

Bobgy:
Hi @ConverJens, the config is for mysql, not gRPC.

I think the size limit you care about was already increased to maximum in later versions of KFP.
https://github.com/kubeflow/pipelines/blob/d9c019641ef9ebd78db60cdb78ea29b0d9933008/backend/src/apiserver/main.go#L95

Can you consider upgrade?

ConverJens:
@Bobgy Right, so the settings only affect the DB but in v1.0.0 of KFP the limit is hardcoded in the gRPC service, right? 
We are planning an upgrade to KF 1.2.0 later this spring but I'll try to just bump KFP in the mean time.

Thanks for the help and clarification!

ConverJens:
@Bobgy A bit of bad news: we are already running KFP 1.0.0 (gcr.io/ml-pipeline/api-server:1.0.0) and the change you mention seems to be present in that version: https://github.com/kubeflow/pipelines/blob/1.0.0/backend/src/apiserver/main.go#L89.

Are you aware of some other newer change that we need to pull in?

Bobgy:
Hi @ConverJens, that's strange. Would you mind submitting a new issue in KFP repo?

ConverJens:
@Bobgy Posted issue here: https://github.com/kubeflow/pipelines/issues/5440

PR #990, opened 2020-03-09
update GKE config for private gke setup

kunmingg:

* Update gke config for private cluster setup
* Provide kfdef entry for private GKE

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/990)

### DIFF
--- a/kfdef/kfdef_gcp_private_dex.yaml
+++ b/kfdef/kfdef_gcp_private_dex.yaml
@@ -1,0 +1,306 @@
+# Kfdef config to install Kubeflow with private GKE cluster.
+# After "kfctl build", edit "gcp_config/cluster-kubeflow.yaml" to enable private options.
+
+# Dex + Istio configured for user authentication, connect to istio-ingressgateway and login for UI access.
+
+apiVersion: kfdef.apps.kubeflow.org/v1
+kind: KfDef
+metadata:
+  namespace: kubeflow
+spec:
+  applications:
+    - kustomizeConfig:
+        repoRef:
+          name: manifests
+          path: application/application-crds
+      name: application-crds
+    - kustomizeConfig:
+        overlays:
+          - application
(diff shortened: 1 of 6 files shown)


kfdef/kfdef_gcp_private_dex.yaml
```diff
@@ -0,0 +1,306 @@
+# Kfdef config to install Kubeflow with private GKE cluster.
```
zhenghuiwang: Should we follow the pattern in https://github.com/kubeflow/manifests/tree/master/kfdef to have a `kfdef_gcp_private_dex.v1.0.0.yaml`?

Same for the private_minimum.yaml

kunmingg: How about publishing `v1.0.0` in separate PR, when the feature is more stable?

zhenghuiwang: > How about publishing v1.0.0 in separate PR

SG.

> when the feature is more stable

Do we expect more changes to this file or we just need to test it?

kunmingg: Might need more change around configs

jlewi:
Why continue to invest in Deployment Manager as opposed to switching to KCC?

abhi-g:
My only concern is with stability of the deployment process. With 1.0 we'd
like to ensure that the users don't have to continuously feel like things
underneath are moving so fast that couldn't keep pace. Also, upgrades will
need to ensure that they continue to work through these changes.

jlewi:
@abhi-g What is the consistency you are going for and how does using DM help you create a consistent experience? Are you planning on switching to ASM?

jlewi:
Closing in favor of #1126

PR #997, opened 2020-03-11
Move pipeline runner KSA override to use-kf-user overlay

Bobgy:
Reported in https://kubeflow.slack.com/archives/CE10KS9M4/p1583870376155300
Fix bug affecting platforms other than GCP introduced in https://github.com/kubeflow/manifests/pull/969, because `kf-user` service account doesn't exist for other deployments.

Move `pipeline-runner` override to `kf-user` service account to `use-kf-user` overlay to only apply for GCP.

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/997)


Bobgy:
/hold
synced with @Jeffwan, `kf-user` also makes sense for AWS, because it will have default bindings like workload identity in GCP.
I will rename current overlay to `use-kf-user` to potential future reuse.

Bobgy:
/unhold
Updated overlay name

Bobgy:
@jlewi Can you help me review this?
Is there a list which deployments will have `kf-user` and which won't?

Bobgy:
/assign @jlewi 
Because more people are asking about this e.g. https://kubeflow.slack.com/archives/CE10KS9M4/p1583870376155300
Can we get this in?

jlewi:
@Bobgy and @Jeffwan can someone help me understand what the problem is?

What does the kf-user service account have to do with GCP or other specific platforms? I would have thought the kf-user service account is created on all platforms?

What pods are running using DefaultPipelineRunnerServiceAccount? Is this the service account used by the pods in the argo workflow or is this one of the shared services as part of the pipelines backend?

If its the pods in the argo workflow then why isn't "kf-user" always the correct service account?

Bobgy:
@jlewi Looks like `kf-user` is only created in aws and gcp deployments: https://github.com/kubeflow/kfctl/search?q=kf-user&type=. @Jeffwan probably knows more about context.

> What pods are running using DefaultPipelineRunnerServiceAccount? Is this the service account used by the pods in the argo workflow or is this one of the shared services as part of the pipelines backend?

They are pods in the argo workflow.

> If its the pods in the argo workflow then why isn't "kf-user" always the correct service account?

Because in some deployments, `kf-user` doesn't exist.

jlewi:
@Bobgy thanks. The "kf-user" service account exists in profile created namespaces. So I think once we move user's pipelines into profile namespaces (as opposed to KF namespaces) that would be the norm for all packages.

jlewi:
You can commit this as is by removing the hold but wanted you to see my comment.

/hold

It looks like with the  overlay the runner value ends up getting set in two places. 

* It is set in the configmap
* It is overwritten by the environment variable

It would be a little cleaner if the value got set in a single place otherwise its more difficult to reverse engineer how its being set and change it if you want to override it. ldeally kustomize would let us apply strategic patches to the contents of the confimap but I don't know if that is supported today.

/lgtm
/approve

Bobgy:
> It looks like with the overlay the runner value ends up getting set in two places.

> It is set in the configmap
It is overwritten by the environment variable
It would be a little cleaner if the value got set in a single place otherwise its more difficult to reverse engineer how its being set and change it if you want to override it. ldeally kustomize would let us apply strategic patches to the contents of the confimap but I don't know if that is supported today.

Thanks for pointing out! That was exactly a problem. I didn't do it because it hasn't been supported by kustomize: https://github.com/kubernetes-sigs/kustomize/issues/680#issuecomment-577865951

Bobgy:
/unhold
I will follow up with configmap override when kustomize supports that

PR #1011, opened 2020-03-16
Fix the aws vpc over manifest.

jlewi:
* It looks like there is a bug in the VPC overlay for AWS. It looks like
  VPC is a patch but it is listed as a resource.

* This PR changes the kustomization.yaml file to list VPC.yaml as a patch
  and not a resource.

* In kustomize "bases" is deprecated. The newer syntax is to just list bases as resources.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/1011)

### DIFF
--- a/aws/aws-alb-ingress-controller/overlays/application/kustomization.yaml
+++ b/aws/aws-alb-ingress-controller/overlays/application/kustomization.yaml
@@ -1,8 +1,7 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
-bases:
-- ../../base
 resources:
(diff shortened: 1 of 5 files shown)


Jeffwan:
Em. nice catch. This is patch. 

/lgtm
/test kubeflow-manifests-presubmit

jlewi:
@Jeffwan Tests are fixed; would you mind re-LGTM'ing?

jlewi:
@Jeffwan ping could you take a look at this please?

Jeffwan:
@jlewi Just see this. Thanks for the fix.

Issue #1012, opened 2020-03-16
Presubmits are failing; workflow names are too long


jlewi:
Here's a failing presubmit
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubeflow_manifests/1011/kubeflow-manifests-presubmit/1239560778768453632/

The workflow fails to be submitted

```

INFO|2020-03-16T14:35:59|/src/kubeflow/testing/py/kubeflow/testing/util.py|72| INFO|2020-03-16T14:35:59|/src/kubeflow/testing/py/kubeflow/testing/run_e2e_workflow.py|92| Importing path kubeflow.kfctl.testing.ci.kfctl_e2e_workflow
INFO|2020-03-16T14:36:00|/src/kubeflow/testing/py/kubeflow/testing/util.py|72| INFO|2020-03-16T14:36:00|/src/kubeflow/testing/py/kubeflow/testing/e2e_tool.py|32| Dumping workflow to /tmp/tmpFiPgPg
Traceback (most recent call last):
  File "/usr/lib/python2.7/runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/src/kubeflow/testing/py/kubeflow/testing/run_e2e_workflow.py", line 532, in <module>
    final_result = main()
  File "/src/kubeflow/testing/py/kubeflow/testing/run_e2e_workflow.py", line 522, in main
    return run(args, file_handler)
  File "/src/kubeflow/testing/py/kubeflow/testing/run_e2e_workflow.py", line 388, in run
    body=wf_result)
  File "/usr/local/lib/python2.7/dist-packages/kubernetes/client/apis/custom_objects_api.py", line 178, in create_namespaced_custom_object
    (data) = self.create_namespaced_custom_object_with_http_info(group, version, namespace, plural, body, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/kubernetes/client/apis/custom_objects_api.py", line 277, in create_namespaced_custom_object_with_http_info
    collection_formats=collection_formats)
  File "/usr/local/lib/python2.7/dist-packages/kubernetes/client/api_client.py", line 321, in call_api
    _return_http_data_only, collection_formats, _preload_content, _request_timeout)
  File "/usr/local/lib/python2.7/dist-packages/kubernetes/client/api_client.py", line 155, in __call_api
[...]

PR #1013, opened 2020-03-16
Use a configmap generator for the spawner_ui_config.yaml

### LINKED ISSUE #1010: jupyter web app config is defaulting to an old image
The default is set here
https://github.com/kubeflow/manifests/blob/4d8b1cb7e6340a69303ce59878fc29f6b3a71a72/jupyter/jupyter-web-app/base/config-map.yaml#L24

To an old 0.7 image. It should probably be updated to one of the images listed in the configmap below.
e.g.

https://github.com/kubeflow/manifests/blob/4d8b1cb7e6340a69303ce59878fc29f6b3a71a72/jupyter/jupyter-web-app/base/config-map.yaml#L27

@kimwnasptd could you submit a fix for this please?

jlewi:
* Using a configmap generator is a bit cleaner because we can store
  the configuration file in a external file rather than inlining it in
  the config map.

* Using a configmap generator also has the advantage that kustomize will
  generate a unique name based on a hash of the content.

  * This has the advantage that if a user updates the config, the hash
    would change so deployments would automatically be redeployed in order
    to pick up the config changes.

* Also update the default image used in the configmap to fix #1010

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/1013)


jlewi:
/assign @kimwnasptd 

@kimwnasptd There are two issues with the tests

#1012 presubmits are failing because the workflow names are too long.

The unittests are not passing after regenerating the unittests based on this change. It looks like a bug in how the generate scripts are generating the expected value for the configmap.

```
   ---                                                                                        ---
   apiVersion: v1                                                                             apiVersion: v1
   data:                                                                                      data:
X    spawner_ui_config.yaml: |2-                                                                spawner_ui_config.yaml: |-

```

The left hand side is the expected value. I don't know why its

```
 spawner_ui_config.yaml: |2- 
```

I'm not going to try to fix that until after fixing #306. The fix to #306 should simplify the logic for generating the expected value. In particular we will run `kustomize build -o ...` to generate the expected value. Running kustomize build is generating correct output so that should fix the problem.

In the meantime it would be helpful if you could take a look at this PR so once those issues are fixed we could go ahead and merge it.

kimwnasptd:
Thanks for the PR @jlewi! to make sure I'm in the same page, until we resolve #306 if I would like to make some change in the configmap would I also need to change the hardcoded expected value in the go files?

Or until #306 is resolved we expect the unit tests for the config map to *not* pass?
Is my understanding correct or am I missing something?

jlewi:
@kimwnasptd #306 will be fixed by kubeflow/manifests#1016; my plan is to merge that first; fix the tests and then merge this.

In the meantime you can review this to see if there are any issues with switching to a generator for the configmap.

jlewi:
@kimwnasptd #1016 was merged and I rebased and updated the tests. I ran the tests locally and they passed so I expect presubmits will pass.

Could you PTAL?

kimwnasptd:
The tests also pass on my local machine.
Two things I noticed:
1. After running `kustomize build jupyter/jupyter-web-app/overlays/application` the generated deployment's ENV Vars `USERID_{HEADER,PREFIX}`  are set to "". Are they substituted later on with `kfctl`?
2. After running `kustomize build jupyter/jupyter-web-app/overlays/istio` the `destination` field in the virtual service's spec has `$(namespace)` in the generated YAML. After applying the yaml I get the error 
```
Error from server: error when creating "STDIN": admission webhook "pilot.validation.istio.io" denied the request: configuration is invalid: domain name "$(namespace).svc.$(clusterDomain)" invalid (label "$(namespace)" invalid)
```

jlewi:
I think my most recent PR broke the unittests because I changed the config map but failed to regenerate the tests.
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubeflow_manifests/1013/kubeflow-manifests-presubmit/1242889171895324672/

Prior to that PR the E2E tests were failing because I wasn't properly setting the name of the config map so that the value in the deployment matched the original name of the generated config map.

So hopeful my latest fix will fix this.

jlewi:
@kimwnasptd tests passed could you LGTm please?

Issue #1014, opened 2020-03-16
unittests should test kustomizations


jlewi:
Right now our unittests are basically running "kustomize build" and comparing that to some expected output.

We basically do this recursively for all "kustomization.yaml" files in the repo.

This isn't really effective. What we want to test with unittests is that various kustomizations generate the expected output. Examples of kustomizations we want to validate

* That patches are applied and generate expected output
* That multiple kustomization packages are correctly combined
* That vars are correctly substituted

As a result validating the output of "kustomize build" doesn't really make sense for low level kustomization files that aren't really applying any kustomizations.

As an example:
https://github.com/kubeflow/manifests/blob/master/tf-training/tf-job-crds/base/kustomization.yaml

The output of "kustomize build" would just be CRD.yaml; there aren't really any kustomizations applied. So running "kustomize build" to generate a golden set of expected manifests and comparing to that is kind of pointless.

Instead our tests should probably doing the following.

* We should have test cases corresponding to applying different kustomizations
   * These should correspond to "kustomization.yaml" files checked into source control some where (e.g. under tests/)
* We should run "kustomize build" to generate the YAML manifests from these manifests and check them in 
   * This makes it easy for reviewers to verify the expected output is correct when we change our       kustomize packages
* The unittests should compare the results of running "kustomize build" to the expected output to ensure the code is in sync with the expected output

Follow on to:

kubeflow/manifests#306 - How do we verify the expected value is correct

jlewi:
As part of the kustomize_v3 refactor we will need to refactor a lot of the kustomize packages.
Right now the way our tests are structured its not easy to verify that the refactor doesn't make any changes to the existing packages.

This is because we don't actually have kustomizations similar to the ones generated by kfctl.

It would be useful to start checking in kustomizations representative of the ones generated by kfctl and adding unittests for each application so that we could easily verify they don't change.

I think for every application we basically want to generate a "legacy" kustomize package that combines the ISTIO and application overlays.

We'd have to change our test generation logic; since we would no longer want to recourse through all manifest directories.

Instead we might want to setup a pattern where we check in kustomization.yaml files into the test directory.

jlewi:
Fixed by #1066

PR #1016, opened 2020-03-16
unittests should compare result of kustomize build to golden set of YAML resources.

### LINKED ISSUE #683: manifests test structure should mirror manifest directory tree
The unittests for every kustomize package is currently in
https://github.com/kubeflow/manifests/tree/master/tests

i.e. the manifest  `a/b/c/kustomization.yaml' results in test

`a-b-c_test.go`.

I think a better approach would be to mirror the manifests directory tree under /tests; e.g.

`a/b/c/kustomization.yaml` maps to 

`tests/a/b/c/test.go`

One advantage of giving each kustomization manifest its own test directory is that we could more easily store other test files associated with that manifest. For example, we could store a golden set of YAML files corresponding to the expected resources (see #306).

In #665 I tried that but modifying 
https://github.com/kubeflow/manifests/blob/fde69d7217380fabb4c6f4682988fc781ebfacb9/hack/generate_tests.py#L22

I ran into problems getting the resulting go tests to compile.  In particular, I go.mod errors complaining about version conflicts for third party dependencies. I'm not sure why that was since none of the third party dependencies should have changed.

/cc @kkasravi

jlewi:
unittests should compare result of kustomize build to golden set of YAML resources.

* Per kubeflow/manifests#306 to allow reviewers to verify that the expected
  output is correct we should check in the result of "kustomize build -o"
  so that reviewers can review the diff and verify that it is correct.

* This also simplifies the test generation code; the python script
  generate_tests.py just recourses over the directory tree and runs "kustomize build -o" and checks in the output into the test_data directory.

* This is different from what the tests are currently doing.

  * Currently what the generation scripts do is generate "kustomization.yaml" files and then generate the expected output from that when the test is run.

  * This makes it very difficult to validate the expected output and to
    debug whether the expected output is correct.

  * Going forward, per #1014, I think what we want to do is check in test cases
    corresponding to kustomization.yaml files corresponding to various
    kustomizations that we want to validate are working correctly

  * Our generate scripts would then run "kustomize build" to generate expected
    output and check that in so that we can validate that the expected output
    is correct.

* Also change the tests data structure so that it mirrors the kustomize directory tree rather than flattening the tests into the "tests" directory.

  * Fix #683

* Right now running the unittests takes a long time

  * The problem is that we generate unittests for every "kustomization.yaml"
    file
  * Per #1014 this is kind of pointless/redundant because most of these
    tests aren't actually testing kustomizations.
  * We will address this in follow on PRs which will add more appropriate
    tests and remove some of these unnecessary/redundant tests.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/1016)


jlewi:
@krishnadurai ping? Could you take a look at this please?

johnugeorge:
Sorry for being late. Missed this one. This is a required change to test the sanity of generated config files.  

One question:

> Our generate scripts would then run "kustomize build" to generate expected
output and check that in so that we can validate that the expected output
is correct.
 
you meant, user has to validate  against the generated output to see if it is the expected ?

jlewi:
> you meant, user has to validate against the generated output to see if it is the expected ?

Correct.  

This PR is rewriting the tests so that they work by comparing the output of `kustomize build` to a set of "golden" YAML files representing the expected output. These are checked into source control.

This allows to do the following

1. We can refactor our kustomize packages and verify that the output hasn't changed because the tests should pass without having to change the YAML for the expected output

1. When we change our kustomize packages we want to verify that this has the expected effect on the generated manifests
 
    * In this case, the PR author would run `kustomize build` to generate and include in the PR the updated manifests

    * The PR author and the reviewer can then review the diff to the expected output to verify the changes have the desired effect

johnugeorge:
Great. LGTM
 
Can you resolve the conflicts?

tests/test_util.go
```diff
@@ -0,0 +1,196 @@
...
+		}
+
+		actualYaml, err := r.AsYAML()
+
+		if err != nil {
+			t.Errorf("Could not generate YAML for resource: %v; error: %v", rKey, err)
+			continue
+		}
+		// Ensure the actual YAML matches.
+		if string(actualYaml) != e.yaml {
```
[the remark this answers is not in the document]
krishnadurai: Here's the diff for reference.

```
diff --git a/tests/test_util.go b/tests/test_util.go
index 9966262..a887324 100644
--- a/tests/test_util.go
+++ b/tests/test_util.go
@@ -96,9 +96,12 @@ func RunTestCase(t *testing.T, testCase *KustomizeTestCase) {
                        t.Errorf("Could not generate YAML for resource: %v; error: %v", rKey, err)
                        continue
                }
+
                // Ensure the actual YAML matches.
-               if string(actualYaml) != e.yaml {
-                       ReportDiffAndFail(t, actualYaml, e.yaml)
+                stringActualYaml := strings.Replace(string(actualYaml), "\r\n", "\n", -1)
+                stringExp := strings.Replace(e.yaml, "\r\n", "\n", -1)
+               if stringActualYaml != stringExp {
+                       ReportDiffAndFail(t, stringActualYaml, stringExp)
                }
        }

@@ -146,11 +149,11 @@ func (r *expectedResource) Key() string {
 }

 // Pretty printing of file differences.
-func ReportDiffAndFail(t *testing.T, actual []byte, expected string) {
+func ReportDiffAndFail(t *testing.T, actual string, expected string) {
        sE, maxLen := convertToArray(expected)
-       sA, _ := convertToArray(string(actual))
+       sA, _ := convertToArray(actual)
        fmt.Println("===== ACTUAL BEGIN ========================================")
-       fmt.Print(string(actual))
+       fmt.Print(actual)
        fmt.Println("===== ACTUAL END ==========================================")
        format := fmt.Sprintf("%%s  %%-%ds %%s\n", maxLen+4)
        limit := 0
```

hack/generate_tests.py
```diff
@@ -2,26 +2,54 @@
...
 
 import argparse
+import jinja2
 import logging
 import os
+import shutil
 import subprocess
 
-TOP_LEVEL_EXCLUDES = ["docs", "hack", "tests", "kfdef"]
+TOP_LEVEL_EXCLUDES = ["docs", "hack", "kfdef", "stacks", "tests"]
```
krishnadurai: Would kfdef need separate handling for testing through this mechanism?
It would be great if we can have kfdefs built and tested through this harness.

jlewi: I think partially and I think that relates to kubeflow/kfctl#57.

If we think of KFDef and kfctl as a shorthand for generating a much larger more verbose set of declarative configs then we can follow a similar approach of testing that KFDef generates the expected configs. i.e. we could run `kfctl build` (but not `kfctl apply` and compare the actual to expected configurations. 

So the general approach would work but the mechanics would be different (i.e. we wouldn't do kustomize build).

tests/test_util.go
```diff
@@ -0,0 +1,196 @@
...
+
+		if !ok {
+			t.Errorf("Expected resources missing resource: %v", rKey)
+			continue
+		}
+
+		actualYaml, err := r.AsYAML()
+
+		if err != nil {
+			t.Errorf("Could not generate YAML for resource: %v; error: %v", rKey, err)
```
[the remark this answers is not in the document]
jlewi: That's a good question. I can't think of a reason we would but checking the error seems harmless and better than just swallowing it.

tests/test_util.go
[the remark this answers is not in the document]
jlewi: Good catch: I filed #1038. As far as I can tell, I'm doing the same comparison the code is currently doing.
https://github.com/kubeflow/manifests/blob/9e6ef8681003c195560fb7b3b75211830c19b7a0/tests/kusttestharness_test.go#L212

So I'd prefer to not change it in this PR. The change seems harmless but there's already enough change to the testing relative to what we currently have that I'd prefer not to introduce one more variable.

jlewi:
@krishnadurai Thanks for the review. PTAL.

hack/generate_tests.py
krishnadurai: This makes sense.

Another aspect I was thinking about testing here is our generation of versioned kfdefs through kustomize - though this will be a narrower approach than your suggestion, though I think this will help us reduce errors in defining kfdefs.

jlewi: Good point.

Issue #1036, opened 2020-03-20
kfdef on v1.0 branch broken


kunmingg:
`kfctl build https://github.com/kubeflow/manifests/blob/v1.0-branch/kfdef/kfctl_gcp_iap.v1.0.0.yaml`

throws error 

`"failed to build kfApp from URI: couldn't generate KfApp:  (kubeflow.error): Code 500 with message: kfApp Generate failed for kustomize: kustomize generate failed Error:  (kubeflow.error): Code 500 with message: couldn't copy application kfserving-gateway","time":"2020-03-20T20:49:29Z"}`

jlewi:
@kunmingg Can we close this? Is it fixed by #1037?

jlewi:
Duplicate of #1029

PR #1037, opened 2020-03-20
Revert "Automated cherry pick of #949: Adds kfserving ingressgateway …

kunmingg:
Resolves https://github.com/kubeflow/manifests/issues/1036

Remove latest change for kfserving ingressgateway 

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/manifests/1037)


richardsliu:
The cherrypick was made to fix v1.0.1 which was missing the kfserving ingressgateway manifests.

Maybe we should just deprecate v1.0.0.

jlewi:
@richardsliu It sounds like we need a more fine grained fix; i.e. leave 1.0.0 unchanged but add it to 1.0.1. But that assumes 1.0.1 hasn't been "frozen" yet.  if 1.0.1 is frozen/finalized then we should start 1.0.2RC.

In keeping with the philosophy though that we should revert first I suggest we revert and then fix.

So on my end
/lgtm
/approve

@richardsliu do you agree with that?

Jeffwan:
@jlewi @richardsliu  What's the current status? Seems master branch and v1.0.1 has the change, but v1.0-branch doesn't. This will be broken when we release v1.0.2, etc.

richardsliu:
New PR: https://github.com/kubeflow/manifests/pull/1072

Jeffwan:
Cool. @richardsliu What's the timeline for v1.0.2? Before we release, can we create an issue to collect some feedbacks on candidate needs to be cherry-picked?

PR #1049, opened 2020-03-25
centraldashboard v3 manifest

jlewi:
* Create a centraldashboard v3 manifest

* For centraldashboard define a kustomization file so that we don't need
to rely on generation by kfctl

* For centraldashboard define a patch that sets environment variables
based on a "global" kubeflow config map. This way we can have a
single configmap that is reused across applications for parameters
that should be shared across applications

Design doc: http://bit.ly/kf_kustomize_v3

Related to #774


jlewi:
/assign @yanniszark 
/assign @kunmingg 
/cc @krishnadurai 

This is ready for review. Split it from #962

jlewi:
@kunmingg could you take a look please?

PR #1050, opened 2020-03-25
manifests v3 versions for profiles and application

jlewi:
* Define v3 versions of the manifests that use the patterns as described
  in http://bit.ly/kf_kustomize_v3

* For application resource we just need to define a new v3 kustomization.yaml that
  combines all the overlays we want and doesn't define the vars that will conflict
  with the other packages

* For profiles we need to

  * Define a new base_v3 package
  * Use a configmap generator to generate a config map containing application specific parameters
  * Make the parameters available to the command line by setting environment variables to
    the config map and then referencing them in the command line.

Related to #774


jlewi:
@kunmingg @yanniszark could one of you  take a look please?

PR #1061, opened 2020-03-30
Create kustomize v3 versions of the GCP packages for istio and iap.

jlewi:
* Define v3 versions of the manifests that use the patterns as described
in http://bit.ly/kf_kustomize_v3

* For ISTIO package use kustomization.yaml to set the namespace in which
  ISTIO is installed.

* For gcp create a new kustomize package to install the IAP ingress
  * This can't be part of the GCP stack because it needs to be in the
    istio-system namespace which is different from the namespace the
    stack is installed in.

Related to #774


jlewi:
unittest failure looks like a flake contacting GitHub.

/test all

PR #1072, opened 2020-04-02
Cherry pick 949 - Kfserving gateway

richardsliu:
Manually cherry-picks PR #949. Removes the kfdef changes to not break v1.0.0.

### DIFF
--- a/istio/kfserving-gateway/base/deployment.yaml
+++ b/istio/kfserving-gateway/base/deployment.yaml
@@ -1,0 +1,168 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: kfserving-ingressgateway
+  labels:
+    app: kfserving-ingressgateway
+    kfserving: ingressgateway
+spec:
+  selector:
+    matchLabels:
+      app: kfserving-ingressgateway
+      kfserving: ingressgateway
+  template:
+    metadata:
+      labels:
+        app: kfserving-ingressgateway
+        kfserving: ingressgateway
+      annotations:
+        sidecar.istio.io/inject: "false"
+    spec:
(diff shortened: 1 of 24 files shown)


Jeffwan:
@richardsliu What's the reason kfdef name breaking the branch? Why v1.0.1 doesn't have the issue?

Jeffwan:
Not sure how CLA work.. @richardsliu Can you amend git change and use your Github?

richardsliu:
@Jeffwan The problem with kfdef is that the v1.0.0 manifests are pinned to a specific commit that does not have the kfserving gateway application. So deployments for v1.0.0 will fail. The manifests for v1.0.1 don't have this problem.

For the CLA, I'll need a confirmation from @krishnadurai.

Jeffwan:
@richardsliu Em. interesting. People use v1.0.0 kfdef should always fetch pinned manifest tar. v1.0.0 kfdef has reference to kf-gateway?  

Another questions is,  this PR revert kfdef to use kf-gateway. What's the plan for v1.0.2?

richardsliu:
Kfdef v1.0.0 uses a pinned version of manifests here: https://github.com/kubeflow/manifests/blob/master/kfdef/kfctl_gcp_iap.v1.0.0.yaml#L425. That commit does not have the kfserving gateway changes, so it will fail.

This PR will merge the kfserving gateway changes back in the branch. And for v1.0.2 we will create a new kfdef file that's pinned to a newer commit, so we won't have any problems.

krishnadurai:
@googlebot I consent.

richardsliu:
The presubmit test is blocked by https://github.com/kubeflow/kfctl/issues/296.

PR #1074, opened 2020-04-03
Sync metadata folder from master to v1-branch

zhenghuiwang:
This PR fixes the issue of v1-branch that metadata-db deployments can't be connected from other pods.

The cause is that v1-branch misses some `istio-sidecar` annotations. This PR adds them back to v1-branch.

### DIFF
--- a/metadata/OWNERS
+++ b/metadata/OWNERS
@@ -1,5 +1,4 @@
 approvers:
   - neuromage
   - prodonjs
-  - rileyjbauer
   - zhenghuiwang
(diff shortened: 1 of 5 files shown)


zhenghuiwang:
@kunmingg the above tests for `manifest v1.0-branch` are broken due to KFAM. Do you know what changes from master need to be cherry picked?

/cc @kunmingg

jlewi:
See: kubeflow/kfctl#296

I'm not sure this is the correct way to fix the issue. Per kubeflow/kfctl#296 the problem right now is that newer versions of kfctl are not backwards compatible with existing 1.0 configs because they turn on istio side car injection.

This change would be a forward fix that requires people to use newer versions of manifests with the newer versions of kfctl.

I think the ideal fix would be to figure out how to make kfctl backwards compatible; i.e. not turn on istio side car injection in the kubeflow namespace for 1.0 manifests.

PR #1075, opened 2020-04-04
v3 version of PodDefaults kustomize package

### LINKED ISSUE #1073: PodDefaults Admission webhook should use cert-manager on GCP
Right now on GCP the PodDefaults is using a bash script to generate the certificate for the admission webhook
https://github.com/kubeflow/manifests/tree/master/admission-webhook/bootstrap

We should switch to using cert-manager.

I think it makes more sense to just switch to cert-manager then upgrade bootstrap for v3.

The bootstrap package is using vars to do substitution into the configmap.

Blocking #1063

jlewi:
Make PodDefault var names unique enough to support package composition

* The PodDefault kustomize manifest defines vars for the purpose of
  ensuring transformations (e.g. namespace) get applied to custom fields
  in the MutatingWebHookCR which kustomize isn't aware of.

* The var names aren't unique enough (e.g. var named namespace) so they
  create conflicts when composing this kustomize package with other packages.

* We make the var names more unique to better support kustomization.

* Make the vars for the certificate overlay unique enough so they won't
  conflict with other vars when composing Kubeflow applications.

* Create a v3 package which combines the cert-manager overlay with the
  application resource.
  * Fixes #1073

* There is no change to the expected output of this unittests; this confirms the refactor doesn't change the output


jlewi:
#1066 needs to be submitted first and then we need to rebase this PR.

jlewi:
@yanniszark @johnugeorge This PR is ready for review.
The changes to the jupyter tests are spurious and will be cleaned up in #1076 which I will merge first.

/hold so I can merge #1076 first

/assign @yanniszark @johnugeorge

PR #1079, opened 2020-04-06
Remove istio-ingress component in kfctl_aws

Jeffwan:
Resolves #

Remove istio-ingress component in kfctl_aws

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`


richardsliu:
@Jeffwan Do you need to change the v1.0.0 manifest? It is considered already released so it is best not to change it.

Jeffwan:
@richardsliu This is kind of critical to us. I think we can not take that risk. It would be better for AWS to make the change.

PR #1083, opened 2020-04-08
Bump KFP to 0.2.5

Bobgy:
**Which issue is resolved by this Pull Request:**
Resolves https://github.com/kubeflow/manifests/issues/984
Part of https://github.com/kubeflow/kubeflow/issues/4929

**Description of your changes:**
Bump KFP to 0.2.5

**Checklist:**
- [x] Unit tests have been rebuilt: 
    1. `cd manifests/tests`
    2. `make generate-changed-only`
    3. `make test`

PR #1087, opened 2020-04-09
Update generated kfdefs

Bobgy:
Part of https://github.com/kubeflow/kubeflow/issues/4803

Update generated kfdefs from kfdef source

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

/assign @richardsliu @jlewi

### DIFF
--- a/kfdef/kfctl_gcp_iap.yaml
+++ b/kfdef/kfctl_gcp_iap.yaml
@@ -262,6 +262,7 @@ spec:
   - kustomizeConfig:
       overlays:
       - application
+      - use-kf-user
       repoRef:
         name: manifests
         path: pipeline/api-service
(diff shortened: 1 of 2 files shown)


jlewi:
Did you manually edit these files or did you run make in manifests/kfctl?
https://github.com/kubeflow/manifests/tree/master/kfdef/source

How will we make it easy to keep the release branch in sync?

#1083 changed master which is the base for the v1.0.0 and v1.0.1 packages which "breaks" them; i.e. if you run make you will end up including the user-spec manifest in the manifest for v1.0.0 and v1.0.1 which is not what we want since we don't want those to change.

Since those configs are finalized and checked into manifests we could just uncouple them from the Makefile so that they would no longer be regenerated.

Bobgy:
@jlewi I ran `hack/build_kfdef_specs.py` and restored changes to published manifests.

Bobgy:
Yes, I agree with `Since those configs are finalized and checked into manifests we could just uncouple them from the Makefile so that they would no longer be regenerated.`

PR #1098, opened 2020-04-10
Create a v3 version of the metadata package 

jlewi:
* v3 version combines the various overlays that  we need.

 * The var names need to be unique enough not to conflict.


jlewi:
Presubmit failed because of go modules problem.

jlewi:
#1103 for broken presubmits.

PR #1110, opened 2020-04-17
Add google-cloudsql overlay for metadata

zhenghuiwang:
This PR adds an overlay for metadata to use Google CloudSQL.

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

### DIFF
--- a/metadata/overlays/google-cloudsql/kustomization.yaml
+++ b/metadata/overlays/google-cloudsql/kustomization.yaml
@@ -1,0 +1,14 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+generatorOptions:
+  # name suffix hash is not propagated correctly to base resources
+  disableNameSuffixHash: true
+commonLabels:
+  kustomize.component: metadata
+configMapGenerator:
+- name: metadata-db-parameters
+  env: params.env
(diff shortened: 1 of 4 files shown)


metadata/overlays/google-cloudsql/kustomization.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+generatorOptions:
+  # name suffix hash is not propagated correctly to base resources
```
jlewi, line 4: How come you are disabling the suffix name hash? This prevents a unique hash from being added to the config map based on the content.

Including the suffix hash is preferred because then the deployment will change if the content is changed and this will trigger a pods update.

Otherwise it can be difficult to tell whether a deployment is referring to the latest version of the config.

zhenghuiwang: Agree it is a good practice to have unique suffix to indicate the version of a configMap, but the generated configmap is not properly written into the metadata-deployment.yaml file if namesuffix is enabled. (See https://github.com/kubernetes-sigs/kustomize/issues/1301)

I updated the comment in the config to include the issue link.

Issue #1062, opened 2020-03-30
Migrate all KFDefs to use Kustomize v3 patterns


jlewi:
Ref: http://bit.ly/kf_kustomize_v3
        #744

KFDef patterns should follow kf_kustomize_v3 patterns. This means

1. Applications in KFDef should not use the overlays or parameters field
1. There should be an application named "kubeflow-apps" that is a kustomize package 
1. Applications should fetch global paramaters from the configmap "kubeflow-config"
1. Applications should fetch application parameters from application specific config maps
    * Application specific config maps should be named "${APPNAME}-config"
    * This is the configmap that is patched by kfctl to contain any application parameters
https://github.com/kubeflow/kfctl/blob/5da15a0d5e77d8b956a4a6c80da5e5da0d160b07/pkg/kfconfig/types.go#L726
1. Use of vars should be minimized
     * The most common and supported use case should be to substitute parameters into
   fields that don't support referencing a configmap directly
     * In this case var names should be unique enough that they won't conflict with vars from
       other applications

We should create an example e.g. #962 and then file PRs to update all the individual KFDef.

We will also need to update all the applications.

jlewi:
Closing this issue; we have sub issues tracking the remaining work.

PR #1113, opened 2020-04-17
Add v1alpha1 XGBoostJob manifests

### LINKED ISSUE #1112: Add manifests for XGBoostJob
Part of https://github.com/kubeflow/xgboost-operator/issues/44.

terrytangyuan:
Signed-off-by: terrytangyuan <terrytangyuan@gmail.com>

Resolves #1112

This adds the manifests for v1alpha1 XGBoostJob. 

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

### DIFF
--- a/xgboost-job/xgboost-operator/base/kustomization.yaml
+++ b/xgboost-job/xgboost-operator/base/kustomization.yaml
@@ -1,0 +1,20 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- cluster-role.yaml
+- cluster-role-binding.yaml
+- crd.yaml
+- deployment.yaml
+- service-account.yaml
+- service.yaml
+namespace: kubeflow
+nameprefix: xgboost-operator-
+configMapGenerator:
+- name: parameters
+  env: params.env
+generatorOptions:
+  disableNameSuffixHash: true
+images:
+  - name: gcr.io/kubeflow-images-public/xgboost-operator
+    newName: gcr.io/kubeflow-images-public/xgboost-operator
+    newTag: 8e29825
--- a/xgboost-job/xgboost-operator/base/kustomization.yaml
+++ b/xgboost-job/xgboost-operator/base/kustomization.yaml
@@ -10,10 +10,8 @@ resources:
 namespace: kubeflow
 nameprefix: xgboost-operator-
 configMapGenerator:
-- name: parameters
+- name: xgboost-operator-config
   env: params.env
-generatorOptions:
-  disableNameSuffixHash: true
 images:
   - name: gcr.io/kubeflow-images-public/xgboost-operator
     newName: gcr.io/kubeflow-images-public/xgboost-operator
--- a/xgboost-job/xgboost-operator/base/params.yaml
+++ b/xgboost-job/xgboost-operator/base/params.yaml
@@ -1,3 +1,0 @@
-varReference:
-- path: spec/template/spec/containers/image
-  kind: Deployment
--- a/xgboost-job/xgboost-operator/base/params.yaml
+++ b/xgboost-job/xgboost-operator/base/params.yaml
@@ -1,0 +1,3 @@
+varReference:
+- path: spec/template/spec/containers/image
+  kind: Deployment
--- a/xgboost-job/xgboost-operator/overlays/application/kustomization.yaml
+++ b/xgboost-job/xgboost-operator/overlays/application/kustomization.yaml
@@ -1,0 +1,14 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+bases:
+- ../../base
+resources:
+- application.yaml
+commonLabels:
+  app.kubernetes.io/name: xgboost-operator
+  app.kubernetes.io/instance: xgboost-operator
+  app.kubernetes.io/managed-by: kfctl
+  app.kubernetes.io/component: xgboostjob
+  app.kubernetes.io/part-of: kubeflow
+  app.kubernetes.io/version: v0.6
+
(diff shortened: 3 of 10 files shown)


terrytangyuan:
I see this error when I execute `make generate-changed-only`: `json: unknown field "env"`. This is caused by different versions of kustomize. Should I modify the `env` in all the existing manifests to `envs` in a separate PR so that kubeflow/manifests could work with kustomize v3?

Related issues: https://github.com/kubernetes-sigs/kustomize/issues/1069, https://github.com/kubeflow/metadata/issues/202

terrytangyuan:
Skimming through the issues and it looks like we are migrating to kustomize v3 https://github.com/kubeflow/manifests/issues/1062. The problem here is that when I run `generate_tests.py` it will traverse through all the existing manifests. In other words, if any of them are not compatible with v3, I won't be able to generate tests for XGBoostJob successfully. 

Any suggestions here? @abhi-g @richardsliu @jlewi

xgboost-job/xgboost-operator/overlays/application/kustomization.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
jlewi, line 1: Application should not depend on base.
The application manifest is not overloading any resources defined in the base package.
So either

1) create a single package which includes the application resource
or
2) Create 2 separate packages one with the base resources and one with the application resource 
    and then create a 3rd package to update them.

xgboost-job/xgboost-operator/overlays/application/kustomization.yaml
```diff
@@ -0,0 +1,14 @@
...
+resources:
+- application.yaml
+commonLabels:
+  app.kubernetes.io/name: xgboost-operator
+  app.kubernetes.io/instance: xgboost-operator
+  app.kubernetes.io/managed-by: kfctl
+  app.kubernetes.io/component: xgboostjob
+  app.kubernetes.io/part-of: kubeflow
+  app.kubernetes.io/version: v0.6
+
```
jlewi, line 14: Don't include version.
See: https://github.com/kubeflow/kfctl/issues/304#issuecomment-616602720

jlewi:
@terrytangyuan I don't understand what the problem is. Which version of kustomize are you using?

I'm using 

```
Version: {Version:kustomize/v3.2.1 GitCommit:d89b448c745937f0cf1936162f26a5aac688f840 BuildDate:2019-09-27T00:10:52Z GoOs:linux GoArch:amd64}
```

And I can run `make generate` off of master. So if `make generate` is failing for you then that suggests a problem with the version of kustomize you have or your pR.

xgboost-job/xgboost-operator/overlays/application/kustomization.yaml
terrytangyuan: Fixed.

xgboost-job/xgboost-operator/overlays/application/kustomization.yaml
terrytangyuan: Currently other operators all follow this same pattern. We can fix them altogether in a follow-up PR.

terrytangyuan:
@jlewi The problem is with kustomize `v3.4`. I downgraded it to `v3.2.1` and it works now. Regarding the dependency on base, currently other operators all follow this same pattern. We can perhaps fix them altogether in a follow-up PR so this PR can focus on bringing XGBoostJob to kubeflow/manifests.

Could you take another look?

xgboost-job/xgboost-operator/base/params.yaml
```diff
@@ -0,0 +1,3 @@
+varReference:
+- path: spec/template/spec/containers/image
```
jlewi, line 2: Why do you need this var?

xgboost-job/xgboost-operator/base/kustomization.yaml
```diff
@@ -0,0 +1,20 @@
...
+- deployment.yaml
+- service-account.yaml
+- service.yaml
+namespace: kubeflow
+nameprefix: xgboost-operator-
+configMapGenerator:
+- name: parameters
+  env: params.env
+generatorOptions:
+  disableNameSuffixHash: true
```
jlewi, line 16: Why are you disabling the suffix hash? Its much better to have a configmap contain a unique content hash so that any deployments using it will get updated when the configmap is used.

xgboost-job/xgboost-operator/base/kustomization.yaml
```diff
@@ -0,0 +1,20 @@
...
+- cluster-role.yaml
+- cluster-role-binding.yaml
+- crd.yaml
+- deployment.yaml
+- service-account.yaml
+- service.yaml
+namespace: kubeflow
+nameprefix: xgboost-operator-
+configMapGenerator:
+- name: parameters
```
jlewi, line 13: I think you want to use a unique enough name so that it won't conflict with configmaps from other applications.
https://github.com/kubeflow/manifests/blob/5de67258ca92a59f29d025d69a566310fce208f5/jupyter/jupyter-web-app/base_v3/kustomization.yaml#L46

We've been using the pattern

${appname}-config

terrytangyuan:
@jlewi Thank you for the detailed review! I've addressed your comments.

jlewi:
@terrytangyuan It looks like your application package is still depending on the base package.

I would suggest organizing it as follows

```
components/
    system - (rename base to system or something else more descriptive e.g. controller)
    application

installs
    full (not sure what a good name would be) - This would be a kustomize package that combines the system and application packages
```

The idea here is that installs contains fully deployable packages which are constructed by combining components.

You can take that as a suggestion though so if you prefer the current layout you can stick with it.

terrytangyuan:
@jlewi Thanks for the clarification. I'd like to keep the current layout for now as currently there are other operators that follow this pattern. We can perhaps address that altogether in the future so this PR can focus on bringing XGBoost operator available here.

PR #1119, opened 2020-04-20
fix roles binding namespace reference

thesuperzapper:
this fixes an issue introduced in #962, which causes the role bindings to be in the wrong namespace if the user is not using `kubeflow`

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

### DIFF
--- a/common/centraldashboard/base/clusterrole-binding.yaml
+++ b/common/centraldashboard/base/clusterrole-binding.yaml
@@ -11,4 +11,4 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: centraldashboard
-  namespace: kubeflow
+  namespace: $(namespace)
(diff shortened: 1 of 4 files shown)


thesuperzapper:
I actually think this issue was around in v1.0.2 as-well, and causes lots of parts of the central dashboard to not work. (CPU utilisation, cluster-admin binding, etc) 

So we might want to cut a release after this is merged.

thesuperzapper:
@krishnadurai, can you please review this?

thesuperzapper:
@krishnadurai we should still try and get this merged, I have rebased to try and clear CI issues, but I cant see the logs so I cant debug it.

thesuperzapper:
This was already addressed in #962, so is redundant.

PR #1122, opened 2020-04-21
* Change the version on the GKE cluster because the version specified

jlewi:
* Change the version on the GKE cluster because the version specified

* In a subsequent PR I will change it to a release channel because that's
    what ASM requires

* cert-manager and application packages need to set the namespace

  * This is because when using stacks we use the overlays which currently
    don't set namespace

  * As a result, application resource can get created in a different namespace
    which causes the deployments to be GC'd.

* Define configs for bootstrapping a management cluster

  * The management cluster will run KCC and be used to create Kubeflow deployments
  * Check in CNRM installs for the system components

* Add a Make rule to reset the kpt setters; this is useful for cleaning up the setters before checking them back in.

* Update istio-operator to latest ASM spec.
    
* Remove ambassador it is no longer used
* Fix annotation on cert-manager; it shouldn't be null


jlewi:
Test failed with

```
INFO     root:util.py:72 time="2020-04-21T01:40:11Z" level=error msg="Permanently failed applying application kfserving-gateway; error:  (kubeflow.error): Code 500 with message: Apply.Run  Error error when creating \"/tmp/kout728885934\": Service \"kfs
erving-ingressgateway\" is invalid: spec.ports[3].nodePort: Invalid value: 32400: provided port is already allocated" filename="kustomize/kustomize.go:249"
INFO     root:util.py:72 Error: failed to apply:  (kubeflow.error): Code 500 with message: kfApp Apply failed for kustomize:  (kubeflow.error): Code 500 with message: Apply.Run  Error error when creating "/tmp/kout728885934": Service "kfserving-ingress
gateway" is invalid: spec.ports[3].nodePort: Invalid value: 32400: provided port is already allocated
```

jlewi:
cert-manager tests failed

Issue #1092, opened 2020-04-10
v3 versions of katib manifests


jlewi:
We need v3 versions of the katib manifests

See #1062 for more info on what this means.

@johnugeorge @gaocegege @hougangliu is there someone from the katib effort who would be willing to pick this up?

PR #1124, opened 2020-04-21
Katib manifests v3 version

andreyvelich:
Fixes: https://github.com/kubeflow/manifests/issues/1092.

For Katib manifest v3 version:
1. I Created `v3` folder for Katib crds and Katib controller.

2. For Katib crds kustomize manifest includes all base resources and application overlay.

3. Katib crds kustomize has `commonLabels` for each crd component.

3. For Katib controller kustomize manifests includes all base resources, application and istio overlay.

4. Katib controller kustomize has `commonLabels` for each controller component.

5. Katib controller kustomize has `images` to specify images for Katib controller components.

6. I renamed `clusterDomain` to `katib-ui-clusterDomain` and `namespace` to `katib-ui-namespace` to make these variables unique. @jlewi What do you think about this change?

7. Katib controller kustomize has `configurations` to identify `katib-ui-clusterDomain` variable.

8. I didn't add [these params](https://github.com/kubeflow/manifests/blob/master/katib/katib-controller/base/params.yaml) to Katib controller kustomize, as we did [here](https://github.com/kubeflow/manifests/blob/master/katib/katib-controller/base/kustomization.yaml#L52). I am not sure that it is needed.

9. I didn't add `generatorOptions: disableNameSuffixHash: true` to v3 kustomize manifests. Is it necessary?

10. I added Katib v3 manifest to GCP example stacks. @jlewi Do we need to make it in this PR?

I didn't add [ibm-storage-config](https://github.com/kubeflow/manifests/tree/master/katib/katib-controller/overlays/ibm-storage-config) to v3 manifest, since it is related only to IBM kfdef. We can create this patch later.

/assign @jlewi @johnugeorge 
/cc @gaocegege @richardsliu 

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`


jlewi:
Thanks.

> I added Katib v3 manifest to GCP example stacks. @jlewi Do we need to make it in this PR?

Probably easier to do it a follow on PR. Smaller PRs are easier to deal with and this way if the manifest breaks GCP stacks we can just revert that PR and not the full PR.

> I renamed clusterDomain to katib-ui-clusterDomain and namespace to katib-ui-namespace to make these variables unique. @jlewi What do you think about this change?

For clusterDomain I'm not sure we want to make the name application specific. The point of using the same vars across the applications is to allow the values to be defined once and consistently set across all applications.

So I think the pattern you want is to have 2 versions of the package

1. For backwards compatibility you should have a kustomization that defines the existing vars in that kustomize
   * This can't be used in the stacks since vars can only be defined once
1. You want a v3 version which doesn't define the vars in the kustomization file; instead these will be defined at a kustomization at the stack level

These packages should share resources either through a common base package or by referencing the same YAML files.

You can take a look at the jupyter package
https://github.com/kubeflow/manifests/blob/master/jupyter/notebook-controller/base_v3/kustomization.yaml

* Here that the vars aren't defined in this kustomization
* The vars are defined in base/kustomization.yaml for backwards compatibility
* Note that base_v3 is pulling in resources from the base package

> katib-ui-namespace 

How is this being used? Is this internal to the package; i.e. its not being used to allow the user to set something but just a way to force kusotmize to substitute the namespace into a field where otherwise it currently wouldn't get substituted?

If it is then this is a good pattern.

jlewi:
How does this relate to @discordianfish's PR #986 ?

As discussed in #986 can we split katib into separate kustomize packages for the controller and the DB and then use composition to combine them? So that we can support incluster and external DBs?

andreyvelich:
@jlewi Thank you for your review.

- I don't update stacks in this PR.
- I renamed `katib-ui-clusterDomain` to `clusterDomain` as you said.
- I removed version from common label. Should we also modify application: https://github.com/kubeflow/manifests/blob/master/katib/katib-crds/overlays/application/application.yaml#L57 in current kustomize implementation?

andreyvelich:
> _[katib/katib-controller/base/kustomization.yaml, line 44 at r1](https://reviewable.io/reviews/kubeflow/manifests/1124#-M5YZcJI-qZMXl2vpalM:-M5YZcJI-qZMXl2vpalN:bynz6dj) ([raw file](https://github.com/kubeflow/manifests/blob/795a23605c2187dc872c3e775b625a2d22c21f51/katib/katib-controller/base/kustomization.yaml#L44)):_
> 
> > ```yaml
> >   fieldref:
> >     fieldpath: data.katib-ui-clusterDomain
> > - name: katib-ui-namespace
> > ```
> 
> This is largely internal to the project; you are just using
> 

Yes, and it always must be `metadata.namespace` from `katib-ui` service. The same as in [jupyter-web-app]( https://github.com/kubeflow/manifests/blob/master/jupyter/jupyter-web-app/overlays/istio/virtual-service.yaml#L22).

andreyvelich:
> How does this relate to @discordianfish's PR #986 ?
> 
> As discussed in #986 can we split katib into separate kustomize packages for the controller and the DB and then use composition to combine them? So that we can support incluster and external DBs?

We can merge https://github.com/kubeflow/manifests/pull/986 PR first.
After this, modify v3 Katib package. What do you think @johnugeorge @gaocegege ?

jlewi:
If we merge #986 first is that creating more work in the long run? Per the discussion in the review thread
https://github.com/kubeflow/manifests/pull/986#discussion_r408173149

The suggestion was to have separate kustomize packages for

* controller
* db-manager
* my-sql

And then use a top level package to compose this into a single package. This way the controller and db-manager packages could be reused in the case of a different backend. #985 is making db-manager an overlay. 

My suggestion would be to do something like the following. In this PR define 

* katib-system
* katib-db-manager
* katib-mysql

Each have these kustomize packages should pull in resources from base (like your v3) package is doing now.

You can then change the v3 package to compose these three kustomize packages. (Maybe rename it to something like standalone?). This package should also define a patch for the db-manager as in 
https://github.com/kubeflow/manifests/blob/3476c3ca6d13966b61a1539748bdeab3187d083d/katib/katib-controller/overlays/db-mysql/katib-db-manager-deployment.yaml 
to set the secrets for mysql

This would largely cover @discordianfish's changes to support the external DB case.

The one other change is you need to change 
base//katib-db-manager-deployment.yaml to remove the DB environment variables for MYSQL; These should be applied via a batch which should be included in base/ (for backwards compatibility) but not katib-db-manager.

katib/katib-controller/v3/kustomization.yaml
```diff
@@ -0,0 +1,42 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: kubeflow
+resources:
+  - ../base/katib-configmap.yaml
```
jlewi, line 5: So per comment: https://github.com/kubeflow/manifests/pull/1124#issuecomment-618634808

I think this should be something like

resources:
- ../katib-system
- ../katib-db-manager
- ../katib-mysql

The resources listed here would then move into the respective packages.

patchesStrategicMerge
- <patch to add MySQL env variables to db manger>

jlewi:
> Should we also modify application: https://github.com/kubeflow/manifests/blob/master/katib/katib-crds/overlays/application/application.yaml#L57 in current kustomize implementation?

Yes I think we should go ahead and remove the mutable bits from the commonLabels from the application resource

katib/katib-controller/v3/kustomization.yaml
andreyvelich: Yes I agree, we can use patch here.
What do you think @discordianfish, can you make these changes in https://github.com/kubeflow/manifests/pull/986 PR?

/cc @johnugeorge @gaocegege

andreyvelich:
> > Should we also modify application: https://github.com/kubeflow/manifests/blob/master/katib/katib-crds/overlays/application/application.yaml#L57 in current kustomize implementation?
> 
> Yes I think we should go ahead and remove the mutable bits from the commonLabels from the application resource

I removed versions from Katib controller Application and Katib Crd Application

andreyvelich:
@jlewi What do you think about `katib-ui-namespace` variable, should we keep it as a `namespace` to be consistent with `jupyter-web-app` or identify it as the Katib variable ?

katib/katib-controller/v3/kustomization.yaml
jlewi: I think in #986 @discordianfish said he didn't have time to make any additional changes. So maybe @andreyvelich you could just make the changes in this PR? Or a follow on PR?

jlewi:
/lgtm
/approve
/hold

@andreyvelich On my end I'm happy with this PR as is. I think we still need to refactor the manifests to support external DB but I'm ok with doing that in a follow on PR.

Ideally someone focused on katib would LGTM this PR as well.

andreyvelich:
> /lgtm
> /approve
> /hold
> 
> @andreyvelich On my end I'm happy with this PR as is. I think we still need to refactor the manifests to support external DB but I'm ok with doing that in a follow on PR.
> 
> Ideally someone focused on katib would LGTM this PR as well.

Sure, I will update it in the following PR. Thank you for the review.
/assign @johnugeorge

katib/katib-controller/v3/kustomization.yaml
discordianfish: I was waiting for you to settle on one way to do it. But now it's probably better to merge this here first. I can look into a new PR to change this.

johnugeorge:
Based on the comment https://github.com/kubeflow/manifests/pull/1124#discussion_r415614797 , @discordianfish will do it in a new PR. Thanks @discordianfish 

Merging this

/hold cancel

Issue #1131, opened 2020-04-22
commonLabels need to be immutable for upgrades - remove version from commonLabels


jlewi:
The commonLabels in our kustomize package currently include the application version.
This breaks upgradability because commonLabels are substituted in selectors which are immutable.

This has been raised in a number of issues

As noted in kubeflow/kubeflow#4873; kustomize commonLabels should only be used for immutable labels
https://kubectl.docs.kubernetes.io/pages/app_management/labels_and_annotations.html

kubeflow/kfctl#304 describes current thinking for how to do upgrades which should make version labels unnecessary.

We should think about unittests or other forms of validation to ensure all manifests adhere to this.

Issue #1121, opened 2020-04-08
certmanager install has race condition - try to create KF certmanager resources before cert manager is available


SamChb72:
/kind bug

Hello,
little problem with cert-manager during kubeflow install:

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]
I encounter the problem
kfctl apply -f ${CONFIG} ends with 
`
failed to apply:  (kubeflow.error): Code 500 with message: kfApp Apply failed for kustomize:  (kubeflow.error): Code 500 with message: Apply.Run  Error error when creating "/tmp/kout285717195": Internal error occurred: failed calling webhook "webhook.cert-manager.io": the server is currently unable to handle the request
`

install exists after a loop of error messages:
`
validatingwebhookconfiguration.admissionregistration.k8s.io/cert-manager-webhook configured
WARN[0106] Encountered error applying application cert-manager:  (kubeflow.error): Code 500 with message: Apply.Run  Error error when creating "/tmp/kout520652292": Internal error occurred: failed calling webhook "webhook.cert-manager.io": the server is currently unable to handle the request  filename="kustomize/kustomize.go:202"
WARN[0106] Will retry in 29 seconds.                     filename="kustomize/kustomize.go:203"
namespace/cert-manager unchanged
`

**What did you expect to happen:**

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]
I follow the install page:
https://www.kubeflow.org/docs/started/k8s/kfctl-k8s-istio/

**Environment:**

- Kubeflow version: (version number can be found at the bottom left corner of the Kubeflow dashboard):
kfctl_k8s_istio.v1.0.1.yaml
- kfctl version: (use `kfctl version`):
kfctl v1.0.1-0-gf3edb9b
- Kubernetes platform: (e.g. `minikube`)
one fresh node install with kubeadm
- Kubernetes version: (use `kubectl version`):
Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.0", GitCommit:"e8462b5b5dc2584fdcd18e6bcfe9f1e4d970a529", GitTreeState:"clean", BuildDate:"2019-06-19T16:40:16Z", GoVersion:"go1.12.5", Compiler:"gc", Platform:"linux/amd64"}
[...]

PR #1134, opened 2020-04-24
Fix certmanager race condition and version numbers.

jlewi:
* To fix the race condition with certmanager (#1121) we move the KF
  issuer into a separate package from the package deploying kubeflow.
  This way we can wait for cert-manager to start before deploying resources.

* Labels need to be immutable otherwise upgrades won't work (see #1131).
  So remove version number from common labels and application selector
  so that apply will work to update resources.


jlewi:
@krishnadurai @johnugeorge @yanniszark could one of you review this please?

PR #1135, opened 2020-04-24
Fix CNRM cluster package.

jlewi:
* namespace should be set in kustomization.yaml; not in the
  actual resource. If we set it in the resource then if we have kustomize
  subpackages that try to patch the resource kustomize won't be able
  to find the original resource.


jlewi:
@Bobgy @rmgogogo @kunmingg Does one of you want to volunteer to review this?

rmgogogo:
I reviewed yesterday but TBH didn't have enough knowledge to catch the details yet.
Any docs I can refer?

LGTM first to unblock your side work.

/lgtm

jlewi:
Thanks @Bobgy  

/approve

PR #1140, opened 2020-04-29
commonLabels need to be immutable to support upgrading.

jlewi:
* Fix kubeflow/manifests#1131

* kustomize commonLabels get subsituted into selector fields. Selector fields
  are immutable. So if commonLabels change (e.g. between versions) then
  we can't reapply/update the existing resources which breaks upgrades
 (kubeflow/kfctl#304)

* For the most part the problematic commonLabels were on our Application
  resources. The following labels were being set

  "app.kubernetes.io/version"
  "app.kubernetes.io/instance"
  "app.kubernetes.io/managed-by"
  "app.kubernetes.io/part-of"

* Version was definetely changing between versions. instance was also changing
  between versions to include the version number.

* managed-by and part-of could also change (e.g. we may not be using kfctl)
* We could still set these labels if we wanted to; we just shouldn't set
  them as commonLabels and/or include them in the selector as the will
  inhibit upgrades with kubectl apply.

* I created a test validate_resources_test.go to ensure none of these
  labels are included in commonLabels

* I created a simple go binary tools/fix_common_labels.go to update
  all the resources.

* generat_tests.py - Delete the code to remove unmatched tests.
  * We no longer generate tests that way and the delete code was going
    to delete valid tests like our new validation test

* Get rid of the clean rule in the Makefile for the same reason.


kkasravi:
Should fix_common_labels.go be a transformer or just a tool that does the insertion?
Just wondering if we should begin to do transforms / generation for other scenarios as well since it fits a transform use case. Also could we use the kustomize LabelTransformer builtin?
https://github.com/kubernetes-sigs/kustomize/blob/master/docs/plugins/builtins.md#labeltransformer

jlewi:
fix_common_labels.go is a one off tool so I don't really think it matters how its implemented. In principle we should never have to use it again because we should never check in kustomization files that are setting these labels in commonLabels.

jlewi:
@yanniszark @johnugeorge could you please take a look?

Issue #1007, opened 2020-03-13
Can't compose generated Kustomizations - due to use of vars


rehevkor5:
Due to the use of "vars" in the generated kustomizations, it's not possible to compose the generated kustomizations together in order to use `kubectl -k` to deploy the whole thing plus additional variant info like env-specific stuff. If you do, errors such as "var clusterDomain already encountered" and "error: found 2 resId matches for var {project ~G_v1_ConfigMap {data.project}} (unable to disambiguate)" occur.

Can all use of vars be removed to prevent this issue?

I am new to Kustomize, so I may be misunderstanding how this stuff should work. But, the var problem is making it difficult to figure out how to maintain kustomizations in source control with additional layers for environment-specific stuff. `kfctl` by itself doesn't provide any mechanism for user-level additions/changes, without manually modifying the output every time it's regenerated (not desirable). It seems like for now I may have to maintain a separate kustomization.yaml for each generated kustomization, so that I can apply a shared base and/or any env-specific config. I guess this means that if I have 36 generated kustomizations, that I need to have `36 * (env count)` individual kustomization files to provide env-specific concerns? If so, this is not a very practical approach.

jlewi:
Related issue: kubeflow/manifests/issues/774

jlewi:
This is one of the disadvantages of vars as noted in: kubernetes-sigs/kustomize#2052 

Our biggest use case for vars is for substituting namespace and clusterDomain into virtual services ([example](https://github.com/kubeflow/manifests/blob/4d8b1cb7e6340a69303ce59878fc29f6b3a71a72/jupyter/jupyter-web-app/overlays/istio/virtual-service.yaml#L22)). As suggested in kubernetes-sigs/kustomize#205 using creating [plugins](https://github.com/kubernetes-sigs/kustomize/tree/master/docs/plugins) for virtual services might be a better long term option. 

We could define a VirtualService transformer similar to the built in [image transformer](https://github.com/kubernetes-sigs/kustomize/blob/master/plugin/builtin/imagetagtransformer/ImageTagTransformer.go)


A custom transformer would allow users to add a virtualServices section to their kustomization.yaml files to change hosts; e.g.

```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
virtualServices: 
 - host: jupyter-web-app-service.kubeflow.svc.cluster.local   
    newHost: jupyter-web-app-service.customnamespace.svc.some-other-domain
--

[...]

PR #1144, opened 2020-04-30
Create an updated style/best practices guide for kustomize packages.

jlewi:
* See http://bit.ly/kf_kustomize_v3

* Its been over a year since we started migrating to kustomize. In that time
kustomize has evolved and we have learned a lot about how to use
kustomize effectively.

* Some of the initial usage patterns have turned out to be problematic (e.g. vars) and there are now better options.

* Update the style guide to better document best practices.

Related to #1062

### DIFF
--- a/docs/KustomizeBestPractices.md
+++ b/docs/KustomizeBestPractices.md
@@ -1,28 +1,26 @@
-<!-- START doctoc generated TOC please keep comment here to allow auto update -->
-<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
-**Table of Contents**  *generated with [DocToc](https://github.com/thlorenz/doctoc)*
-
-- [Kustomize Best Practices](#kustomize-best-practices)
-  - [Package layout.](#package-layout)
-  - [Command Line substitution](#command-line-substitution)
-  - [Eschew vars](#eschew-vars)
-    - [Internal subsitution of fields Kustomize isn't aware of.](#internal-subsitution-of-fields-kustomize-isnt-aware-of)
-    - [Global substitution](#global-substitution)
-  - [Have separate packages for CR's and instances of the custom resource](#have-separate-packages-for-crs-and-instances-of-the-custom-resource)
-  - [CommonLabels should be immutable](#commonlabels-should-be-immutable)
-    - [1b. Removing common attributes across resources](#1b-removing-common-attributes-across-resources)
-    - [1c. Parameters](#1c-parameters)
-    - [1d. Overlays](#1d-overlays)
-
-<!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
 # Kustomize Best Practices
 
   This doc provides best practices for writing Kubeflow kustomize packages.
 
-## Package layout.
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Table of Contents**
 
-If your application consists of loosely coupled component e.g. backend, front-end, database consider defining these as separate kustomize packages
+- [Layout package to support composability](#layout-package-to-support-composability)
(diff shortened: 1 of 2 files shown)


jlewi:
@Bobgy @rmgogogo @yanniszark @johnugeorge @krishnadurai @yanniszark 
Could you take a look at this please?

Bobgy:
/lgtm
since I reviewed kustomize v3 doc

/cc @rmgogogo you may be interested in the part about var replacement

docs/KustomizeBestPractices.md
```diff
@@ -1,22 +1,133 @@
...
+
+## CommonLabels should be immutable
+
+As noted [here](https://kubectl.docs.kubernetes.io/pages/reference/kustomize.html#commonlabels) commonLabels get applied to
+selectors which are immutable. Therefore, commonLabels should be immutable across versions of a package to avoid causing
+problems during upgrades.
+
+For more info see [kubeflow/manifests#1131](https://github.com/kubeflow/manifests/issues/1131)
+
+###Resource file naming
```
terrytangyuan, line 130: Missing a space after `###`

terrytangyuan:
@jlewi Thanks for putting this together! This is very useful. One general comment: would it be a good idea to provide an example that follows all these best practices so that we can point to specific lines from this example when illustrating each best practice.

docs/KustomizeBestPractices.md
```diff
@@ -1,22 +1,133 @@
...
+This use of vars is acceptable because the var is internal to the kustomize package and can be given a unique enough name to prevent
+conflicts when the package is composed with other applications.
+
+### Global substitution
+
+One of the most problematic use cases for vars in Kubeflow today is substituting a user supplied value into multiple applications.
+
+Currently we only have one use case which is substituting in cluster domain into virtual services ([ref](https://docs.google.com/document/d/1jBayuR5YvhuGcIVAgB1F_q4NrlzUryZPyI9lCdkFTcw/edit#heading=h.vyq4iltpirga)).
+
+We would ultimately like to get rid of the use of vars in these cases but have not settled on precise solutions. Some possible options are
```
Klaven, line 102: IMO stating that you should not use global substitution except where there is no other option is enough in the best practices, no need to delve into possible solutions. Maybe just make a overarching issue where discussion can be had and just link there "if you want more info"?

Its fine as is. just my preference.

yanniszark:
@jlewi thanks a lot for putting this together!
Some thoughts are:

* In the design doc, the `--load_restrictor none` option is mentioned in order to reuse patches. Should we mention that we allow its use and encourage reusing patches in this doc?
* Should we mention the global Kubeflow configmap for common options (e.g., `kubeflow-userid`)?

docs/KustomizeBestPractices.md
jlewi: I think I'd like to nudge people towards using kpt setters. I guess we could just as easily link to an issue with recommended patterns and describe uses there.

jlewi: Thanks @Klaven this is a good suggestion. I thought about following your suggestion and just linking to the bug kubeflow/manifests#1007. The issue is getting quite long though so then I thought it would be valuable to kind of summarize it here.  I also want to nudge people to start trying/using `kpt setters`.

jlewi:
@yanniszark 

> In the design doc, the --load_restrictor none option is mentioned in order to reuse patches. Should we mention that we allow its use and encourage reusing patches in this doc?

Done.

> Should we mention the global Kubeflow configmap for common options (e.g., kubeflow-userid)?

I'd forgotten about that. The global configmap seems like an inferior solution to kpt setters. So I think I'd rather nudge people to kpt setters.

The problem with the global configmap is that we then need to have variations of the application package which use the global config map as opposed to the application configmap. With kpt setters we should be able to kustomize package complete on its own while also allowing global parameters to be set once.

jlewi:
@yanniszark @krishnadurai could one of you LGTM please?

jlewi:
Thanks

/approve

Issue #984, opened 2020-03-04
Bump pipeline version to 0.2.5 in v1.0-branch


Jeffwan:
Pipeline version in Kubeflow v1.0 is 0.2.0 which has been released for 2 months. 
I suggest to bump the version to 0.2.5 in v1.0-branch as it contains some performance critical improvements.

Examples:
https://github.com/kubeflow/pipelines/issues/3165

/cc @richardsliu @Bobgy @Ark-kun

Bobgy:
Yep, I think we should do this. I'd like to know others' opinions too.

Related issues include https://github.com/kubeflow/kubeflow/issues/4829

Jeffwan:
Hi @Bobgy, Not sure any compatibility concerns to use 0.2.5 in v1.0-branch? I think the plan should be use 0.2.5 is master, run some manual tests and if cherry-pick back to v1.0-branch?

Bobgy:
I don't think there are any compatibility issues, KFP has been backward compatible from 0.2.0 to 0.2.5

Jeffwan:
@Bobgy Cool, thanks for the confirmation. Kindly ping @Ark-kun @rmgogogo for their feedbacks

Bobgy:
@Ark-kun @rmgogogo pinging again

elikatsis:
This (important) issue is also resolved in KFP>0.2.0 (included in release 0.2.1): https://github.com/kubeflow/pipelines/issues/2764

Bobgy:
I'd like to send a PR to update KFP to 0.3.0 instead, because we'd make sure kfp is not broken due to https://github.com/kubeflow/kfctl/pull/131.

KFP 0.3.0 included the fix: https://github.com/kubeflow/pipelines/pull/3305

Bobgy:
I didn't see this issue is about v1.0-branch. Actually I only intend to upgrade to 0.3.0 in master branch, because it was not as heavily tested as 0.2.5. I'd still recommend 0.2.5 for v1.0-branch.

Jeffwan:
We need to discuss whether we like to cut v1.1-branch. If it's coming soon, it's fine to leave v1.0-branch with existing version

Bobgy:
@Jeffwan Jeremey mentioned v1.1 is targeted for June, so looks like updating v1.0-branch will be the best option now.

apryiomka:
Do we know when this might be merged to Kubeflow v1.0.2 manifest?

PR #1148, opened 2020-05-01
Bump KFP to 0.5.0

eedorenko:
Resolves https://github.com/kubeflow/pipelines/issues/3441
                https://github.com/kubeflow/pipelines/issues/2237
                https://github.com/kubeflow/manifests/issues/984

Bump Kubeflow pipeline version to 0.5.0

### DIFF
--- a/kfdef/kfctl_k8s_istio.yaml
+++ b/kfdef/kfctl_k8s_istio.yaml
@@ -353,5 +353,4 @@ spec:
   repos:
   - name: manifests
     uri: https://github.com/kubeflow/manifests/archive/master.tar.gz
-
   version: master
--- a/kfdef/kfctl_k8s_istio.yaml
+++ b/kfdef/kfctl_k8s_istio.yaml
@@ -352,5 +352,6 @@ spec:
     name: seldon-core-operator
   repos:
   - name: manifests
-    uri: https://github.com/kubeflow/manifests/archive/master.tar.gz
+    uri: /Users/efedorenko/work/kubeflowmanifests
+
   version: master
--- a/kfdef/kfctl_k8s_istio.yaml
+++ b/kfdef/kfctl_k8s_istio.yaml
@@ -352,6 +352,6 @@ spec:
     name: seldon-core-operator
   repos:
   - name: manifests
-    uri: /Users/efedorenko/work/kubeflowmanifests
+    uri: https://github.com/kubeflow/manifests/archive/master.tar.gz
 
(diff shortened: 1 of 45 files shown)


Bobgy:
@eedorenko in addition to bumping the version, you also need to mirror non-refactoring changes in https://github.com/kubeflow/pipelines/commits/master/manifests/kustomize to here.

eedorenko:
Could you please clarify what non-refactoring changes you mean?

Bobgy:
Hi @eedorenko , FYI KFP 1.0.0-rc.2 has been integrated with kubeflow in https://github.com/kubeflow/manifests/pull/1248.

Jeffwan:
@Bobgy This is different source from the PR for 1.0.0. Not all boostrap tool are using stacks I think. This is still worth to merge it

Bobgy:
@Jeffwan problem is, this PR didn't include all upstream manifest changes to support a proper KFP manifest update, some features might be broken.

If you tested this branches work for you, I don't have any objections.

but KFP team will only keep maintain the stacks because of the manifest syncing cost. Would you consider making a wrapper `pipeline/installs/base/kustomization.yaml` file which includes `pipeline/installs/generic` so that the stacks manifest could be used with kfctl too (not sure on details, but it sounds doable)?

eedorenko:
@Bobgy what upstream manifest changes do you think should be included?
This branch has been tested. We've been working with it since May 1st.

Bobgy:
I briefly went through all changes between 0.2.5 and 0.5.0.

I think you need https://github.com/kubeflow/pipelines/commit/f882f361cc705580261128e33523fe80d781b30f#diff-17145b7d84b2980d3edd18ba00b95d0e.
Also, KFP caching servers were added in 0.5.0, but they are not included in manifests here.

This is the best I can help, you will need to troubleshoot and maintain this deployment by yourselves.

Bobgy:
Also, I'm not sure we should merge this to v1.0-branch. There won't be kubeflow 1.0.3 release, because everyone is planning on 1.1.

Jeffwan:
@Bobgy  Agree. We won't have v1.0.3 release,

@eedorenko v1.1.0 will be officially released next week. Are you ok to migrate to v1.1.0? If so, we can close this PR

eedorenko:
@Jeffwan Yes, I am Ok with migrating to  v1.1.0.
So, if v1.0.3 doesn't happen I am closing this PR.

PR #1153, opened 2020-05-06
Add kubeflow namespace resource to enable istio injection

Bobgy:
Resolves https://github.com/kubeflow/kubeflow/issues/3866

Add kubeflow namespace resource with istio injection label to enable automatic sidecar injection.

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`


Bobgy:
@jlewi Can you please review this?

Bobgy:
Test failures don't seem related.
/retest

kubeflow-namespace/base/namespace.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
```
jlewi, line 1: I think you should include cert-manager as well like
https://github.com/kubeflow/gcp-blueprints/blob/master/kubeflow/instance/kustomize/namespaces/namespaces.yaml

I might suggest the following hierarchy

namespaces/all/kustomization.yaml

This way if we need to create different combinations of namespaces we can do so.

jlewi: Actually, you might need to keep "base" for legacy reasons as kfctl depends on it today.

Bobgy: sure, will change like that

Bobgy: @jlewi PTAL

Bobgy:
Test failure blocked by https://github.com/kubeflow/manifests/issues/1157

jlewi:
@Bobgy Is your PR rebased on the latest master?
Did you try regenerating the tests by running `make generate`?

Bobgy:
@jlewi Can you lgtm again? tests are passing now

Issue #1157, opened 2020-05-07
Presubmit tests fail in master


Bobgy:
Steps to reproduce:
1. cd tests
2. make test

I tried `cd tests && make generate`, although some yamls are regenerated `make test` still fails.

/assign @jlewi 
Who should be the best person to fix it?

Error log
```
=== RUN   TestCommonLabelsImmutable
    TestCommonLabelsImmutable: validate_resources_test.go:57: Skipping directory ../.git
    TestCommonLabelsImmutable: validate_resources_test.go:57: Skipping directory ../.github
    TestCommonLabelsImmutable: validate_resources_test.go:83: ../katib/components/katib-controller/kustomization.yaml has forbidden commonLabel app.kubernetes.io/managed-by
    TestCommonLabelsImmutable: validate_resources_test.go:83: ../katib/components/katib-controller/kustomization.yaml has forbidden commonLabel app.kubernetes.io/instance
    TestCommonLabelsImmutable: validate_resources_test.go:83: ../katib/components/katib-controller/kustomization.yaml has forbidden commonLabel app.kubernetes.io/part-of
    TestCommonLabelsImmutable: validate_resources_test.go:83: ../katib/installs/katib-external-db/kustomization.yaml has forbidden commonLabel app.kubernetes.io/managed-by
    TestCommonLabelsImmutable: validate_resources_test.go:83: ../katib/installs/katib-external-db/kustomization.yaml has forbidden commonLabel app.kubernetes.io/instance
    TestCommonLabelsImmutable: validate_resources_test.go:83: ../katib/installs/katib-external-db/kustomization.yaml has forbidden commonLabel app.kubernetes.io/part-of
    TestCommonLabelsImmutable: validate_resources_test.go:83: ../katib/installs/katib-standalone/kustomization.yaml has forbidden commonLabel app.kubernetes.io/managed-by
    TestCommonLabelsImmutable: validate_resources_test.go:83: ../katib/installs/katib-standalone/kustomization.yaml has forbidden commonLabel app.kubernetes.io/instance
    TestCommonLabelsImmutable: validate_resources_test.go:83: ../katib/installs/katib-standalone/kustomization.yaml has forbidden commonLabel app.kubernetes.io/part-of
    TestCommonLabelsImmutable: validate_resources_test.go:57: Skipping directory ../tests
--- FAIL: TestCommonLabelsImmutable (0.11s)
FAIL
FAIL    github.com/kubeflow/manifests/tests     1.720s
[...]

PR #1158, opened 2020-05-08
Revert "katib: Add per component packages"

Bobgy:
Reverts kubeflow/manifests#1141
Fixes https://github.com/kubeflow/manifests/issues/1157

Presubmit and postsubmit tests all start to fail after #1141 
Try if reverting fixes these.


Bobgy:
unit tests have passed

Issue #1174, opened 2020-05-15
Lots of resources have status fields which breaks ACM


jlewi:
A lot of our K8s resources (e.g. Jupyter CRD) have status set
https://github.com/kubeflow/manifests/blob/d82342c88cb943635d6842db3153f9909181a067/jupyter/notebook-controller/base/crd.yaml#L65

This breaks ACM validation and likely other validation as well. I suspect having status fields breaks apply. So we need to fix this as part of upgrades (kubeflow/kfctl#304)

As part of fixing this we should add a validation test.

Related to: kubeflow/gcp-blueprints#4 

/assign @jlewi

jlewi:
It looks like the existence of status fields in CRDs is rampant. 

Argo doesn't have it
https://github.com/argoproj/argo/blob/master/manifests/base/crds/workflow-crd.yaml

Not seeing it in cert-manager either
https://github.com/jetstack/cert-manager/blob/master/deploy/crds/crd-clusterissuers.yaml

KFServing has it:
https://github.com/kubeflow/kfserving/blob/e25758f6da020e3e8ad63f0880ab95de6af8395b/config/default/crds/base/serving.kubeflow.org_inferenceservices.yaml#L6662

Notebook has it
https://github.com/kubeflow/kubeflow/blob/c7150a18b3a567a7f1d59893a0ef2adc1245ae14/components/notebook-controller/config/crd/bases/kubeflow.org_notebooks.yaml#L4265

kfctl doesn't have it.
https://github.com/kubeflow/kfctl/blob/master/deploy/crds/kfdef.apps.kubeflow.org_kfdefs_crd.yaml

My conjecture is that this is a bug in controller-runtime that got fixed in later versions.
Looks like kfctl is using 0.4.0
https://github.com/kubeflow/kfctl/blob/aaba770b461a27fc59d4a7960df48a3da003eda2/go.mod#L61

Looks like notebook controller is using 0.2.0
https://github.com/kubeflow/kubeflow/blob/master/components/notebook-controller/go.mod

jlewi:
This is largely fixed by #1175 and #1179.

The remaining problem is in the istio manifests. I have a prototype of a kustomize function transformer to fix those.

PR #1175, opened 2020-05-16
CRDs should not have status field.

jlewi:
* kubeflow/manifests#1174 lots of CRDs have status fields which
  causes ACM to complain that the CRD is invalid.

* This PR cleans up those CRDs and adds an appropriate validation test.

* Fix typos in AWS application specs (kubeflow/testing#668)


Jeffwan:
/lgtm

Thanks for the fix

jlewi:
Not sure why tide isn't merging this.

/approve

PR #1192, opened 2020-05-21
Updated cluster-role.yaml to run Kubeflow Faring from Kubeflow Pipeline

raviranjan0309:
Which issue is resolved by this Pull Request:
Resolves #
When someone runs Kubeflow Faring from Kubeflow Pipeline, he/she will get below error:
IOError: [Errno 2] No such file or directory: ‘/etc/secrets/user-gcp-sa.json’

Reason:
When Kubeflow Fairing tries to launch the pod for model training like LightGBM distributed parallel training, it tries to check the presence of the secret by using list instead of get and when list fails due to lack of permission, it just does not attach any secrets to the pod. This causes the whole Kubeflow pipeline to fail.

Description of your changes:
Updated the pipeline-runner ClusterRole to allow the pipeline-runner to list the secrets. This helped me run Kubeflow fairing from the Kubeflow pipeline successfully.

Checklist:

Unit tests have been rebuilt:
cd manifests/tests
make generate-changed-only
make test


raviranjan0309:
@krishnadurai Triggered test case run. Please review for approval.

raviranjan0309:
@krishnadurai The test case failed again. I could not able to find the root cause. I am doing a fresh fork and then will make a change. I am closing both the PR and raise a new one on the latest master branch.

PR #1196, opened 2020-05-22
Add GCS scope to the cluster; create an OWNERs file.

jlewi:
* Add GCS scope to the VM service account otherwise we won't be
  able to pull private images.

* Create an OWNERs file inside the GCP directory.


jlewi:
/assign @Bobgy 
@Bobgy please take a look

PR #1203, opened 2020-05-26
Add arm64 support to istio node selectors

MrXinWang:
Current istio manifests do not have architecture support for arm64 in nodeSelectorTerms. This commit adds the related arm64 supports.

Signed-off-by: Henry Wang <<Henry.Wang@arm.com>>


jlewi:
/assign @krishnadurai 

@krishnadurai looks like the active OWNER of the ISTIO configs so he should probably review these changes

/unassign @jlewi
/unassign @Bobgy

MrXinWang:
Thanks @jlewi :)

PR #1209, opened 2020-05-28
Updated cluster-role.yaml to run Kubeflow Faring from Kubeflow Pipeline

raviranjan0309:
Resolves #
When someone runs Kubeflow Faring from Kubeflow Pipeline, he/she will get below error:
**IOError: [Errno 2] No such file or directory: ‘/etc/secrets/user-gcp-sa.json’**

**Reason:**
When Kubeflow Fairing tries to launch the pod for model training like LightGBM distributed parallel training, it tries to check the presence of the secret by using list instead of get and when list fails due to lack of permission, it just does not attach any secrets to the pod. This causes the whole Kubeflow pipeline to fail.

Updated the pipeline-runner ClusterRole to allow the pipeline-runner to list the secrets. This helped me run Kubeflow fairing from the Kubeflow pipeline successfully.

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

### DIFF
--- a/pipeline/pipelines-runner/base/cluster-role.yaml
+++ b/pipeline/pipelines-runner/base/cluster-role.yaml
@@ -9,6 +9,7 @@ rules:
   - secrets
   verbs:
   - get
+  - list
 - apiGroups:
   - ""
   resources:
--- a/tests/stacks/examples/alice_gcp/test_data/expected/apps_v1_deployment_jupyter-web-app-deployment.yaml
+++ b/tests/stacks/examples/alice_gcp/test_data/expected/apps_v1_deployment_jupyter-web-app-deployment.yaml
@@ -26,12 +26,12 @@ spec:
           valueFrom:
             configMapKeyRef:
               key: userid-header
-              name: kubeflow-config-4bkkg42k5m
+              name: kubeflow-config
         - name: USERID_PREFIX
           valueFrom:
             configMapKeyRef:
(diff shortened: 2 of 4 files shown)


Bobgy:
I think there's a bug that causes unintended generated tests for tests/stacks/examples/alice_gcp/test_data/expected/apps_v1_deployment_jupyter-web-app-deployment.yaml and the other one that removed config name suffix.

You may try to revert those files manually and only keep the one related to your change to pass unit tests.
/cc @jlewi

jlewi:
@Bobgy if you think there is an issue with the tests can you please file an issue?

Bobgy:
> @Bobgy if you think there is an issue with the tests can you please file an issue?

@jlewi Sure, created https://github.com/kubeflow/manifests/issues/1217

raviranjan0309:
@Bobgy I have referred the issue [#1217](https://github.com/kubeflow/manifests/issues/1217) to run my test cases successfully with below Kustomize version 
```
kustomize version
Version: {Version:kustomize/v3.2.1 GitCommit:d89b448c745937f0cf1936162f26a5aac688f840 BuildDate:2019-09-27T00:10:52Z GoOs:linux GoArch:amd64}
```
I have created a new PR with the updated successful test case run. Please review the PR [#1332](https://github.com/kubeflow/manifests/pull/1332)

PR #1218, opened 2020-06-01
Manifest changes to support GCP blueprints on private GKE.

jlewi:
* kubeflow/gcp-blueprints#33 is tracking GCP blueprints on private GKE with VPC-SC

  * This PR doesn't fully enable that but it includes a lot of necessary
    changes.

* cluster-private-patch.yaml is a cluster patch that turns on a lot of
  settings to deploy GKE with private GKE

  * For ease of use we make the master publicly accessible anywhere; users
    could configure that behavior if desired using patch overlays.

* Use kpt setters to name all the networking resources (firewall rules, networks, etc...)

  * This ensures the names are unique based on the KF deployment name and won't conflict with
    existing rules.

  * The setters also ensures that the references get set correctly; e.g. the firewall rules
    correctly refer the newly created network.

* Add a CNRM resource to enable CloudDNS.

  * Per kubeflow/gcp-blueprints#31 we should probably use CNRM and not AnthosCLI to enable
    all required services.

* Add a kpt setter to control firewall rule logging

  * Enabling firewall rule logging can be useful to debug why connections are blocked.

    Enable logging on firewall rules.

* Add an extra firewall rule for ISTIO

  *Per https://istio.io/docs/setup/platform-setup/gke/ we need to manually create an additional firewall rule to allow traffic to the ISTIO pilot webhook port.

* Add a NAT to allow outbound internet egress

  * Egress is still blocked by firewall rules
  * Per kbueflow/gcp-blueprints#34 this was an attempt to make it possible
    to pull images from DockerHub and Quay.IO. This was partially
    succesful; pulling from DockerHub works but for Quay.IO the firewall
    rules are strill blocking required connections.

* Fix the v3 version of the cert-manager package.

  * #1134 moved the kubeflow issuer into its own package to avoid
    race conditions

   * That refactored means that the v3 packages no longer included the
     actual cert-manager resources
   * This PR fixes that by having the v3 package pull in the base package

Resolves #

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`


Bobgy:
> For ease of use we make the master publicly accessible anywhere; users
could configure that behavior if desired using patch overlays.

Is this always a risk? e.g. will other users be able to access that api server?

other than that
/lgtm

jlewi:
> Is this always a risk? e.g. will other users be able to access that api server?
@Bobgy This means the master is accessible from any IP but there is still authorization enabled. 

See here for more info:
https://cloud.google.com/kubernetes-engine/docs/how-to/private-clusters#endpoints

PR #1220, opened 2020-06-01
Add Katib CRD to components

andreyvelich:
I added Katib CRDs to components, see discussion: https://github.com/kubeflow/manifests/pull/1124#issuecomment-618634808.

Also, I added `commonLabels` to `components/katib-db-manager` and `components/katib-db-mysql` and I removed `commonLabels` from `installs/katib-standalone` since it is not required.

@jlewi What do you think about Katib manifests v3 version? `katib/installs` and `katib/components` satisfies the V3 manifests rules.
I can delete current v3 manifests and make `installs` and `components` under `v3` folder.

/assign @jlewi @gaocegege @johnugeorge 

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

### DIFF
--- a/katib/components/katib-db-mysql/kustomization.yaml
+++ b/katib/components/katib-db-mysql/kustomization.yaml
@@ -10,3 +10,6 @@ images:
   - name: mysql
     newTag: "8"
     newName: mysql
+commonLabels:
+  app.kubernetes.io/component: katib
+  app.kubernetes.io/name: katib-controller
--- a/katib/components/katib-crds/kustomization.yaml
+++ b/katib/components/katib-crds/kustomization.yaml
@@ -1,0 +1,11 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: kubeflow
+resources:
+  - ../../katib-crds/base/experiment-crd.yaml
+  - ../../katib-crds/base/suggestion-crd.yaml
+  - ../../katib-crds/base/trial-crd.yaml
+  - ../../katib-crds/overlays/application/application.yaml
+commonLabels:
+  app.kubernetes.io/component: katib
+  app.kubernetes.io/name: katib-crds
--- a/katib/components/katib-db-manager/kustomization.yaml
+++ b/katib/components/katib-db-manager/kustomization.yaml
@@ -8,3 +8,6 @@ images:
   - name: gcr.io/kubeflow-images-public/katib/v1alpha3/katib-db-manager
     newTag: v0.8.0
     newName: gcr.io/kubeflow-images-public/katib/v1alpha3/katib-db-manager
+commonLabels:
+  app.kubernetes.io/component: katib
+  app.kubernetes.io/name: katib-controller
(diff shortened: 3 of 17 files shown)


katib/components/katib-db-mysql/kustomization.yaml
```diff
@@ -10,3 +10,6 @@ images:
   - name: mysql
     newTag: "8"
     newName: mysql
+commonLabels:
+  app.kubernetes.io/component: katib
+  app.kubernetes.io/name: katib-controller
```
jlewi, line 15: Is this the right name? Did you mean to call everything katib-controller even if its a db?

jlewi:
> I can delete current v3 manifests and make installs and components under v3 folder

I didn't quite understand this. What are you proposing?

katib/components/katib-db-mysql/kustomization.yaml
andreyvelich: Currently, it works like this. 
We have one Application: https://github.com/kubeflow/manifests/blob/master/katib/katib-controller/overlays/application/application.yaml, for katib-controller, katib-db-manager and katib-mysql with this label (https://github.com/kubeflow/manifests/blob/master/katib/katib-controller/overlays/application/kustomization.yaml#L4).
If we want to make unique label for each component, do we need to create individual application for them?

andreyvelich:
> > I can delete current v3 manifests and make installs and components under v3 folder
> 
> I didn't quite understand this. What are you proposing?

I mean, we created Katib V3 manifest for `katib-controller` folder (https://github.com/kubeflow/manifests/tree/master/katib/katib-controller/v3) and `katib-crds` folder (https://github.com/kubeflow/manifests/tree/master/katib/katib-crds/v3).

After we migrated to folder structure `katib/components` and `katib/installs` these folders (`katib/katib-controller` and `katib/katib-crds`) are legacy (but we still use them in kfdef manifests). 

In the new V3 version, I think we can use this new structure with components and installs for Katib.
So we can remove V3 version that related to old folders.

And I am asking do we need to point that new folder structure is related to V3 or it is not necessary?

jlewi:
> In the new V3 version, I think we can use this new structure with components and installs for Katib.
So we can remove V3 version that related to old folders.
Sounds good to me.

> And I am asking do we need to point that new folder structure is related to V3 or it is not necessary?

I may not have understood what you are saying here. I think using the new folder structure is fine. You don't need to call it V3 .

/lgtm
/approve

/hold 
In case you want to do the additional cleanup in this PR; otherwise you can remove the hold and submit as is.

andreyvelich:
> > In the new V3 version, I think we can use this new structure with components and installs for Katib.
> > So we can remove V3 version that related to old folders.
> > Sounds good to me.
> 
> > And I am asking do we need to point that new folder structure is related to V3 or it is not necessary?
> 
> I may not have understood what you are saying here. I think using the new folder structure is fine. You don't need to call it V3 .
> 
> /lgtm
> /approve
> 
> /hold
> In case you want to do the additional cleanup in this PR; otherwise you can remove the hold and submit as is.

Thanks! I will submit another PR to remove old v3 folders.

PR #1229, opened 2020-06-04
Changed OIDC EnvoyFilter to fit recent API

caffeinism:
The method of describing envoy filters in istio has changed since v1.3. The manifest of kubeflow-istio-dex uses version 1.3 of istio, but envoyfilter follows the old api format. Since istio moved to 1.6 (latest version), the API format is in place, and since this API format can express more abundantly, I think it is correct to change to the latest API format.

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`


caffeinism:
@googlebot I signed it!

krishnadurai:
@caffeinism can you please generate tests for this?

Issue #1063, opened 2020-03-30
GCP KFDef's should follow kustomize v3 patterns


jlewi:
Ref: http://bit.ly/kf_kustomize_v3
        #1062

GCP KFDef patterns should follow kf_kustomize_v3 patterns. See #1062 for more info.

jlewi:
There is a prototype PR in #962.  It is being split up and merged as individual PRs.

jlewi:
The initial stack has been committed. Next steps

* Setup auto deployment of the new stack
* Setup E2E tests to run against this so we can start getting signal about what doesn't work.

jlewi:
For 1.1 we will probably provide a story for deploying on GCP using anthoscli, kpt, and kustomize. This will use/depend on the new V3 manifests.

It remains to be seen whether we will also provide a KFDef.

jlewi:
Preliminary blueprints are in: https://github.com/kubeflow/gcp-blueprints

jlewi:
A couple work items remain to close this out

* #1237 remove the KFDefs to avoid confusion
* Add Katib to the blueprint
* Add pipelines to the blueprint
* Rename https://github.com/kubeflow/manifests/tree/master/stacks/gcp to "generic" or something else
    * It will not be used for GCP
    * It is still useful as an example

PR #1237, opened 2020-06-09
Remove GCP KFDefs and E2E tests.

### LINKED ISSUE #818: Need to modularize scope manifests tests to reduce flakiness
Right now when we submit a PR to manifests we end up triggering tests for all applications (e.g. ones that have them).

That's unnecessary. As an example in kubeflow/manifests #4699 we are modifying only the jupyter-web-app manifests but we end up triggering tests for pytorch.

This is not scalable and will just make the tests more flaky as we add more applications.

Modifications to manifests for application A should only trigger tests for application A.

To realize that I think we need a more modular way to write our tests. e.g. we want to have a "module" containing the tests for each Application.

We can then compose these modules into different tests. So for example modifying the KFDef spec should potentially trigger the tests for all applications.

One way to do this would be to rely on Tekton

1. Define a Tekton Task for each application's tests
1. Define Pipeline's corresponding to different combinations of Task
    e.g one pipeline to just run App A's test and one pipeline to run all App tests
1. Use our existing "include" feature to determine which pipelines to trigger based on the files modified.

To do this we would need to

1. Define a catalog of reusable tasks and pipelines
1. Sync that catalog to our test cluster
1. Figure out where specific pipelineruns for each repo would be defined
1. Modify the test infra to trigger those runs.

jlewi:
* Per #1063 with 1.1 GCP will no longer use KFDef to deploy
  Kubeflow; isntead we will be using blueprints kubeflow/gcp-blueprints

  * So we remove all the GCP KFDefs

  * Cleanup the kfdef/OWNERs; remove Googlers since there are no longer
    anymore GCP configs.

* Fix #818 - stop running E2E tests that deploy Kubeflow on every
  commit.

  * We now have unittests aimed at validating the manifests; so it
    no longer makes sense to fully deploy KF and verify its working.

    * E2E tests should now happen downstream.

    * We also need to remove the tests since we are removing GCP KFDefs on
      which they depend.


Bobgy:
makes sense
/lgtm
/approve

krishnadurai:
Sure @jlewi. I've created a tracking issue with what I think would be missed without an E2E test.

/hold cancel

krishnadurai:
You can self approve.

PR #1238, opened 2020-06-09
Make the KFDef stacks example non GCP specifc.

jlewi:
* For GCP will not be using KFDef with 1.1; instead we will be using
  blueprints (see #1063)

* Therefore I updated the existing example of deploying using stacks
  to make it non GCP specific so that it could be used as an example
  for platforms that still want to use KFDef.

  * #1062 is tracking migrating all KFDefs to the new versions
    of our kustomize packages


jlewi:
/assign @yanniszark 
/assign @animeshsingh 
/assign @Jeffwan 

@yanniszark @animeshsingh @Jeffwan who wants to review this?

stacks/examples/kfdef_stacks.experimental.yaml
```diff
@@ -5,15 +5,8 @@ metadata:
...
-  # directories is when the resources needed to be installed in a different namespace.  
-  # TODO(jlewi): This should be replaced by ACM when its ready.
-  # Its a separate package because it needs to be in a different namespace
-  - kustomizeConfig:
-      repoRef:
-        name: manifests
-        path: istio/gcp-1-1-6
-    name: istio-stack
+  # TODO: If you actually want to install ISTIO as part of KFDef setup you would
+  # need to add the appropriate package.
```
Bobgy, line 9: Q: Which istio version should application authors expect? I know for GCP it's 1.4+ (from memory, not sure if accurate). Did we upgrade from 1.1 for others, is there a plan?

e.g. KFP multi user mode will ideally use new authorization CRDs introduced after 1.2, it would make others not compatible

jlewi: I don't think we have published any requirements about minimum or maximum supported ISTIO versions. So right now I would say its up to each individual application owner to decide which versions of ISTIO to be compatible with.

PR #1245, opened 2020-06-11
[AWS] Upgrade ingress APIVersion to support k8s 1.16

### LINKED ISSUE #869: GCP and AWS Change ingress resources to networking.k8s.io/v1
See: https://github.com/kubeflow/kubeflow/issues/4680

https://github.com/kubeflow/manifests/blob/830e1a0c87befdaec217564efb2b06ddc02a0cd6/gcp/iap-ingress/base/ingress.yaml#L1

Need to upgrade to networking.k8s.io/v1 to support newer versions of Kubernetes.

Ref: kubeflow/manifests#375

It looks like this was introduced in 1.8 so there shouldn't be any blocker to upgrading our ingresses now
https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/

Jeffwan:
Resolves #869 

NetworkPolicy in the extensions/v1beta1 API version is no longer served is 1.16. We migrate to use the networking.k8s.io/v1 API version, available since v1.8. 

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

### DIFF
--- a/aws/istio-ingress/base/ingress.yaml
+++ b/aws/istio-ingress/base/ingress.yaml
@@ -1,4 +1,4 @@
-apiVersion: extensions/v1beta1 # networking.k8s.io/v1beta1
+apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
   annotations:


jlewi:
@Jeffwan I think it might still be v1beta1 for ingress not v1
https://kubernetes.io/docs/concepts/services-networking/ingress/

Issue #1048, opened 2020-03-24
[Pipelines] Reuse KFP standalone manifests in kubeflow/manifests


Bobgy:
Part of https://github.com/kubeflow/manifests/issues/1062

Depends on http://bit.ly/kf_kustomize_v3

Above mentioned kustomize v3 plan will allow [Kubeflow Pipelines Standalone deployment manifests](https://github.com/kubeflow/pipelines/tree/master/manifests/kustomize) get reused by Kubeflow.

Following OTS strategy:
* set up a bot that syncs kubeflow pipelines standalone manifest changes with a PR to `kubeflow/manifest/pipelines/standalone` after each release
* in kubeflow/manifest repo, use stacks that refer to and customize kfp standalone manifests' specific components using overlays

Then most kfp upstream manifest changes can be synced to kubeflow with minimal maintenance efforts, because `kubeflow/manifest/pipeline/standalone` folder won't have conflicts

/cc @jlewi

jlewi:
@Bobgy To complete http://bit.ly/kf_kustomize_v3 we will want a single kustomize package that installs all the pipeline applications. Should we base that off the manifest that's currently in kubeflow/manifests/pipelines or should we copy for the kfp standalone manifests?

Bobgy:
@jlewi I think copying from kfp standalone manifests is better considering the long term goal.

No matter what, http://bit.ly/kf_kustomize_v3 has to be a big rewrite, so I think it will be cost efficient if we can do both at the same time.

Bobgy:
Some parts of kfp standalone manifests may not be generic enough, I will also structure it using patterns introduced in http://bit.ly/kf_kustomize_v3 to get a good base.

May I ask what would be your rough timeline for kustomize v3 plan, will it be part of KF 1.1? My plan is pretty packed already until end of April, just heads up.

Bobgy:
I can actually adjust my plan accordingly depending on timeline for this. This saves a lot of efforts for KFP and KF's integration.

jlewi:
Yes we are targeting Kubeflow 1.1 which is end of june.

Thanks

jlewi:
@Bobgy any update on this? It would be great to have an updated manifest that we could begin testing with GCP blueprints and multiuser.

Bobgy:
@jlewi Thanks for the reminder. A quick question: is there still any value merging KFP multi user mode to non v3 manifest?

jlewi:
[...]

Issue #897, opened 2020-02-11
Enable Metadata Writer for KFP


jlewi:
See this comment
https://github.com/kubeflow/manifests/issues/796#issuecomment-584491877

KFP and metadata need to be configured correctly to enable the metadata writer service.

/cc @elikatsis @Ark-kun

jlewi:
@Ark-kun any interest in picking this up?

Some relevant links
https://github.com/kubeflow/pipelines/tree/master/backend/metadata_writer

elikatsis:
@jlewi We first need to merge https://github.com/kubeflow/pipelines/pull/3108

jlewi:
@elikatsis @dushyanthsc any update on this?

elikatsis:
Hi @jlewi,

I believe MetadataWriter implementation as well as the related KFP manifests are compatible with KF manifests. So it should be easy to extend these manifests to deploy metadata writer.

I'll let @dushyanthsc or @rmgogogo verify.

dushyanthsc:
We updated the manifest to be consistent. Is there anything else required to change on the Metadata (gRPC) deployments to enable metadata-writer @Ark-kun ?

jlewi:
Downgrading to p2 and removing from KF 1.1. because it doesn't look like we have a clear owner so it doesn't seem like its going to get completed.

/cc @aronchick @karlschriek

Ark-kun:
> Is there anything else required to change on the Metadata (gRPC) deployments to enable metadata-writer @Ark-kun ?

No. Metadata Writer is standalone and should be compatible with the rest of the deployments (after the GRPC vs. HTTP service names were fixed).

Ark-kun:
@jlewi I think we just need to add the Metadata Writer deployment (deployment, SA, role, binding) manifests to KF.

Ark-kun:
Looks like https://github.com/kubeflow/manifests/pull/1248 fixes this issue.

PR #1248, opened 2020-06-12
Pipelines kustomize v3 manifest

Bobgy:
Resolves https://github.com/kubeflow/manifests/issues/1048
Resolves https://github.com/kubeflow/manifests/issues/897
Resolves https://github.com/kubeflow/pipelines/issues/3993

Add kustomize v3 manifest for pipelines.
In addition to existing pipelines components, I also integrated the newly introduced components:
* cache server
* metadata writer

/cc @Ark-kun 

I verified these with gcp blueprint deployment, caching and metadata writer both works properly.

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`


Bobgy:
@jlewi I removed most of the vars after rebasing to kfp rc2 manifest with changes in https://github.com/kubeflow/pipelines/pull/3978.

The remaining usages of var are:
* virtual service: clusterDomain and namespace, https://github.com/kubeflow/manifests/pull/1248/files#diff-adc2be74d4efe3becc4b6ea7bec4d15bR37
* minio and mysql: pv name and pd name
https://github.com/kubeflow/manifests/pull/1248/files#diff-adc2be74d4efe3becc4b6ea7bec4d15bR37 and https://github.com/kubeflow/manifests/blob/f68a7fbb2917d450b685da800d3d4612d14536d2/pipeline/minio/installs/gcp-pd/kustomization.yaml#L16-L32 ( they are using the same pattern )

I'm not sure why we are parameterizing PV name, it was consistent with KF 1.0 manifest. I'll take a look at commit history to find why. I'm guessing that's because persistence volume is immutable after creation, so for proper devops action of changing a PV, it's best to customize the name to use a new PV.

Because this PR is huge. I suggest we separate a PR for how to get rid of the last 4 vars (if we do want to).

Bobgy:
I did some investigation, the PV, PVC name parameterization by var was added in https://github.com/kubeflow/manifests/pull/104. However, it didn't explain any reasoning about why they were added.

@gabrielwen Do you have context why we needed to parameterize PV and PVC name for pipeline's mysql and minio db?

jlewi:
Thanks @Bobgy I think most of the vars you mentioned are for internal substitution; e.g.
namespace. Per the [guide](https://github.com/kubeflow/manifests/blob/master/docs/KustomizeBestPractices.md#internal-subsitution-of-fields-kustomize-isnt-aware-of) this is ok as long as we give the vars unique enough names so we don't conflict with similar vars from other packages.

A better solution for clusterDomain is being tracked in https://github.com/kubeflow/manifests/issues/1007#issuecomment-619404552

You can keep that one as is for now. On GCP we can probably just switch to `kpt setters` since we are already using `kpt setter`.

> @gabrielwen Do you have context why we needed to parameterize PV and PVC name for pipeline's mysql and minio db?

I think it might have been related to how we were creating the PD. I think we may have been using deployment manager (and now KCC) to create the PD and thus needed to substitute in the correct name to the package. 

I think the reasoning was we wanted to use KCC/DM and not the dynamic provisioner to create the PD because we wanted to make sure the PD wasn't destroyed if KF was deleted because we didn't want to destroy any data.

I'm good with submitting PR as is and filing issues to track fixing things later.

Bobgy:
@jlewi the conflicts were caused by `env` to `envs` change: https://github.com/kubeflow/manifests/commits/master/pipeline/upstream

but kfp standalone manifests don't support `envs` yet, because we wanted to maintain `kubectl kustomize` compatibility, which uses a kustomize v2.x version: https://github.com/kubernetes-sigs/kustomize#kubectl-integration.

I'll ask the PR owner for a resolution.

Bobgy:
I updated kfp pull upstream script to use `sed` for a quick HACK that transforms `env:` to `envs:` during pull: https://github.com/kubeflow/manifests/pull/1248/commits/a842b7b042ed22edbd37f2c4678fe0121e8aa3a4

What do you think?

jlewi:
SGTM

Although you could also write a kpt FN transformer.

kustomize 2.x is very old. A lot of things won't work if you rely on it.

/lgtm
/approve

Bobgy:
> Although you could also write a kpt FN transformer.

Thanks for sharing, will try that next time


> kustomize 2.x is very old. A lot of things won't work if you rely on it.

convenience from kubectl is the major reason we keep using it. We will consider upgrading when we get the first blocking issue

PR #1284, opened 2020-06-18
Add namespace selector and test case for admission

### LINKED ISSUE #1213: Unittest to validate admission web hooks don't apply to themselves
We've experienced a number of problems in Kubeflow where admission webhooks don't use a namespace or objectselector to prevent the admission hook from applying to itself.

This can cause outages  e.g.

kubeflow/kubeflow#4808
kubeflow/kfserving#568

We should be able to prevent this by ensuring Kubeflow admission hooks use a namespace selector or object selector to prevent them from applying to themselves.

We should add a unittest to try to verify this. We can model the unittest on
https://github.com/kubeflow/manifests/blob/master/tests/validate_resources_test.go

This test searches all KF manifests and verifies that certain resources conform to certain standards.

Tomcli:
Resolves #1213 
Resolves kubeflow/kubeflow#4808

- Update the profile control to the latest version, which includes the `app.kubernetes.io/part-of: kubeflow-profile` label when creating a new namespace
- Update admission mutating webhook with namespaceSelector using the new label.
- Add new unit case to make sure all the mutating webhooks that change the pod will need to have either namespaceSelector or objectSelector.

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

### DIFF
--- a/admission-webhook/webhook/base/mutating-webhook-configuration.yaml
+++ b/admission-webhook/webhook/base/mutating-webhook-configuration.yaml
@@ -10,6 +10,9 @@ webhooks:
       namespace: $(podDefaultsNamespace)
       path: /apply-poddefault
   name: $(podDefaultsDeploymentName).kubeflow.org
+  namespaceSelector:
+    matchLabels:
+      app.kubernetes.io/part-of: kubeflow-profile
   rules:
   - apiGroups:
     - ""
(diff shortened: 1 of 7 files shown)


jlewi:
@Tomcli thanks so much for doing this!

/lgtm
/approve

PR #1287, opened 2020-06-19
Update ingresses and deployments to work with 1.16

### LINKED ISSUE #375: Numerous kubeflow manifests can't deploy on Kubernetes 1.16
Old declaration of key objects such as DaemonSet, StatefulSet and Deployment can't be deployed on Kubernetes 1.16 anymore if they are using the old apiVersion such as extensions/v1beta<X> or the apps/v1beta<X> 

- Deployments are supposed to use apps/v1 since kubernetes 1.9. => Need updates.
- Ingress will using extensions/xxx will be rejected in kubernetes 1.20 => Need updates.
- CRD definition just started to use v1 in K8s. To ensure they still works => Unchanged.

K8s 1.16 rejects Deployment for which the selector has not been set => Needs updates.

jlewi:
* ingress is now networking.k8s.io/v1beta1 this was introduced in
  in 1.8 so there shouldn't be any blocker to upgrading our ingresses now
  https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/

  * Most of the ingresses had already been updated

* Deployments are now apps/v1 not extensions/v1beta1

* Delete gcp/basic-auth-ingress rather than updating it.

* Delete kfdef/generic/istio rather than update it

  * It doesn't look like this has been modified since it was copied over
    from the kubeflow/kubeflow repo in 2019. ISTIO configs should be in /istio

* Add a test for deprecated Kubernetes resources fixes #375


jlewi:
/assign @yanniszark 
/assign @Jeffwan 

@Jeffwan  Could you please LGTM AWS changes?

@yanniszark can you please confirm we can delete the istio config in kfdef/generic?

yanniszark:
> @yanniszark can you please confirm we can delete the istio config in kfdef/generic?

I confirm.

jlewi:
Tests are failing for unrelated changes which will be fixed by #1288 so we need to merge that first.

PR #1299, opened 2020-06-22
Add Support For AWS Dex Basic-Authentication

### LINKED ISSUE #1164: [AWS] Consider to use dex solution for basic aws manifest
https://github.com/kubeflow/manifests/blob/master/kfdef/source/master/kfctl_aws.yaml

We used a public application load balancer in the past. With some security concerns, we remove public one and ask user to use port-forward way to check dashboard. 

Recently we get a few asks from users on basic authentication, multi-user for [non-cognito manifest]((https://github.com/kubeflow/manifests/blob/master/kfdef/source/master/kfctl_aws_cognito.yaml) ).

Since community already has dex solution, I think we should consider to reuse this for  `kfctl_aws.yaml`

There're a few good things
- AWS folks can also help maintain and improve Dex solution. 
- Even Coginito setup is recommended and it's production grade solution, it's kind of complicated for users just like to try Kubeflow, for example, for workshop purpose. 
- two AWS manifest will be multi-user enabled and all our examples can be unified. 

Engineer stories
- Update kfctl_aws.yaml to include dex manifests
- Update kfctl to support new manifest orchestration
  - Add logic to handle basic auth
  - Handle dex installation in kfctl aws plugin 
- Still consider to use Application Load balancer instead of ELB (serviceType=Loadbalancer).

PatrickXYS:
Resolves #1164 

Integrated Dex as basic authentication option in AWS Kubeflow.

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

### DIFF
--- a/kfdef/kfctl_aws.yaml
+++ b/kfdef/kfctl_aws.yaml
@@ -5,46 +5,6 @@ metadata:
 spec:
   applications:
   - kustomizeConfig:
-      parameters:
-      - name: namespace
-        value: istio-system
-      repoRef:
-        name: manifests
-        path: istio/istio-crds
-    name: istio-crds
-  - kustomizeConfig:
-      parameters:
-      - name: namespace
-        value: istio-system
-      repoRef:
-        name: manifests
-        path: istio/istio-install
-    name: istio-install
-  - kustomizeConfig:
-      parameters:
-      - name: namespace
-        value: istio-system
-      repoRef:
-        name: manifests
-        path: istio/cluster-local-gateway
-    name: cluster-local-gateway
-  - kustomizeConfig:
-      parameters:
-      - name: clusterRbacConfig
-        value: 'OFF'
-      repoRef:
-        name: manifests
-        path: istio/istio
-    name: istio
-  - kustomizeConfig:
-      parameters:
-      - name: namespace
-        value: istio-system
-      repoRef:
-        name: manifests
-        path: istio/add-anonymous-user-filter
-    name: add-anonymous-user-filter
-  - kustomizeConfig:
       repoRef:
         name: manifests
         path: application/application-crds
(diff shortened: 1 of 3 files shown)


dex-auth/dex-crds/base/params.env
```diff
@@ -2,8 +2,8 @@
 dex_domain=dex.example.com
 # Set issuer to https if tls is enabled
 issuer=http://dex.example.com:32000
-static_email=admin@example.com
```
Jeffwan, line 5: This is a default change. User who not explicitly specify credentials may need it. Please don't touch it.

kfdef/kfctl_aws.yaml
```diff
@@ -4,58 +4,50 @@ metadata:
...
         name: manifests
         path: istio/istio
     name: istio
-  - kustomizeConfig:
-      parameters:
-      - name: namespace
-        value: istio-system
-      repoRef:
-        name: manifests
-        path: istio/add-anonymous-user-filter
```
Jeffwan, line 45: Good. We don't need this anymore

kfdef/kfctl_aws.yaml
```diff
@@ -350,6 +387,16 @@ spec:
         name: manifests
         path: mpi-job/mpi-operator
     name: mpi-operator
+  - kustomizeConfig:
+      overlays:
```
Jeffwan, line 391: Why do you need cognate overlay?

dex-auth/dex-crds/base/params.env
Jeffwan: This is used by a few  kfdefs

kfdef/kfctl_aws.yaml
PatrickXYS: We don't need, copy & paste error here, fixed.

dex-auth/dex-crds/base/params.env
PatrickXYS: Yeah I agree, fixed

kfdef/kfctl_aws.yaml
```diff
@@ -374,9 +419,9 @@ spec:
     spec:
       auth:
         basicAuth:
-          password:
-            name: password
-          username: admin
+          password: $2y$12$ruoM7FqXrpVgaol44eRZW.4HWS8SAvg6KYVVSCIwKQPBmTpCm.EeO
```
Jeffwan, line 422: Does user need to pass the encoded string?

PatrickXYS: Right, fixed.

Issue #1252, opened 2020-06-15
Manifests release 1.1


krishnadurai:
Opening this issue to track releasing manifests as part of 1.1.

- Ref: kubeflow/kubeflow#5022
- We will need to
    - [x] Cut a release branch
    - [ ] Generate Manifests for each kfdef

/assign @krishnadurai

jlewi:
Looks like the postsubmit tests are failing. Looks like the TFJob operator tests are out of sync.

[kubeflow-manifests-postsubmit-unit-d6a25c6-5684-13a5-1216454708.log.txt](https://github.com/kubeflow/manifests/files/4813760/kubeflow-manifests-postsubmit-unit-d6a25c6-5684-13a5-1216454708.log.txt)

jlewi:
@krishnadurai Can you fix the tests and then once the tests are fixed cut the branch?

jlewi:
@krishnadurai looks like the tests are still failing

```
===== ACTUAL END ==========================================
  EXPECTED                                                                ACTUAL
  --------                                                                ------
  apiVersion: admissionregistration.k8s.io/v1beta1                        apiVersion: admissionregistration.k8s.io/v1beta1
  kind: MutatingWebhookConfiguration                                      kind: MutatingWebhookConfiguration
  metadata:                                                               metadata:
    annotations:                                                            annotations:
      cert-manager.io/inject-ca-from: kubeflow/admission-webhook-cert         cert-manager.io/inject-ca-from: kubeflow/admission-webhook-cert
    labels:                                                                 labels:
      app: admission-webhook                                                  app: admission-webhook
      app.kubernetes.io/component: poddefaults                                app.kubernetes.io/component: poddefaults
      app.kubernetes.io/name: poddefaults                                     app.kubernetes.io/name: poddefaults
      kustomize.component: admission-webhook                                  kustomize.component: admission-webhook
    name: admission-webhook-mutating-webhook-configuration                  name: admission-webhook-mutating-webhook-configuration
  webhooks:                                                               webhooks:
  - clientConfig:                                                         - clientConfig:
      caBundle: ""                                                            caBundle: ""
      service:                                                                service:
        name: admission-webhook-service                                         name: admission-webhook-service
        namespace: kubeflow                                                     namespace: kubeflow
        path: /apply-poddefault                                                 path: /apply-poddefault
    name: admission-webhook-deployment.kubeflow.org                         name: admission-webhook-deployment.kubeflow.org
[...]

PR #1300, opened 2020-06-22
Synchronize the tests

jlewi:
* Related to #1252

### DIFF
--- a/tests/stacks/ibm/test_data/expected/~g_v1_service_mysql.yaml
+++ b/tests/stacks/ibm/test_data/expected/~g_v1_service_mysql.yaml
@@ -10,6 +10,8 @@ metadata:
 spec:
   ports:
   - port: 3306
+    protocol: TCP
+    targetPort: 3306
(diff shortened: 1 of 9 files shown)


tests/stacks/ibm/test_data/expected/~g_v1_service_mysql.yaml
```diff
@@ -10,6 +10,8 @@ metadata:
 spec:
   ports:
   - port: 3306
+    protocol: TCP
+    targetPort: 3306
```
animeshsingh, line 14: believe the ports are generic and independent of cloud provider - and accessed internally for mysql/minio

jlewi: @animeshsingh This is the expected output i.e. the results of running `kustomize build`. So if you think this expected output is not "correct" it indicates a bug in the kustomize manifest somewhere.

PR #1311, opened 2020-06-23
KFServing and KNative manifests Owners files

animeshsingh:

### DIFF
--- a/kfserving/OWNERS
+++ b/kfserving/OWNERS
@@ -1,5 +1,5 @@
 approvers:
    - adrian555
    - animeshsingh
-   - dansun
    - Tomcli
+   - yuzisun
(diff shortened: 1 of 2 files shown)


animeshsingh:
@adrian555 @Tomcli @yuzisun please lgtm

additionally @yuzisun do we want to add someone else from bloomberg?

animeshsingh:
thanks @krishnadurai for pointing that

didn't see that before - but I don't expect these files to be duplicate of OWNERS file in actual application repos - the focus here is folks who would be focussing on "integration" and "integration-testing" with Kubeflow as opposed to development of application itself.

cc @yuzisun

animeshsingh:
on that note, ideally @krishnadurai you should be in KNative owners file as well, if you don't mind to be added

krishnadurai:
@animeshsingh go ahead to add my name.

PR #1341, opened 2020-07-03
Update to KFP 1.0.0-rc.3

Bobgy:
**Checklist:**
    1. `cd manifests/tests`
    3. `make test`


Bobgy:
/assign @jlewi 
Major manifest improvements in rc3:
* mlpipeline-minio-artifact-secret is contained in minio package and referred to in different deployments
* kustomization.yaml images field are pushed to lowest level packages, so that when adding overlays on top of it, we don't need to add `images` filed override again.
* metadata-writer has its own package now, so there's no need to refer to its individual components

Issue #1136, opened 2020-04-25
kustomize package to setup a namespace for using Kubeflow (alternative to profile controller)


jlewi:
We need a kustomize package that sets up a namespace for using Kubeflow. This package should basically do everything the profile controller is doing. e.g.

* Create default service accounts
* Create role bindings
* Platform specific overlays (e.g. setup workload identity for GCP)

A declarative approach to namespace setup should better support platform teams looking to setup and control kubeflow deployments on behalf of multiple teams. See the discussion in this [thread](https://groups.google.com/forum/#!topic/kubeflow-discuss/whS_j4RlEJo)

Creating a controller to setup namespaces made sense if we were trying to create an abstraction to hide the details of namespace to support programmatic creation e.g. through the UI.

However, as discussed in that thread we are seeing friction around that approach particularly for platform teams.

To support platform-teams I think we want to move in the direction of GitOps. If a team is using GitOps to bootstrap individual namespaces then using a controller makes less sense then just providing a kustomize package for the relevant K8s resources.

lalithvaka:
Some additional features to think about as following
1. Kustomize Declarative approach for ResourceQuota for each namespace
2. Group based profile / namespace support
3. Multi owner support for profile / namespace
2. Externalized enable/disable profile self serv capability - (Enable in Dev Env but disable in Prod)
3. Default resource quota setup for self serv profiles if they are enabled (Declarative?).

discordianfish:
In general I like a declarative approach but as described here, it would require some glue I'd prefer we could avoid:

We're using an authenticating (via LDAP) reverse proxy in front of kubeflow that sets trusted user identity headers. That means right now we just need to grant users access to that proxy and when they log in the first time, they can create their namespace. With the proposed change, we would have add glue to poll LDAP and add/delete namespaces from the kustomize package dynamically.

That's definitely doable but I feel like many people would have to solve that somehow.

So the ideal solution to me would be if the profile controller creates a namespace (non-interactively) automatically based on some sort of declarative template.

jlewi:
@discordianfish  

> So the ideal solution to me would be if the profile controller creates a namespace (non-interactively) automatically based on some sort of declarative template.

I don't think its necessarily a one size fits all. If you want to to propose and build out the case for doing things programmatically/automatically that sounds good to me.

On GCP what we are seeing is that 

1. Platform teams want to own creating and setting up namespaces
1. GitOps is becoming the pattern of choice for managing infrastructure

[...]

PR #1342, opened 2020-07-03
feat: KFP multi user mode PR1 - enable multi user mode without istio authorization

Bobgy:
Resolves https://github.com/kubeflow/pipelines/issues/4159

This is the first step of enabling KFP multi user mode.
This makes all KFP multi user features available, but not all in-cluster traffic are protected by istio.

This is a good intermediate state to break up the work of enabling multi user mode.

Changes:
* Add a kustomization package in `pipeline/installs/multi-user` which is a drop-in replacement for `pipeline/install/generic`, but multi user mode enabled.

Note, `stacks/gcp` is still pointing to `pipeline/installs/generic` because we need to discuss how KFP multi user mode should be released -- as the single option or allow users to choose from both.

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`


pipeline/installs/multi-user/pipelines-profile-controller/sync.py
```diff
@@ -0,0 +1,274 @@
+# Copyright 2020 Google LLC
```
jlewi, line 1: What is this doing?

pipeline/installs/multi-user/pipelines-profile-controller/sync.py
```diff
@@ -0,0 +1,274 @@
...
+        desired_resources = [
+            {
+                "apiVersion": "v1",
+                "kind": "Secret",
+                "metadata": {
+                    "name": "mlpipeline-minio-artifact",
+                    "namespace": namespace,
+                },
+                "data": {
+                    "accesskey": "bWluaW8=",  # base64 for minio
```
jlewi, line 60: Are we really checking in access keys? Is this safe?

pipeline/installs/multi-user/pipelines-profile-controller/sync.py
jlewi: @Bobgy it looks like this is a controller to setup a namespace for pipelines? Instead of using a controller why  not just publish a kustomize / kpt package that makes it easy to spin up a namespace with all the resources a user needs?
This is in line with
https://github.com/kubeflow/manifests/issues/1136

Bobgy: Your understanding is correct. #1136 does seem a better approach to me!

I'd prefer merging this as is, so that we can start testing KFP multi user mode.
Then we can remove it when #1136 is done. I'd be happy to contribute a package for setting up KFP namespace.

Bobgy: What do you think?

pipeline/installs/multi-user/pipelines-profile-controller/sync.py
Bobgy: Yes, it would be recommended to customize it after deployment.

pipeline/installs/multi-user/pipelines-profile-controller/sync.py
jlewi: Why is this blocked on #1136?  It looks like your controller is just creating a bunch of K8s resources in the namespace in which KFP is going to be used.

Could you just create a kustomize package containing those resources and then tell users that to setup a namespace to
use KFP they need to do something like

```
<kustomize it for their namespace; e.g. using kpt or >
kustomize build  | kubectl apply -f -
```

Bobgy: I think current profile controller provides self service, everyone coming can just use right away. It makes more sense if the deployment is done automatically for them.

I understand customization needs exist for sure, #1136 will likely be the best way forward. I need some time to think about it. Efforts lacking is setting up the package and some documentation to use them. I'd want to avoid doing too much things in one step.

My personal priority is getting this PR and a second PR with istio authorization in, so that previously designed KFP multi user mode is working properly first. Then I can start thinking about substantial improvements like the approach in #1136 .

Bobgy:
Thanks @jlewi !

pipeline/installs/multi-user/pipelines-profile-controller/composite-controller.yaml
```diff
@@ -0,0 +1,43 @@
...
+    updateStrategy:
+      method: InPlace
+  - apiVersion: rbac.istio.io/v1alpha1
+    resource: servicerolebindings
+    updateStrategy:
+      method: InPlace
+  hooks:
+    sync:
+      webhook:
+        url: http://kubeflow-pipelines-profile-controller/sync
```
maganaluis, line 43: @Bobgy This will likely fail on other platforms or on-prem installations, I had to point to the internal service name: 

```url: http://profile-controller.kubeflow.svc.cluster.local:80/sync  ```

Perhaps this can be configurable via Kustomize vars.

Bobgy: Thanks.

is using prefix a GCP specific thing?
I thought it's kubernetes standard behavior.

let me investigate

Jeffwan:
@Bobgy Multi-user manifest changes will only in pipeline/installs?  Any plans to add support for non-v3 manifest? https://github.com/kubeflow/manifests/tree/master/pipeline

Bobgy:
> @Bobgy Multi-user manifest changes will only in pipeline/installs? Any plans to add support for non-v3 manifest? https://github.com/kubeflow/manifests/tree/master/pipeline

@Jeffwan The next PR of enabling istio authorization will add sth to `stacks/gcp`, but this one is completely platform agnostic.
There will be no plan for KFP team to maintain non-v3 manifest because of extra cost.

Issue #1364, opened 2020-07-08
[Pipelines Multi User] integrate with non GCP platforms


Bobgy:
I've merged [the first PR](https://github.com/kubeflow/manifests/pull/1342) for enabling KFP multi user mode.

Platform owners can include this generic module https://github.com/kubeflow/manifests/tree/master/pipeline/installs/multi-user, instead of `pipeline/installs/generic` to test KFP multi user mode right now.

Additional change per platform to integrate with KFP multi user mode:
* KFP api server will be protected by istio authz, only requests with user identity that can pass the public endpoint auth have a chance getting accepted (the mechanism is similar to notebook servers). Therefore, KFP sdk client needs to be able to authenticate for other platforms' auth service. I know dex is one thing we need to support, what are other platforms using? (Issue for dex is https://github.com/kubeflow/kubeflow/issues/4912, there has been some progress.)

If your platform cannot integrate well with KFP multi user mode before KF 1.1 release, backup options:
* still use non multi user mode as before
* use multi user mode, but disable authz (I realized we need some code changes in KFP to support this, please let me know if you have interest in this.)

Bobgy:
/cc @Jeffwan @animeshsingh @yanniszark 
Did I miss anyone?

/cc @jlewi 
Can you help me notify other platform owners?

Bobgy:
Also
/cc @maganaluis 
might be able to share his attempts deploying it with istio + dex

Bobgy:
https://github.com/kubeflow/manifests/pull/1365 is also a blocker for other platforms

Bobgy:
The next PR for securing multi user mode with istio mTLS is https://github.com/kubeflow/manifests/pull/1368.
After this PR, KFP multi user mode is complete. I'll use a separate PR to switch GCP stack to it (already tested the manifest locally, everything works well).

Bobgy:
/assign @Jeffwan @animeshsingh @yanniszark @jbottum 
FYI, KFP multi user mode verified working properly in Kubeflow 1.1 on GCP, all the work is done.

Recommend going ahead and try integrating it with your platforms, feel free to let me know if there are any blockers.

PatrickXYS:
AWS test against multi-user KFP here: https://github.com/kubeflow/manifests/pull/1410

PatrickXYS:
> Also
> /cc @maganaluis
> might be able to share his attempts deploying it with istio + dex

Also, AWS has tested dex and multi-user KFP and works fine.

[...]

PR #1365, opened 2020-07-08
Use kubeflow userid header and prefix config for KFP servers

Bobgy:
Part of https://github.com/kubeflow/manifests/issues/1364

This let KFP servers read from common configuration for userid prefix and userid header.


jlewi:
Tests are likely failing because of
https://github.com/kubeflow/testing/issues/723

PR #1368, opened 2020-07-09
feat: KFP multi user mode PR2 - secure KFP with istio mTLS and authz

Bobgy:
Resolves https://github.com/kubeflow/pipelines/issues/4160

Secure KFP multi user mode with istio mutual TLS and authz.

Note, this doesn't let any stacks use kfp multi user mode yet. I'll update gcp stack in a separate PR.

Limitation: as discussed in https://github.com/kubeflow/kubeflow/issues/4865#issuecomment-655900561, we don't protect kfam with istio sidecar. The implication is any workload in the cluster can send requests to kfam to change contributors for namespaces. The security risk will be removed when kfam's responsibility of managing permissions will be removed. (likely not in KF 1.1)

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`


Bobgy:
/assign @jlewi 

I still have some small changes left, but maybe you can review this first.

pipeline/installs/multi-user/istio-authorization-config.yaml
```diff
@@ -0,0 +1,120 @@
...
+metadata:
+  name: ml-pipeline-ui
+  namespace: kubeflow
+spec:
+  rules:
+  - services:
+    - ml-pipeline-ui.kubeflow.svc.cluster.local
+---
+apiVersion: "rbac.istio.io/v1alpha1"
+kind: ServiceRole
```
Bobgy, line 12: Note service role is deprecated for istio, but they still work in istio 1.4.x.

I'd prefer leaving this here for a while and migrate when all platforms upgrade to later istio versions.

Bobgy:
OK, this is ready

PR #1374, opened 2020-07-10
feat: Add application and common labels to KFP and various fixes

Bobgy:
Meets Kubeflow application requirement: https://github.com/kubeflow/community/blob/master/guidelines/application_requirements.md

There were also a few small fixes in this PR:
* kfp profile controller, using deployment v1 api instead, because v1beta1 is deprecated after K8s 1.14
* change `userid-header=x-goog-authenticated-user-email` to lowercase
* fix kfp profile controller b64encoding exception

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`


Bobgy:
The test failure is
```
=== RUN   TestCommonLabelsImmutable
   TestCommonLabelsImmutable: validate_resources_test.go:62: Skipping directory ../.git
   TestCommonLabelsImmutable: validate_resources_test.go:62: Skipping directory ../.github
   TestCommonLabelsImmutable: validate_resources_test.go:88: ../pipeline/installs/generic/kustomization.yaml has forbidden commonLabel app.kubernetes.io/part-of
   TestCommonLabelsImmutable: validate_resources_test.go:88: ../pipeline/installs/multi-user/kustomization.yaml has forbidden commonLabel app.kubernetes.io/part-of
   TestCommonLabelsImmutable: validate_resources_test.go:62: Skipping directory ../tests
```

We haven't published pipeline installation publically, can we bypass the test failure and merge it?

jlewi:
@Bobgy What does it not being public have to do with anything. Its the part-of label that is problematic. Can you just remove it?

Bobgy:
May I ask why part of is forbidden?
Which should I use instead?

jlewi:
kustomize applies commonLabels to all selectors. As a result they need to be immutable across versions otherwise "apply" will fail because it will try to change selector labels.

This was in part because of how part-of labels were being used in previous editions of Kubeflow; i.e. part-of was referring to a string identifying the Kubeflow version.

You have a couple options

1. Set part-of labels but don't use commonLabels to do it so they don't get applied to selectors
1. Add an exception to the test (https://github.com/kubeflow/manifests/blob/58e99b277176044d4a4958c67ab42819ed6eb7e7/tests/validate_resources_test.go#L52) to allow part-of labels for KFP
1. Just remove part-of from commonLabels

Unless you have a strong reason for wanting the part-of label to be set; its probably just easiest to remove it for now.

Bobgy:
OK, I will use the name label then. Thanks

Bobgy:
/assign @jlewi 
Can you take a look again?

Bobgy:
@jlewi I changed `userid-header=x-goog-authenticated-user-email` in params.env to lowercase. KFP multi user mode seems to require it to be all lowercase. Is this correct? Do others tolerate casing?

PR #1375, opened 2020-07-10
Fix for Seldon custom namespace installs of kubeflow

### LINKED ISSUE #1022: Seldon fails if KF is deployed to a custom namespace
Kubeflow can be generally deployed to any namespace by specifying `metadata.namespace` in KFDef file.

The current version of Seldon fails though, since the `kubeflow` name is hardcoded in the `resource.yaml`.

e.g.
* https://github.com/kubeflow/manifests/blob/master/seldon/seldon-core-operator/base/resources.yaml#L3473-L3476
* https://github.com/kubeflow/manifests/blob/master/seldon/seldon-core-operator/base/resources.yaml#L3514
* https://github.com/kubeflow/manifests/blob/master/seldon/seldon-core-operator/base/resources.yaml#L3603

The above (and few more occurrences) prevent from deploying to a namespace where name != `kubeflow`.

I was able to parameterize it using Kustomize `vars`, but since the `resource.yaml` is generated it does not make sense to submit a PR for it.

I was also thinking about using my approach with vars and simple using a string `$(namespace)` in `Makefile`, but problem with that is that the namespace is templated into manifests names as well, so the `params.yaml` would be hard to maintain. Is it necessary to have `namespace` in the names of the resources - have you ever run into any conflict? 

Any thoughts what would be the right approach?

cc @cliveseldon

cliveseldon:
Resolves #1022 

 * Add kustomizations to allow namespace to be used for 
   * Certificates
   * Webhooks
   * Kubeflow gateway setting
 * Update to Seldon 1.2.1

Also need to change validation tests as CRDs can have status fields and this was being rejected.

**Checklist:**
- [*] Unit tests have been rebuilt: 
    1. `cd manifests/tests`
    3. `make test`


cliveseldon:
@vpavlin 
@yuzisun

cliveseldon:
@animeshsingh this will need your approval as well as it affects the ibm stack that imports seldon
@jlewi I think we need this for 1.1 release. A similar issue will need fixing for kfserving
@ryandawsonuk can you review from seldon side?

ryandawsonuk:
Looks good from Seldon perspective

seldon/Makefile
```diff
@@ -7,8 +7,14 @@
 
 seldon-core-operator/base: clean-kustomize
 	mkdir -p seldon-core-operator/base
-	cd seldon-core-operator/base && helm template -f ../../values.yaml seldon-core seldon-core-operator --repo https://storage.googleapis.com/seldon-charts --namespace kubeflow --version 1.1.0  > resources.yaml
+	cd seldon-core-operator/base && helm template -f ../../values.yaml seldon-core seldon-core-operator --repo https://storage.googleapis.com/seldon-charts --namespace kubeflow --version 1.2.1  > resources.yaml
+	sed -i 's#cert-manager.io/inject-ca-from:.*#cert-manager.io/inject-ca-from: $$(CERTIFICATE_NAMESPACE)/$$(CERTIFICATE_NAME)#g' seldon-core-operator/base/resources.yaml
```
adrian555, line 11: Is this supposed to run on MacOS? If yes, please fix the `sed -i` command.

cliveseldon: The present OWNERS run linux. Could we punt this to a future small PR fix? @adrian555

adrian555: k

vpavlin:
I briefly tested it and it looks good!

Thanks a lot @cliveseldon 

/lgtm

PR #1376, opened 2020-07-10
Add knative and kfserving to GCP stack

jlewi:
* Related to: kubeflow/gcp-blueprints#74

Fix common labels for knative and kfserving.

  * There were typos and they were including labels that by convention
    we currently don't want in common labels.


jlewi:
There are only 3 modified files
kfserving/installs/generic/kustomization.yaml 
knative/installs/generic/kustomization.yaml 
stacks/gcp/kustomization.yaml

PR #1381, opened 2020-07-14
knative shouldn't be part of GCP stack kustomization

jlewi:
* related to kubeflow/gcp-blueprints#74

* knative needs to be installed in its own namespace so we don't want
  to make it part of the kubeflow namespace kustomization package

* We will have to install it separately in the blueprint.


jlewi:
/approve

@Bobgy Please try to do "/approve" in addition to lgtm so that PRs can be merged if you have approval rights.

Bobgy:
I see, will do so next time.

PR #1389, opened 2020-07-15
Revert CRD status exception and remove status from Seldon CRD

### LINKED ISSUE #1386: validate_tests should not allow an exception to custom definitions to have status
#1375 added an exception to our unittest to ensure resources don't have status fields to allow customresourcedefinitions to have
status.

https://github.com/kubeflow/manifests/blob/8a9573495aba423a7048e0bc28f6df6309df4544/tests/validate_resources_test.go#L233

I don't think we want to allow status fields.

The original motivation for this test #1179 was to support git ops tools. Having status fields in resources is problematic for gitops tools because (one proposal) for the semantic are

> In general, what is declared in git is "the subset of fields and values that my objects must have set." Other fields and values can be present, but they can not conflict. Empty status is problematic because an empty status in git is essentially declaring "I always want an empty status on this resource." Beyond that, declaring a status in the source of truth is wrong from a semantic point of view.

These are the semantics expressed by ACM at least. Nonetheless, I think it is consistent with the K8s pattern that you should only be updating and specifying the spec.

It looks like the problem might be that earlier versions of kubebuilder were including CRD's in the status.

I think its better if we handle the cleanup of status in kubeflow/manifests and don't force users to have to deal with it.

The unittest is intended to capture examples that break this (e.g. the Seldon CRD) and then fix it. There are a couple approaches to fixing it

1. Application owners can choose to manually edit the CRDs or other resources and remove status
1. We can invest in writing a kustomize function to transform all resources and remove resources
    * This would be fairly straightforward https://github.com/kubeflow/kfctl/blob/master/kustomize-fns/remove-namespace/main.go

Proposal

* revert exception to custom resource definitions in #233 
* manually update the CRD to remove status

@cliveseldon does that seem reasonable?

cliveseldon:
Resolves #1386 

 * Revert exception in manifest test to again enforce status for all resources
 * Remove `status` section from Seldon CRD.

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

### DIFF
--- a/seldon/seldon-core-operator/base/resources.yaml
+++ b/seldon/seldon-core-operator/base/resources.yaml
@@ -4238,12 +4238,6 @@ spec:
   - name: v1alpha3
     served: true
     storage: false
-status:
-  acceptedNames:
-    kind: ''
-    plural: ''
-  conditions: []
-  storedVersions: []
 ---
 # Source: seldon-core-operator/templates/clusterrole_seldon-manager-role.yaml
 apiVersion: rbac.authorization.k8s.io/v1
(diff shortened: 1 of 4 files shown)


cliveseldon:
\assign @jlewi

PR #1393, opened 2020-07-17
Check in expected kpt output for Kptfile refactoring

jlewi:
* Per kubeflow/gcp-blueprints#89 we need to get rid of the legacy
  partial setters and move to using a KptFile and substitutions.

* In preparation for that we want to check in a set of test data
  that is the result of running our kpt cfg set with a given set of
  values

* This test data will be used to verify that the refactoring to use
  a KptFile doesn't change the output.

* After adding the KptFile we can simply regenerate the testdata
  and then look at the diff to ensure there are no unexpected changes.

### DIFF
--- a/hack/generate_gcp_kpt_data.py
+++ b/hack/generate_gcp_kpt_data.py
@@ -1,0 +1,61 @@
+"""Regenerate tests."""
+
+import fire
+import logging
+import os
+import shutil
+import subprocess
+
+class Generator:
+  @staticmethod
+  def write_gcp_kpt(kpt="kpt"):
+    """Create test data based on running the kpt commands.
+
+    This will allow us to see any diffs if we refactor the commands
+    """
+    repo_root = subprocess.check_output(["git", "rev-parse", "--show-toplevel"])
+    repo_root = repo_root.decode()
+    repo_root = repo_root.strip()
+
+    data_dir = os.path.join(repo_root, "tests", "testdata_gcp_kpt")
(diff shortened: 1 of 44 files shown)


tests/testdata_gcp_kpt/gcp/v2/README.md
```diff
@@ -0,0 +1,88 @@
+# Alpha: Kubeflow on KCC Installation Guide
+
+This instruction explains how to set up Kubeflow on top of Config Connector (KCC) and Anthos Service Mesh (ASM).
```
Bobgy, line 3: nit: remove this file?

hack/generate_gcp_kpt_data.py
```diff
@@ -0,0 +1,61 @@
...
+    repo_root = repo_root.decode()
+    repo_root = repo_root.strip()
+
+    data_dir = os.path.join(repo_root, "tests", "testdata_gcp_kpt")
+    if os.path.exists(data_dir):
+      shutil.rmtree(data_dir)
+
+    logging.info("Removing directory %s", data_dir)
+
+    subdirs = ["v2", "Kptfile"]
```
Bobgy, line 26: We also call kpt setters on GCP stack, shall we include it in the snapshot?

jlewi: Lets do it in a follow on PR just so we can get this merged.
My initial PR isn't converting the setters in stacks.

PR #1398, opened 2020-07-20
Convert v1 to v2 setters & substituions in gcp

jlewi:
* The latest version of kpt started choking on gcp/v2 because we were
  still using the old style setters and substitutions.

* This PR creates a kptfile to use the new setter and substitutions.

* hack/create_kptfile.py contains a script to generate lot of the setters
  and substitutions.

* kf-vm-sa.yaml shouldn't specify the namespace; this will get set in an overlay

* Move workload identity bindings for kf-admin KSA from kubeflow/instance in gcp blueprints repo into this repository.

related to gcp-blueprints#89


jlewi:
/assign @Bobgy 

If you look at the diff for testdata_gcp_kpt you can see that the files are mostly unchanged indicating that the substitutions match what we had before.

The exception is one or two bug fixes; e.g.

* There are a couple places where I  we were using zone when we should have used location
* I created separate setters for the management cluster whereas before they were sharing the same setters as the non management cluster.

jlewi:
/assign @Ark-kun 

@Ark-kun Could you review this since its 10PM in shanghai?

Bobgy:
/lgtm
/approve

Approving first, going to take a deeper look

hack/create_kptfile.py
```diff
@@ -0,0 +1,360 @@
+"""This is a script created for the updating of the GCP Kpt packages.
+
+The purpose of this is to upgrade to using setters and substitutions
+"""
```
Bobgy, line 4: Do we just use this for upgrade?
How should users create new substitutions when working with blueprint?

jlewi: This is a onetime script. We can delete it in a subsequent PR. If we need to create additional substitutions we can do so using the kpt CLI.

When doing these types of conversions its just a good practice IMO to check in a script so we have a record in case we need to redo it.

PR #1403, opened 2020-07-22
Update ObjectStoreConfig.PipelineFolder in api-server configmap

slenky:
This PR adds a missing parameter in manifests/apiserver which was added by kubeflow/pipelines#2080 

This parameter is crucial for Kubeflow 1.0.2 deployments as it breaks the Pipeline UI functionality: kubeflow/pipelines#4254

Resolves kubeflow/pipelines#4254

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

### DIFF
--- a/pipeline/api-service/base/config-map.yaml
+++ b/pipeline/api-service/base/config-map.yaml
@@ -15,6 +15,7 @@ data:
         "AccessKey": "minio",
         "SecretAccessKey": "minio123",
         "BucketName": "mlpipeline",
+        "PipelineFolder": "pipelines",
         "Secure": false
       },
       "InitConnectionTimeout": "6m",


pipeline/api-service/base/config-map.yaml
```diff
@@ -15,6 +15,7 @@ data:
         "AccessKey": "minio",
         "SecretAccessKey": "minio123",
         "BucketName": "mlpipeline",
+        "PipelineFolder": "pipelines",
```
Bobgy, line 18: nit: in latest code, it was renamed to https://github.com/kubeflow/pipelines/blob/ac66626a2c4c0222158817ac3c920b090af4245b/backend/src/apiserver/client_manager.go#L339, I think the `PipelineFolder` was never released.

Bobgy:
Did you verify if the new config can solve the issue for you?

pipeline/api-service/base/config-map.yaml
slenky: @Bobgy version 0.2.5 explicitly uses `PipelineFolder` https://github.com/kubeflow/pipelines/blob/0.2.5/backend/src/apiserver/client_manager.go#L331

slenky:
> Did you verify if the new config can solve the issue for you?

Yes, adding that `PipelineFolder` fixes an issue for Kubeflow 1.0.2 so I reckon it makes sense to cherry-pick that commit only to 1.0

pipeline/api-service/base/config-map.yaml
Bobgy: Ohhh, that's a miserable mistake...
I thought their fix went into the same release.

I don't think we should use a release with a wrong env var, that would further confuse next upgrade.
Shall we bump to 0.3.0 instead or directly try to integrate KFP 1.0.0 in https://github.com/kubeflow/manifests/tree/master/pipeline/installs/generic.

Bobgy: hmmm, but the cost might be too high for just fixing this issue.
I think this is fine.

When people upgrade, they should have the config overriden automatically.

Bobgy:
I think you also need to update test snapshot

>  Unit tests have been rebuilt:
cd manifests/tests
make generate-changed-only
make test

PR #1426, opened 2020-07-26
Migrate istio and dex to V3

### LINKED ISSUE #1424: Application v3 manifest - Istio and Dex
Ref: https://github.com/kubeflow/manifests/issues/1062

I wonder if we'll migrate Istio and Dex components to V3. It will be great for us to have at least a `base_v3` directory under the components folder.

By following below principals to establish base_v3 without changing original base+overlay folder:

```
Applications in KFDef should not use parameters field
Applications should fetch global paramaters from the configmap "kubeflow-config"
Applications should fetch application parameters from application specific config maps
Application specific config maps should be named "${APPNAME}-config"
Use of vars should be minimized
The most common and supported use case should be to substitute parameters into
fields that don't support referencing a configmap directly
In this case var names should be unique enough that they won't conflict with vars from
other applications
```

Since we AWS side has dependencies on Istio and Dex, there's motivation for us to help with migration.

PatrickXYS:
Resolves #1424 

1. Migrate Istio, Istio-1-3-1, and Dex applications to V3.
2. Adopt corresponding changes for AWS.

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`


PatrickXYS:
/cc @Jeffwan Can you help review AWS changes?
/cc @krishnadurai @yanniszark Can you help review Istio + Dex changes?

Btw, let me know if you think it's necessary to divide Istio+Dex migration and AWS changes into 2 PR.

Jeffwan:
@PatrickXYS I would suggest to split AWS changes and common changes to separate PR. then it would be easier to track or cherry-pick in the future

krishnadurai:
I agree with @Jeffwan's suggestion of splitting the PR into 2.

PatrickXYS:
@krishnadurai I have removed AWS change and will send another PR after this PR merged.

Can you help review the PR?

PatrickXYS:
Just for ping @krishnadurai 

I did not change any code related to istio / dex components. Only change made here is removing AWS changes.

Would you minding take another look? And I only created `base_v3` folder, this will be no impact on existing components/depencies.

istio/oidc-authservice/base_v3/kustomization.yaml
```diff
@@ -0,0 +1,84 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- ../base/service.yaml
+- ../base/statefulset.yaml
+- ../base/envoy-filter.yaml
+- ../base/pvc.yaml
```
Jeffwan, line 7: probably overlay/application.yaml is needed as well. I think besides `param -> config`, there's no need to migrate to v3 for this one. Can user use `overlay/applications.yaml` directly?

istio/istio/base_v3/kustomization.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- ../base/kf-istio-resources.yaml
```
Jeffwan, line 4: This is same as istio/istio/base/kustomize?

istio/oidc-authservice/base_v3/kustomization.yaml
PatrickXYS: > param -> config

Yes this is the main reason I want to migrate to V3. 

> Can user use overlay/applications.yaml directly?

Yes they can.

I'm okay with not migrating these components to V3. I think it might cause some incompatibility issues in the future. Since vendors (AWS/GCP) have moved to V3, and use `config` as the routine. 

For example, https://github.com/kubeflow/manifests/blob/master/stacks/aws/application/istio/kustomization.yaml#L7-L10

if we don't migrate to V3, these configMap will be `XX-parameters`, what do you think?

```
- name: istio-parameters
  behavior: merge
  envs:
  - params.env
```

istio/istio/base_v3/kustomization.yaml
PatrickXYS: Right, they are the same.

PatrickXYS:
@terrytangyuan @animeshsingh @Bobgy Can you take a look? 

It's now lgtm but need to be approved for shipment.

karlschriek:
@PatrickXYS  was there ever a separate PR for the AWS part? I don't see it linked anywhere and as far as I can tell the official aws kfdef still installs 1.1.6.

Looking at the rolled-back changes in this PR, it looks like bumping to 1.3.1 only requires replacing 

```
    - kustomizeConfig:
        repoRef:
          name: manifests
          path: stacks/aws/application/cluster-local-gateway
      name: cluster-local-gateway
    - kustomizeConfig:
        repoRef:
          name: manifests
          path: stacks/aws/application/istio
        name: istio

```

with 

```
    - kustomizeConfig:
        repoRef:
          name: manifests
          path: istio/cluster-local-gateway/base_v3
      name: cluster-local-gateway
    - kustomizeConfig:
        repoRef:
          name: manifests
          path: stacks/aws/application/cognito-istio
      name: istio
```

Is that correct?

karlschriek:
No, can't be right, not in 1.2.0 at least. `stacks/aws/application/cognito-istio` doesn't exist

PR #1431, opened 2020-07-29
[pending] Uses standard header "Authorization" instead of "X-Auth-Token" for Kubeflow with oidc-authservice + dex

shawnzhu:
…I clients like kfp SDK.

Supports https://github.com/kubeflow/kubeflow/issues/4912#issuecomment-665387957

When using KFP SDK to access pipelines via public endpoint, the `existing_token` argument will be sent over request header `Authorization` which will NOT be accepted without this change. since it already used the standard header name `Authorization` in `arrikto/oidc-authservice` from https://github.com/arrikto/oidc-authservice/pull/25, it should pick it up for kubeflow that depends on it.

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

### DIFF
--- a/istio/oidc-authservice/base/envoy-filter.yaml
+++ b/istio/oidc-authservice/base/envoy-filter.yaml
@@ -17,7 +17,7 @@ spec:
           allowedHeaders:
             patterns:
             - exact: "cookie"
-            - exact: "X-Auth-Token"
+            - exact: "authorization"
         authorizationResponse:
           allowedUpstreamHeaders:
             patterns:
--- a/istio/oidc-authservice/base/statefulset.yaml
+++ b/istio/oidc-authservice/base/statefulset.yaml
@@ -17,7 +17,7 @@ spec:
     spec:
       containers:
       - name: authservice
-        image: gcr.io/arrikto/kubeflow/oidc-authservice:6ac9400
+        image: gcr.io/arrikto/kubeflow/oidc-authservice:fef11c3
         imagePullPolicy: Always
         ports:
         - name: http-api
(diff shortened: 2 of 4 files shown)


swiftdiaries:
Hey @shawnzhu 
How would downstream applications like the jupyter-web-app handle this?
Would it break client apps already checking this explicitly?
Example: https://github.com/kubeflow/kubeflow/blob/master/components/jupyter-web-app/backend/kubeflow_jupyter/common/base_app.py#L14

shawnzhu:
@swiftdiaries 

> How would downstream applications like the jupyter-web-app handle this?

That's a good call. I noticed that the header `X-Auth-Token` is used in two places:

https://github.com/kubeflow/kubeflow/blob/dcaeee742984533a2aff34bb28e14ea5aa9ae74a/components/jupyter-web-app/frontend/src/app/uis/rok/services/rok.service.ts#L46

and:

https://github.com/kubeflow/kubeflow/blob/523c02b6508761aedc5913116ac6afc3dc170b14/components/crud-web-apps/common/frontend/kubeflow-common-lib/projects/kubeflow/src/lib/services/rok/rok.service.ts#L70

> Would it break client apps already checking this explicitly?

My idea is deprecating the header `X-Auth-Token` by sending token via both `X-Auth-Token` and `Authorization` headers to support backward compatibility prior to this PR.

swiftdiaries:
Sounds good, thanks @shawnzhu !

shawnzhu:
This PR should not be merged until seeing progress from https://github.com/kubeflow/kubeflow/pull/5303

yanniszark:
@shawnzhu thanks for your PR!
Please keep in mind that quite a few things have changed in the AuthService since it was last updated in Kubeflow.

The things that I believe will break are:
* The logout mechanism has changed to support CSRF protection. As such, the CentralDashboard should be updated to support the new mechanism. cc @lalithvaka who implemented it (https://github.com/kubeflow/kubeflow/pull/5174)
* The settings have changed a bit, so you need to make sure that they are up-to-date.

Please let me know what you think!

Jeffwan:
@shawnzhu Any progress on this PR? We plan to cut release end of next week. Please let me know if you want to include it in v1.2

DavidSpek:
@shawnzhu Is this PR going to move forward for release 1.3?

shawnzhu:
> Is this PR going to move forward for release 1.3?

I hope so.

Issue #1450, opened 2020-08-01
cert-manager namespace missing control plane label


jlewi:
cert-manager namespace is missing the control plane label
https://github.com/kubeflow/manifests/blob/011c81ec2be8875d8032b79b62340ab924258230/namespaces/base/namespaces.yaml#L12

This can cause deadlock because cert-manager pods might not get deployed because the KFServing admission webhooks aren't valid because they can't get a certificate because cert-manager isn't running.

I think we haven't observed this because we deploy cert-manager first.

However, when deploying with GitOps (ACM) I hit this problem.

We should cherry-pick the fix onto the 1.1 branch before closing this issue.

Jeffwan:
Can be closed since PR is merged

PR #1451, opened 2020-08-01
cert-manager namespace needs label control-plane

jlewi:
* cert-manager namespace needs the control-plane label to prevent
  application of kfserving webhooks.

related to: #1450

### DIFF
--- a/namespaces/base/namespaces.yaml
+++ b/namespaces/base/namespaces.yaml
@@ -10,3 +10,4 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: cert-manager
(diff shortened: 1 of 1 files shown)


namespaces/base/namespaces.yaml
```diff
@@ -10,3 +10,4 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: cert-manager
+  control-plane: kubeflow
```
Bobgy, line 13: Did you forget labels:?

jlewi: I did thank you.

PR #1487, opened 2020-08-13
Create v1.1.0 Azure Kfdef manifests

eedorenko:
- KFDef manifest for v1.1 to install on Azure
- Anonymous envoy filter for Istio-1.6.x


eedorenko:
@berndverst FYI

berndverst:
/assign @yanniszark

This is fine from an Azure perspective

argo/base_v3/kustomization.yaml
```diff
@@ -17,10 +17,10 @@ commonLabels:
 images:
 - name: argoproj/argoui
   newName: argoproj/argoui
-  newTag: v2.3.0
+  newTag: v2.4.3
 - name: argoproj/workflow-controller
   newName: argoproj/workflow-controller
-  newTag: v2.3.0
+  newTag: v2.7.5
```
yanniszark, line 23: Is this change necessary for Azure?
I'm a bit reluctant to make this change in this PR.
How is this going to affect Kubeflow Pipelines users? Is Pipelines compatible with this Argo version?
I think we should split this into a separate PR and validate compatibility.

yanniszark (changes requested):
Hi @eedorenko! The PR overall looks good, only a minor change.
I would suggest also the following:
- Add the people responsible for manifests on Azure's side to `stacks/azure/OWNERS` and `kfdef/OWNERS`.
- Update the website instructions accordingly, if necessary.

argo/base_v3/kustomization.yaml
eedorenko: Ok. I took it out. I will create a separate PR for that

eedorenko:
>Add the people responsible for manifests on Azure's side to stacks/azure/OWNERS and kfdef/OWNERS.
Yes, once they are approved as contributors

>Update the website instructions accordingly, if necessary.
Yep. On our list. Once this PR is merged.

eedorenko:
@yanniszark done!

eedorenko:
The PR to add Azure owners https://github.com/kubeflow/manifests/pull/1524

PR #1488, opened 2020-08-13
fix: update kfp profile controller's version

Bobgy:

Fixes https://github.com/kubeflow/pipelines/issues/4368

Resolves problem:
> Right now, KFP profile controller deploys visualization server and artifact fetcher servers of 1.0.0-rc.3 version, but they should be deploying 1.0.0 now.

There were some tfma/tfdv visualization exceptions fixed between 1.0.0-rc.3 and 1.0.0, so we need to cherry pick this fix.

Use KFP version from upstream, so it doesn't need to get bumped each time.

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`


tests/stacks/ibm/multi-user/test_data/expected/apps_v1_deployment_kubeflow-pipelines-profile-controller.yaml
```diff
@@ -28,6 +28,11 @@ spec:
         - python
         - /hooks/sync.py
         env:
+        - name: KFP_VERSION
+          valueFrom:
+            configMapKeyRef:
+              key: appVersion
+              name: pipeline-upstream-install-config-c958fm776d
```
Bobgy, line 35: Because of the kustomize issue: https://github.com/kubernetes-sigs/kustomize/issues/2609.
We need to stop using commonPrefix to make this configMapKeyRef correctly find pipeline-upstream-install-config and append the suffix.

pipeline/installs/multi-user/pipelines-profile-controller/deployment.yaml
```diff
@@ -15,8 +15,13 @@ spec:
...
         envFrom:
         - configMapRef:
-            name: profile-controller-env
+            name: kubeflow-pipelines-profile-controller-env
         env:
+        - name: KFP_VERSION
+          valueFrom:
+            configMapKeyRef:
+              name: pipeline-upstream-install-config
+              key: appVersion
```
Bobgy, line 24: This is the fix

PR #1513, opened 2020-08-25
Remove http from the GCP ingress and explicitly set the default backend.

jlewi:
* Ref https://cloud.google.com/kubernetes-engine/docs/concepts/ingress-xlb#disabling_http force all traffic from client to ingress to be https.

* Fix kubeflow/gcp-blueprints#116

* Explicitly set the default backend to the ISTIO ingress gateway.


Bobgy:
/lgtm
/approve
Thanks!

Issue #415, opened 2019-10-02
Remove tensorboard manifest and from KFDef


Snapple49:
Hi!

Was poking around in the generated kubeflow installation based on the `kfctl_k8s_istio.yaml` manifest, which, in the version I have, by default contains a Tensorboard deployment. When I asked about it @jlewi pointed out that it should not be there by default and asked me to post an issue about it. I can't find a version number for it, I don't remember what version of kfctl I was using at the time but I downloaded the kfctl manifest at least before September 14, 22:25 if that helps!

jlewi:
We should remove tensorboard from our KFDef's
https://github.com/kubeflow/manifests/blob/67eabbfd907dbb212b2fa39baeddb3a4b7e9e743/kfdef/kfctl_gcp_iap.yaml#L193

And remove the corresponding manifest
https://github.com/kubeflow/manifests/tree/master/tensorboard

This is just causing confusion. All its doing is deploying a single instance of tensorboard. This is harmless but useless and confusing.

If people want to run tensorboard the correct thing to do would be for them to create a Deployment and VirtualService. There's not much point to us providing a kustomize manifest for this.

At best we should publish an example YAML file containing the resources (e.g. in the docs).

Jeffwan:
@jlewi I can help on this. BTW, do we want  `tensorboard-controller` ? I think that is helpful

jlewi:
Thanks @Jeffwan 

I think the tensorboard controller is still a prototype.

kubeflow/kubeflow#3578 is tracking building a web app to manage tensorboard instances. Creating a manifest to deploy the tensorboard-controller would probably be a part of that work; assuming the web app ended up using the controller.

/assign @Jeffwan

Jeffwan:
@jlewi Get it. I will file a PR to remove tensorboard in kfdef

PR #1515, opened 2020-08-25
Add missing files for k8s_istio.v1.1.0 KFDef

pvaneck:
Running through install instructions for using the `kfctl_k8s_istio.v1.1.0.yaml` KFDef file, I was running into issues with some manifests missing. This PR creates those missing files, following the current paradigm of having these exist in the `stacks/kubernetes/application` directory. Once I added these, Kubeflow deployed successfully.

Also removed Tensorboard from the KFDef. (see https://github.com/kubeflow/manifests/issues/415)

Can generate tests if these changes make sense.

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

### DIFF
--- a/kfdef/kfctl_k8s_istio.v1.1.0.yaml
+++ b/kfdef/kfctl_k8s_istio.v1.1.0.yaml
@@ -86,11 +86,6 @@ spec:
         name: manifests
         path: stacks/kubernetes/application/spartakus
     name: spartakus
-  - kustomizeConfig:
-      repoRef:
-        name: manifests
-        path: stacks/kubernetes/application/tensorboard
-    name: tensorboard
   repos:
   - name: manifests
     uri: https://github.com/kubeflow/manifests/archive/v1.1-branch.tar.gz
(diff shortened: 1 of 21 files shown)


PatrickXYS:
It makes sense to me, can you generate tests?  

@pvaneck

kfdef/kfctl_k8s_istio.v1.1.0.yaml
```diff
@@ -86,11 +86,6 @@ spec:
         name: manifests
         path: stacks/kubernetes/application/spartakus
     name: spartakus
-  - kustomizeConfig:
```
PatrickXYS, line 89: If we remove tensorboard from manifests, can you also remove folders `stacks/kubernetes/application/tensorboard`?

pvaneck: `stacks/kubernetes/application/tensorboard` never seemed to have existed, so we should be good there. :)

krishnadurai (approved):
/lgtm
/approve

PR #1541, opened 2020-09-03
Kubernetes API Destination Rule

maganaluis:
This commit adds an Istio Destination Rule for the Kubernetes API to ensure that any client calls to the API are consistent across all platforms.

Resolves # https://github.com/kubeflow/pipelines/issues/4407

This change creates a destination rule for the Kubernetes API, which will be applied when platforms use the istio/istio/base_v3 or as part of a patch. 

This rule ensure that calls to the Kubernetes API are consistent across all platforms, as mentioned on my issue the Kubeflow Pipelines and the Jupyter Web App were timing out on Azure. 

After applying this rule I was able to test both with the Kubernetes Python Client and the Golang Client that calls the API were successful regardless of the idle time. In the case of Azure it will result on a timeout after a 5 minute idle time. 

Resources:

https://preliminary.istio.io/latest/docs/reference/config/networking/destination-rule/#ConnectionPoolSettings
https://github.com/golang/go/blob/release-branch.go1.10/src/net/http/transport.go#L40-L51

Jobs to replicate issue and test fix: 

- https://github.com/maganaluis/k8s-api-python
- https://github.com/maganaluis/k8s-api-golang

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

### DIFF
--- a/istio/istio/base_v3/destination-rules.yaml
+++ b/istio/istio/base_v3/destination-rules.yaml
@@ -1,0 +1,13 @@
+apiVersion: networking.istio.io/v1alpha3
+kind: DestinationRule
+metadata:
+  name: kubernetes-api
+spec:
+  host: "kubernetes.default.svc.cluster.local"
+  trafficPolicy:
+    connectionPool:
+      tcp:
+        connectTimeout: 10s
+        tcpKeepalive:
+          time: 75s
+          interval: 75s
--- a/istio/istio/base_v3/kustomization.yaml
+++ b/istio/istio/base_v3/kustomization.yaml
@@ -3,6 +3,7 @@ kind: Kustomization
 resources:
 - ../base/kf-istio-resources.yaml
 - ../base/cluster-roles.yaml
+- destination-rules.yaml
 namespace: kubeflow
 generatorOptions:
   disableNameSuffixHash: true


istio/istio/base/istio-rules.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: networking.istio.io/v1alpha3
```
PatrickXYS, line 1: Since we have migrated to manifest v3, for most of the cases, we won't suggest adding files into base folder but base_v3 folder. 

Because we kind-of "freeze" all components' base directory and treat them as read-only.

It's okay if istio folder OWNERS think we can make the change.

krishnadurai:
@PatrickXYS yes I'll take a look this Monday

istio/istio/base/istio-rules.yaml
krishnadurai: @maganaluis can you please accommodate this change and instead make this in base_v3?

krishnadurai, line 1: The convention is to name this file as `destination-rules.yaml`.

krishnadurai (changes requested):
@maganaluis thanks for this PR. Can you also see that this change is made for istio 1.3.1? We have 2 versions of istio maintained in Kubeflow.

istio/istio/base/istio-rules.yaml
maganaluis: Done

maganaluis: Removed any references on the base folder.

maganaluis:
@krishnadurai I removed the references to the base folder, and renamed the file to follow naming convention. I ran the test cases as well locally, but there was nothing generated related to my change... not sure if this has changed but I follow these steps: 

https://github.com/kubeflow/manifests/blob/master/tests/README.md

maganaluis:
> @maganaluis thanks for this PR. Can you also see that this change is made for istio 1.3.1? We have 2 versions of istio maintained in Kubeflow.

Forgot to mention that the rule will work from Istio 1.1 all the way to the latest version, so the Kubeflow supported versions should be included in this range. We've been testing it with Istio 1.6.10

PatrickXYS:
/lgtm

Can we move forward?

PatrickXYS:
@maganaluis Can you help rebase master and to merge?

maganaluis:
@PatrickXYS The repository doesn't exist anymore. However there is no conflicts as far as I can see, the last change on the Istio folder was on Aug 6

maganaluis:
Closing this PR due to GitHub bug https://github.com/isaacs/github/issues/168

PR #1562, opened 2020-09-17
fix mysql deployment strategy

thesuperzapper:
When trying to restart MySQL instances, they get stuck because the old container will refuse to give up the lock on the DB:
* `kubectl rollout restart deployment/katib-mysql  -n kubeflow`
* `kubectl rollout restart deployment/metadata-db  -n kubeflow`

This PR fixes that by setting the deployment strategy to `Recreate`.

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

### DIFF
--- a/metadata/overlays/db/metadata-db-deployment.yaml
+++ b/metadata/overlays/db/metadata-db-deployment.yaml
@@ -5,10 +5,12 @@ metadata:
   labels:
     component: db
 spec:
+  replicas: 1
   selector:
     matchLabels:
       component: db
-  replicas: 1
+  strategy:
+    type: Recreate
   template:
     metadata:
       name: db
--- a/katib/katib-controller/base/katib-mysql-deployment.yaml
+++ b/katib/katib-controller/base/katib-mysql-deployment.yaml
@@ -11,6 +11,8 @@ spec:
     matchLabels:
       app: katib
       component: mysql
+  strategy:
+    type: Recreate
   template:
     metadata:
       name: katib-mysql
--- a/tests/katib/installs/katib-standalone-ibm/test_data/expected/apps_v1_deployment_katib-mysql.yaml
+++ b/tests/katib/installs/katib-standalone-ibm/test_data/expected/apps_v1_deployment_katib-mysql.yaml
@@ -16,6 +16,8 @@ spec:
       app.kubernetes.io/component: katib
       app.kubernetes.io/name: katib-controller
       component: mysql
+  strategy:
+    type: Recreate
   template:
     metadata:
       annotations:
--- a/tests/katib/installs/katib-standalone/test_data/expected/apps_v1_deployment_katib-mysql.yaml
+++ b/tests/katib/installs/katib-standalone/test_data/expected/apps_v1_deployment_katib-mysql.yaml
@@ -16,6 +16,8 @@ spec:
       app.kubernetes.io/component: katib
       app.kubernetes.io/name: katib-controller
       component: mysql
+  strategy:
+    type: Recreate
   template:
     metadata:
       annotations:
--- a/tests/stacks/aws/test_data/expected/apps_v1_deployment_katib-mysql.yaml
+++ b/tests/stacks/aws/test_data/expected/apps_v1_deployment_katib-mysql.yaml
@@ -16,6 +16,8 @@ spec:
       app.kubernetes.io/component: katib
       app.kubernetes.io/name: katib-controller
       component: mysql
+  strategy:
+    type: Recreate
   template:
     metadata:
       annotations:
--- a/tests/stacks/aws/test_data/expected/apps_v1_deployment_metadata-db.yaml
+++ b/tests/stacks/aws/test_data/expected/apps_v1_deployment_metadata-db.yaml
@@ -12,6 +12,8 @@ spec:
     matchLabels:
       component: db
       kustomize.component: metadata
+  strategy:
+    type: Recreate
   template:
     metadata:
       annotations:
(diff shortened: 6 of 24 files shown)


thesuperzapper:
/assign @andreyvelich @zhenghuiwang

can you please review?

andreyvelich:
@thesuperzapper Thank you for updating this, have you checked that when `katib-mysql` pod is restarted `katib-db-manager` connects to the new DB properly ?

thesuperzapper:
@andreyvelich well as far as I can see it works

andreyvelich:
@thesuperzapper Thank!
Can you open a PR to make changes in Katib [manifests](https://github.com/kubeflow/katib/blob/master/manifests/v1beta1/mysql-db/deployment.yaml) also, please?

andreyvelich (commented):
/lgtm

thesuperzapper:
Gentle ping: @gaocegege @johnugeorge

thesuperzapper:
Have you had a chance to review this? @andreyvelich @zhenghuiwang

andreyvelich (approved):
@thesuperzapper Sorry for the long reply!
/lgtm
/approve

thesuperzapper:
I still need one of the metadata owners to LGTM this

Can one of @neuromage @prodonjs @zhenghuiwang please take a look?

PatrickXYS:
Can you try re-run new test?

/test ?

thesuperzapper:
I am still waiting on one of @neuromage  @prodonjs  @zhenghuiwang to LGTM this

I want to get this in before Kubeflow 1.2

PatrickXYS:
@thesuperzapper I don't think any of them are active in kubeflow community now.

You might need to find review from other one, after that, try re-run test because we have migrated to shared test-infra

metadata/overlays/db/metadata-db-deployment.yaml
```diff
@@ -5,10 +5,12 @@ metadata:
   labels:
     component: db
 spec:
+  replicas: 1
   selector:
     matchLabels:
       component: db
-  replicas: 1
+  strategy:
```
Jeffwan, line 12: em. Technically, we should use `statefulset` for mysql. because pod under deployment is controlled by ReplicaSet and replacement will be created immediately if you do an upgrade.  

This workaround has some down side, existing pods will be killed before new ones created. The problem is these are not all DB layers. `Recreate` strategy may cause downtime to its dependencies like KFP or Katib components. 
It requires this layer has the resilience on db failure. 

I have not tested this and not sure if anything abnoarmal there.

thesuperzapper: @Jeffwan if you see this PR's description, currently, if you try and restart the MySQL Deployment, it will hang forever.

This is because the old Pod wont give up the lock on the DB files, so the new one can never be "ready" unless you manually kill the old Pod.

Jeffwan: I saw the problem.. My asking is if you see abnormal on katib or KFP side when you did upgrade with `Recreate ` strategy because DB service will have down time. connection maybe lost and if application doesn't handle this well, it could be an issue..

thesuperzapper: @Jeffwan, we don't really have a choice, there will have to be some downtime as only one MySQL instance can lock the DB files at a time.

We just need to design Katib / KFP to be able to handle short downtimes of MySQL (like using exponential-backoff retries)

Jeffwan: sounds good

andreyvelich:
@thesuperzapper Can you try to rebase, to fix the tests?

thesuperzapper:
@andreyvelich @Jeffwan I have rebased

andreyvelich:
@thesuperzapper I think you need to run `make generate-changed-only` again to fix unit tests.

thesuperzapper:
@andreyvelich when I run `make generate-changed-only` on a rebased version of my branch, I get an error, is master broken? Or is it something to do with my setup?

PatrickXYS:
@thesuperzapper If there's something wrong in unit test, you might need to go figure it out, keeping unit test succeed is the bottom line.

Jeffwan:
@thesuperzapper Can you share the errors? What's the errors msg? What kustomize version you have in your environment. Recent change all pass UT and I think master should be good.

thesuperzapper:
@Jeffwan it seems like kustomize 3.8.1 is not working with our repo, we probably need to fix this for users

I have rolled my local version back to 3.6.1 and `make generate-changed-only` works again

PatrickXYS:
Copy from manifest README

```
This repo contains kustomize packages for deploying Kubeflow applications.

If you are a contributor authoring or editing the packages please see Best Practices. Note, please use kustomize v3.2.1 with manifests in this repo, before #538 is fixed which will allow latest kustomize to be used.
```

https://github.com/kubeflow/manifests

thesuperzapper:
@PatrickXYS, yeah, but we should aim to support any widely used versions of kustomize

PatrickXYS:
@thesuperzapper Ideally we should do so, let's create a GitHub issue to talk about it.

/lgtm
/approve
/hold

Can you cherry-pick this PR into v1.2-branch as well?

Jeffwan:
this should be good to go.

/hold cancel

thesuperzapper:
@PatrickXYS, I dont think I have the required permissions to chery-pick this commit

PatrickXYS:
@thesuperzapper You won't be able to write directly to v1.2-branch, but send a cherry-pick PR

PR #1563, opened 2020-09-17
ASM Mesh id should be proj-${PROJECTNUBMER}

jlewi:
* Fix kubeflow/gcp-blueprints#117
* The pattern for the mesh id needs to be proj-${PROJECTNUMBER}
* Dedube the mesh substitutions; there were two different substitutions "mesh-id" and "asm-mesh-id" for the mesh id.

### DIFF
--- a/gcp/Kptfile
+++ b/gcp/Kptfile
@@ -697,6 +697,7 @@ openAPI:
         setter:
           name: gcloud.project.projectNumber
           value: "147474701642"
+          isSet: true
     io.k8s.cli.substitutions.asm-gcp-metadata:
       x-k8s-cli:
         substitution:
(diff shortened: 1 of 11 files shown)


gcp/v2/asm/istio-operator.yaml
```diff
@@ -25,14 +25,14 @@ spec:
       istio-ingressgateway:
         type: NodePort
     global:
-      meshID: "project-id_us-central1_name" # {"$kpt-set":"asm-mesh-id"}
+      meshID: "proj-147474701642" # {"$kpt-set":"asm-mesh-id"}
```
Bobgy, line 28: You probably forgot to reset the id?

jlewi: Good catch. I updated the gcp-reset make command and reran it. A bunch of other values got reset as well.

gcp/Kptfile
```diff
@@ -20,17 +20,17 @@ openAPI:
...
         setter:
           name: gcloud.compute.region
-          value: us-central1
+          value: REGION
     io.k8s.cli.setters.location:
       x-k8s-cli:
         setter:
           name: location
-          value: us-east1-d
+          value: location
```
Bobgy, line 33: nit: shall we make the default value's casing consistent?

Bobgy (commented):
Thank you!
/LGTM
/Approve

One nitpicking, but I think we can first get this in.

PR #1566, opened 2020-09-22
disable istio-sidecar for pipelines mysql

thesuperzapper:
I am not 100% sure this is correct, but I believe the pipelines MySQL pod does not need an istio-sidecar.

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

### DIFF
--- a/pipeline/mysql/installs/generic/deployment-patch.yaml
+++ b/pipeline/mysql/installs/generic/deployment-patch.yaml
@@ -1,0 +1,9 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: mysql
+spec:
+  template:
+    metadata:
+      annotations:
+        sidecar.istio.io/inject: "false"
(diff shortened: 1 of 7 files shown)


thesuperzapper:
@Bobgy @IronPan @rmgogogo  can one of you please review?

I am fairly sure this is a sensible change, because the [base mysql disables istio-sidecar injection](https://github.com/kubeflow/manifests/blob/master/pipeline/mysql/base/deployment.yaml#L11), but we have omitted it for [./pipeline/installs/generic](https://github.com/kubeflow/manifests/tree/master/pipeline/installs/generic).

But I might be wrong.

Bobgy:
Mysql is a sensitive data store in KFP multi-user mode, so we need istio sidecar there to enforce access control.
There are configs in https://github.com/kubeflow/manifests/blob/d51b77c8caf0c417d259d79878da962644800036/pipeline/installs/multi-user/istio-authorization-config.yaml#L22.

The base mysql is only left behind the same as Kubeflow 1.0, we can delete it now.

thesuperzapper:
@Bobgy cool I will close this

I am trying to resolve https://github.com/kubeflow/pipelines/issues/4530

Issue #1543, opened 2020-09-03
kubeflow 1.1.0 install failed when use 1.18 or higher kubernetes


whalecold:
The seldon CRD [APIV3Schema](https://github.com/kubeflow/manifests/blob/ed615418ef15e54c9a68979d8d2fc45e17bab7b1/seldon/seldon-core-operator/base/resources.yaml#L2490) is invalid in the kubernetes of 1.18，Could we set the `protocol`  field to `Required` ? Alternative, using the v1 CRD API.

Ref: https://github.com/kubernetes/kubernetes/issues/91395#issuecomment-633261523

tomalbrecht:
Same here on k3s using https://github.com/kubeflow/manifests/blob/master/kfdef/kfctl_k8s_istio.v1.1.0.yaml

```bash
k3s --version
k3s version v1.18.8+k3s1 (6b595318)
```

```bash
...
servicerolebinding.rbac.istio.io/bind-cache-server-admission-webhook unchanged
servicerolebinding.rbac.istio.io/bind-gateway-ml-pipeline-ui unchanged
servicerolebinding.rbac.istio.io/bind-ml-pipeline-internal unchanged
persistentvolumeclaim/katib-mysql unchanged
persistentvolumeclaim/metadata-mysql unchanged
persistentvolumeclaim/minio-pvc unchanged
persistentvolumeclaim/mysql-pv-claim unchanged
validatingwebhookconfiguration.admissionregistration.k8s.io/seldon-validating-webhook-configuration-kubeflow configured
[...]

PR #1567, opened 2020-09-22
Adding Openshift stack

nakfour:
This is to upgrade all changes needed to run Kubeflow on OCP 4.x to kustomize v3.
Resolves #

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`


crobby (approved):

crobby (commented):
Looks like its on the right path.  Great work!

vpavlin:
Metadata fail to start because metadata-db livenessprobe is failing. The liveness probe needs to use `MYSQL_USER` to check mysql availability. You can use my commit here:

https://github.com/vpavlin/manifests/commit/d9bce160281988f8e498a96a8f9633e479670b6a

nakfour:
@vpavlin now I get this error below in on of the metadata-db pods that is showing up "not ready" then keeps crashing
```
[ERROR] InnoDB: Unable to lock ./ibdata1 error: 11
```

nakfour:
@jlewi @animeshsingh is there a way to get detailed logs from the kubeflow-manifests-presubmit job? Its failing and we have no indication why?

Jeffwan:
@nakfour 

Did you run test locally? 

seems Argo component doesn't load logs correctly. 

http://testing-argo.kubeflow.org/workflows/kubeflow-test-infra/kubeflow-manifests-presubmit-unit-1567-02bd38e-1376-0d7e?tab=workflow&nodeId=kubeflow-manifests-presubmit-unit-1567-02bd38e-1376-0d7e-202129096&sidePanel=logs%3Akubeflow-manifests-presubmit-unit-1567-02bd38e-1376-0d7e-202129096%3Amain

Jeffwan:
If you plan to have this in v1.2 release, please let us know https://github.com/kubeflow/kubeflow/issues/5224

nakfour:
@Jeffwan yes ran test manually and ran into issues documented here : https://github.com/kubeflow/manifests/issues/1596
I do not see the argo issue you are referring to, please point it out?

kfdef/kfctl_openshift.v1.1.0.yaml
```diff
@@ -0,0 +1,103 @@
...
+metadata:
+  name: kubeflow
+  namespace: kubeflow
+spec:
+  applications:
+# This needs to be first to set up permissions
+  - kustomizeConfig:
+      repoRef:
+        name: manifests
+        path: stacks/openshift/application/openshift/openshift-scc
```
Tomcli, line 12: should we change the `application` directory name to plural similar to `stacks`?

kfdef/kfctl_openshift.v1.1.0.yaml
```diff
@@ -0,0 +1,103 @@
...
+      repoRef:
+        name: manifests
+        path: stacks/openshift/application/katib
+    name: katib
+  # commenting this out since it is included in kubeflow-apps
+  #- kustomizeConfig:
+  #    repoRef:
+  #      name: manifests
+  #      path: stacks/openshift/application/pipeline
+  #  name: pipeline
```
Tomcli, line 89: Is it possible to install `pipeline` and the Kubeflow control plane separately? Right now in this kfdef, I don't see a way to opt-out pipeline like other `application`.

kfdef/kfctl_openshift.v1.1.0.yaml
Tomcli: Discard this comment, somehow the singular application is being used on other stacks as well.

kfdef/kfctl_openshift.v1.1.0.yaml
nakfour: @Tomcli we would like to do that, however we tried and it did not work. We documented the issue here in a comment: https://github.com/nakfour/manifests/blob/openshift_stack/stacks/openshift/kustomization.yaml
and here https://github.com/opendatahub-io/manifests/issues/54
In summary, pipeline needs profiles setup first, if we take profiles as a single application and run it, $namespace for the virtualservice is not being replaced, so it fails. The plan is to keep it in the main app and re-visit this later.

Tomcli: Technically you can name two applications with the same name. This way kfctl will build those two applications into a single kustomization and shares the same variables. However, you will need to use the kfctl 1.2 (master branch) because kfctl 1.1 has few bugs with the same application name.

nakfour: @Tomcli not sure I understand your comment. Which two applications have the same name?

moficodes: @nakfour https://github.com/kubeflow/manifests/blob/master/kfdef/kfctl_ibm_customstack.yaml#L62-L112 
all these are part of the same application

Tomcli: To clarify, when you put the same name on two applications, kfctl build will group those two applications into a single kustomize. Here you can put `pipelines` and `base` to have the same applications.name `kubeflow-apps`, so they can share the `$namespace` variable.

animeshsingh: @nakfour have requested @Tomcli and @moficodes to review your PR so that we can adopt a consistent style across RedHat and IBM minimum, and it will be easier for us to get our customized changes in

animeshsingh:
> @jlewi @animeshsingh is there a way to get detailed logs from the kubeflow-manifests-presubmit job? Its failing and we have no indication why?

@nakfour can you rebase your PR on the latest and rerun. Hopefully that clears you up

Tomcli:
For the tests I opened a quick PR #1604 to fix it, it looks like we are using a very old kyaml library in our tests and it cannot load non SequenceNode yaml.

animeshsingh:
thanks @Tomcli !

animeshsingh:
@nakfour please rebase once, and hopefully this shall be good.

nakfour:
Thanks @Tomcli and @animeshsingh, rebased, the tests skip the non SequenceNode, but still fail on code outside our stack
```
=== RUN   TestValidK8sResources
--- FAIL: TestValidK8sResources (0.98s)
    validate_resources_test.go:249: Path ../istio/istio-install/base/istio-install-test.yaml; resource jaeger-query; has empty annotations; if no annotations are present the field shouldn't be present
    validate_resources_test.go:249: Path ../istio/istio-install/base/istio-install-test.yaml; resource tracing; has empty annotations; if no annotations are present the field shouldn't be present
    validate_resources_test.go:236: Path ../jupyter/notebook-controller/base/notebook-controller-test.yaml; resource notebooks.kubeflow.org; has status field
```

Tomcli:
> Thanks @Tomcli and @animeshsingh, rebased, the tests skip the non SequenceNode, but still fail on code outside our stack
> 
> ```
> === RUN   TestValidK8sResources
> --- FAIL: TestValidK8sResources (0.98s)
>     validate_resources_test.go:249: Path ../istio/istio-install/base/istio-install-test.yaml; resource jaeger-query; has empty annotations; if no annotations are present the field shouldn't be present
>     validate_resources_test.go:249: Path ../istio/istio-install/base/istio-install-test.yaml; resource tracing; has empty annotations; if no annotations are present the field shouldn't be present
>     validate_resources_test.go:236: Path ../jupyter/notebook-controller/base/notebook-controller-test.yaml; resource notebooks.kubeflow.org; has status field
> ```

You need to use kustomize v3.2.1 since that's the kustomize version for generating all these tests. You can see that the CI passes without these issues.

nakfour:
@Tomcli I am using Kustomize v3.4.0

Tomcli:
> @Tomcli I am using Kustomize v3.4.0

From https://github.com/kubeflow/manifests#manifests
Anything above Kustomize v3.2.1 may have problems with the tests.

kfdef/kfctl_openshift.v1.1.0.yaml
nakfour: @animeshsingh @Tomcli looking at kfctl_ibm.v1.1.0.yaml each app has a different name, same for kfctl_k8s_istio.v1.1.0.yaml. I am suggesting we keep kfctl_openshift.v1.1.0.yaml as it is now. 
/cc @vpavlin @crobby

vpavlin: @Tomcli having the same name for multiple apps sounds like a dirty workaround to an actual problem - like sharing a set of common/global variables/parameters - are there any plans to resolve this more systemically?

One feature I have in mind for the operator is to set state (success/failure) for each application name in `status` of the CR - this would allow users to see how the deployment went - obviously how do we figure out if the deployment was successful is a separate question..but if all the apps have the same name we lose this option, so I don't think this is a good idea in general

stacks/openshift/application/katib/katib-mysql-deployment.yaml
```diff
@@ -0,0 +1,11 @@
...
+  path: /spec/template/spec/containers/0/livenessProbe/exec/command/2
+  value: 'mysqladmin ping -uroot'
+- op: replace
+  path: /spec/template/spec/containers/0/readinessProbe/exec/command/2
+  value: 'mysql -D ${MYSQL_DATABASE} -uroot -e ''SELECT 1'''
+- op: remove
+  path: /spec/template/spec/containers/0/args
+- op: replace
+  path: /spec/template/spec/containers/0/volumeMounts/0/mountPath
+  value: /var/lib/mysql/data
```
vpavlin, line 11: missing new line

kfdef/kfctl_openshift.v1.1.0.yaml
Tomcli: As of now, kfctl doesn't have a way to define global variables. I opened an issue at https://github.com/kubeflow/kfctl/issues/429 to address this issue. It's fine to keep the `kfctl_openshift.v1.1.0.yaml` as it for now, but it would be nice to have some comments in the kfdef to clarify what's in the base application.

PatrickXYS:
Can you try re-run new test?

/test ?

kfdef/kfctl_openshift.v1.1.0.yaml
animeshsingh: @vpavlin @nakfour @Tomcli @moficodes is there an other alternative way in lieu of this to be consistent? What we don't like is bundling together something like pipelines with base stack in core openshift stacj, which means for someone just looking at pipelines they will be sort of forced to get the core components as well..... We ideally want to figure out an easier way to swap-in/swap out pipelines from OpenShift manifests, and hence aligning on a consistent style across will make more sense, else its a combination of multiple kfdef files floating for OpenShift

animeshsingh: > One feature I have in mind for the operator is to set state (success/failure) for each application name in `status` of the CR - this would allow users to see how the deployment went - obviously how do we figure out if the deployment was successful is a separate question..

@vpavlin @nakfour if this is something you can drive at your end, believe would be helpful for the overall odh umbrella as well. Not sure of the effort required in terms of the timeline et all.

PatrickXYS:
/hold

Hold until rebase and re-run test

kfdef/kfctl_openshift.v1.1.0.yaml
nakfour: @animeshsingh I agree we need a way to allow users to add/remove pipelines, we need to table this discussion for next release. Please open an issue for it and we can track.

nakfour:
@PatrickXYS rebase done
/test all

kfdef/kfctl_openshift.v1.1.0.yaml
animeshsingh: Thanks @nakfour 
Next release of what?

crobby:
I'm in the process of trying this out on OpenShift 4.6 and it looks like Seldon will not work due to a known issue:  https://github.com/kubeflow/manifests/issues/1543

nakfour:
@crobby the seldon issue should be fixed in the upgrade: https://github.com/kubeflow/manifests/pull/1600

kfdef/kfctl_openshift.v1.1.0.yaml
Tomcli: @nakfour I think it's possible to add/remove pipelines as an independent application. I see there's some conflicts on putting multi-user kfp on the single user openshift settings. Also there's a bug on the scc settings that's probably failing the deployment.

After apply the below fixes, we are able to deploy your openshift kfdef on OCP 4.3 with the pipeline application. On OCP 4.5 only seldon is failing but it should be resolved in #1600 after the merge.
https://github.com/nakfour/manifests/compare/openshift_stack...Tomcli:openshift-refactor

Tomcli: I do think able to add/remove pipelines is important because users can have an easier time to switch between single and multi-user kfp.

crobby:
I think notebooks spin-up is also failing due to the fact it's trying to use user/group ids that are not permitted in OpenShift.

nakfour:
@crobby istio side car injection requires  NET_ADMIN capability as described in this blog: https://www.openshift.com/blog/increasing-security-of-istio-deployments-by-removing-the-need-for-privileged-containers
in order to change the ip tables. So in OCP we have to give each pod privileged access in that namespace which is not recommended. I will try and see if I can disable side car injection.

kfdef/kfctl_openshift.v1.1.0.yaml
animeshsingh: @vpavlin @nakfour would be good to resolve this before we merge this PR. I have mentioned many times we are currently the biggest user for KF on OpenShift, and if we want consistency and not wanting multiple KF on OpenShift instructions mechanisms and kfdefs to proliferate, its best we align on one way. Additionally, I am not sure if you are planning to launch KF 1.2 on OpenShift, or KF 1.1.  If it's KF 1.1 we can wait a while, resolve and align before we get this in.

vpavlin: > We ideally want to figure out an easier way to swap-in/swap out pipelines from OpenShift manifests,

Can we first try to get this merged and then revisit this, as @nakfour suggested? I understand your priorities, but I think it is worth having this in ASAP and then refactor in smaller chunks

> if this is something you can drive at your end,

Yes, that is the plan, we need this for our deployments internally and on MOC

> I am not sure if you are planning to launch KF 1.2 on OpenShift

The last we talked about this, I believe the agreement was we'd skip 1.1 and target 1.2 instead to catch up with KF....Do I remember correctly, @nakfour?

Tomcli: > Can we first try to get this merged and then revisit this, as @nakfour suggested? I understand your priorities, but I think it is worth having this in ASAP and then refactor in smaller chunks

I found a few bugs in the openshift/application. https://github.com/nakfour/manifests/compare/openshift_stack...Tomcli:openshift-refactor can provide an easier way to swap-in/swap out single user pipelines.

stacks/openshift/application/openshift/openshift-scc/kubeflow-anyuid-scc-istio.yaml
```diff
@@ -0,0 +1,52 @@
...
+- system:serviceaccount:istio-system:istio-cleanup-old-ca-service-account
+- system:serviceaccount:istio-system:istio-mixer-post-install-account
+- system:serviceaccount:istio-system:istio-mixer-service-account
+- system:serviceaccount:istio-system:istio-pilot-service-account
+- system:serviceaccount:istio-system:istio-sidecar-injector-service-account
+- system:serviceaccount:istio-system:istio-sidecar-injector-service-account
+- system:serviceaccount:istio-system:istio-galley-service-account
+- system:serviceaccount:istio-system:prometheus
+#Component: istio/cluster-local-gateway
+- system:serviceaccount:cluster-local-gateway-service-account
```
Tomcli, line 45: missing istio-system namespace, without it Kubeflow cannot communicate with the cluster-local-gateway
```suggestion
- system:serviceaccount:istio-system:cluster-local-gateway-service-account
```

stacks/openshift/application/pipeline/kustomization.yaml
```diff
@@ -0,0 +1,43 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: kubeflow
+resources:
+- ../../../../pipeline/minio/installs/generic
+- ../../../../pipeline/mysql/installs/generic
+- ../../../../pipeline/installs/multi-user
```
Tomcli, line 7: Why are we using multi-user kfp over here? In this PR's kfdef, there's no authentication services and istio sidecar injection to provide istio auth. Thus, users cannot run pipelines on the multi-user kfp because they cannot authenticate themselves.

Tomcli: Here is how to make it run with single user, https://github.com/nakfour/manifests/compare/openshift_stack...Tomcli:openshift-refactor

stacks/openshift/application/openshift/openshift-scc/kubeflow-anyuid-scc-istio.yaml
nakfour: @Tomcli good catch, thanks.

stacks/openshift/application/pipeline/kustomization.yaml
nakfour: We disabled istio-injection, as posted in a comment before. We at the moment cannot pull pipeline into applications because of its dependency on profiles. We discussed this before

Tomcli: Yes, I understand you want to disable istio-injection, so we should using the single user kfp isn't it? With the single user kfp, there shouldn't be a dependency on profiles.

Tomcli: `../../../../pipeline/installs/generic` is the single user mode kfp

nakfour: Ok but. we would like users to run pipelines in their own namespace generated by the profile controller. Isn't that the whole point of multi-user pipelines

nakfour:
@PatrickXYS @Tomcli  how can we get logs on what is failing in the e2e test? Running tests locally pass, and this stack is specific to Openshift cluster.

stacks/openshift/application/pipeline/kustomization.yaml
Tomcli: Sure, but in that case you will need to provide authentication services and istio sidecar injection to authenticate incoming user traffics. And the current istio sidecar injection is not ideal for openshift.

Tomcli: right now with this deployment, you cannot submit pipelines as a user.

nakfour: Yes switched pipeline to generic and pulled out into kfdef

Tomcli:
thanks @nakfour 
/lgtm

animeshsingh:
Thanks @nakfour 
/lgtm
/approve

PR #1579, opened 2020-10-12
Explicitly disable container native load balancing (neg)

fabito:
# Problem

`envoy-ingress` backend services unhealthy when installing Kubeflow on GKE version `1.17.9-gke.1504`

# Solution

Add the `cloud.google.com/neg: '{"ingress": false}'` annotation to `istio-ingressgateway` to disable **container-native load balancing** (neg) which is enabled by default for GKE clusters >= 1.17 ( https://cloud.google.com/kubernetes-engine/docs/how-to/container-native-load-balancing#using )

https://github.com/kubeflow/gcp-blueprints/pull/141


Bobgy:
/assign @jlewi 
Can you help me take a look? I don't know enough about these GCP features.

PR #1582, opened 2020-10-14
[IBM] split full stack to components to support finer grain customization used in KfDef

adrian555:
Resolves #

This PR splits the full Kubeflow stack for IBM Cloud platform in `stacks/ibm/kustomization.yaml` to a set of components, including `control plane`, `argo`, `pipelines`, `notebooks` etc. This enables a finer grain customization of Kubeflow components when users using `KfDef` configuration file to install Kubeflow. For example, if `notebooks` is not part of the `KfDef` configuration, then the Kubeflow's notebook server component won't be installed.

A sample `KfDef` configuration file `kfctl_ibm_customstack.yaml` is provided.

**Checklist:**
- [ x ] Unit tests have been rebuilt: 
    1. `cd manifests/tests`
    3. `make test`

### DIFF
--- a/stacks/ibm/application/istio/kustomization.yaml
+++ b/stacks/ibm/application/istio/kustomization.yaml
@@ -2,9 +2,9 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 namespace: kubeflow
 resources:
-- ../../../../istio/istio/base
+- ../../../../istio/istio/base_v3
 configMapGenerator:
-- name: istio-parameters
+- name: istio-config
   behavior: merge
   envs:
   - params.env
--- a/stacks/ibm/components/seldon-core-operator/kustomization.yaml
+++ b/stacks/ibm/components/seldon-core-operator/kustomization.yaml
@@ -1,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: kubeflow
+resources:
+- ../../../../seldon/seldon-core-operator/overlays/application
--- a/kfdef/kfctl_ibm_customstack.yaml
+++ b/kfdef/kfctl_ibm_customstack.yaml
@@ -1,0 +1,144 @@
+apiVersion: kfdef.apps.kubeflow.org/v1
+kind: KfDef
+metadata:
+  namespace: kubeflow
+spec:
+  applications:
+  # Install istio in a different namespace: istio-system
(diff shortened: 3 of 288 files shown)


adrian555:
cc @Tomcli @pvaneck @animeshsingh

stacks/ibm/components/seldon-core-operator/kustomization.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: kubeflow
+resources:
+- ../../../../seldon/seldon-core-operator/overlays/application
```
pvaneck, line 5: Since a `stacks/ibm/application/seldon-core-operator` exists, should that be referenced here instead?

adrian555: @pvaneck I was debating whether `seldon` should go with components as it was part of the `stacks/ibm/kustomization.yaml` when we first created the full stack version. Then the thought goes that `seldon` may not necessarily a component of Kubeflow, so I would like to provide it as an application. So I ended up providing both kustomization.yaml file. At some point in time we may remove seldon from both the full stack and the components.

stacks/ibm/application/istio/kustomization.yaml
```diff
@@ -2,9 +2,9 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 namespace: kubeflow
 resources:
-- ../../../../istio/istio/base
+- ../../../../istio/istio/base_v3
```
pvaneck, line 5: Maybe not particularly relevant to this PR, but since some resources here are being updated to `base_v3`, should `stacks/ibm/application/istio-1-3-1-stack/kustomization.yaml` and `stacks/ibm/application/cluster-local-gateway-1-3-1/kustomization.yaml` now also reference their respective `base_v3` directories?

adrian555: Minor note: the equivalent of `istio-1-3-1-stack` is `istio-stack`. I updated the manifests to use `base_v3`.

Tomcli:
Thanks @adrian555, can we have a follow up PR that do the same for our multi-user kfdef?
/lgtm

animeshsingh (approved):
/lgtm

As discussed, we should follow this on with an option to provide KFP with Tekton

As20075 (commented):
انا

PR #1588, opened 2020-10-20
add permissions for kubectl top pods

munagekar:

This allows the usage of kubectl top pods. 

This is useful when setting resource requests when using kubeflow/pipelines. I think this will be useful to most users of kubeflow.

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

### DIFF
--- a/kubeflow-roles/base/cluster-roles.yaml
+++ b/kubeflow-roles/base/cluster-roles.yaml
@@ -333,3 +333,10 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - metrics.k8s.io
+  resources:
+  - pods
+  verbs:
+  - get
+  - list


PatrickXYS:
@munagekar 

For the kubeflow role, we want to keep minimum permission at first, can you explain what functionality will be effected without this permission?

Or it would be better to give a usage-list.

munagekar:
@PatrickXYS 

Thank you for the review. 

This allows contributors to use the `kubectl top pods` command. This shows the cpu usage and memory usage, if metric-server is installed on kubernetes server. So users will be able to check the Jupyter Notebook and Kubeflow Pipelines Pod Usage and set the resource requirement and limit.

If this does not align with current design, feel free to close this PR.

PatrickXYS:
@munagekar I understand your point, but why metrics-server has to use Kubeflow-Role?

Can you create your own k8s role and your metrics server use it?

munagekar:
Thank you for your reply, I understand the design philosophy. I will close this PR

PR #1604, opened 2020-11-03
[Fix tests] Skip non SequenceNode yaml to work around on the old kyaml bug

Tomcli:
Related #1567

Since we are still using kyaml v0.1.11 for our tests, it has some bugs when parsing non SequenceNode yaml. Although it's resolved in the newer release of kyaml, updating kyaml will dramatically change the current generated tests. Therefore, we skip all the files with `wrong Node Kind for` since we only need to check the proper kubernetes yaml.

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

### DIFF
--- a/tests/validate_resources_test.go
+++ b/tests/validate_resources_test.go
@@ -311,6 +311,10 @@ func TestCheckWebhookSelector(t *testing.T) {
 		nodes, err := reader.Read()
 
 		if err != nil {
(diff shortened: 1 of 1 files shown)


animeshsingh:
Thanks @Tomcli 
/approve

 cc @jlewi

PR #1609, opened 2020-11-04
Add Exit Steps for Unit Test

PatrickXYS:
Since current unit doesn't provide Exit Step, which upload artifacts and delete test-dir. 
image

Adding Exithandler to handle onExit logic

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

Commenting E2E test without testing E2E workflow, hold until Unit Test succeed, will add E2E test back in the PR
/hold

### DIFF
--- a/prow_config.yaml
+++ b/prow_config.yaml
@@ -13,45 +13,45 @@ workflows:
       - presubmit
       - postsubmit
       - periodic
-  - py_func: kubeflow.kfctl.testing.ci.kfctl_e2e_workflow.create_workflow
-    name: e2e
-    job_types:
-      - presubmit
-      - postsubmit
-      - periodic 
-    include_dirs:
-      - admission-webhook/*
-      - application/*
-      - argo/*
-      - aws/*
-      - cert-manager/*
-      - common/
-      - default-install/*
-      - dex-auth/*
-      - experimental/*
-      - istio-1-3-1/*
-      - istio/*
-      - jupyter/*
-      - katib/*
-      - kfdef/*
-      - kfserving/*
-      - knative/*
-      - kubebench/*
-      - kubeflow-roles/*
-      - metacontroller/*
-      - metadata/*
-      - mpi-job/*
-      - mxnet-job/*
-      - namespaces/*
-      - pipeline/*
-      - profiles/*
-      - pytorch-job/*
-      - seldon/*
-      - spark/*
-      - stacks/*
-      - tektoncd/*
-      - tf-training/*
-      - xgboost-job/*
-    kwargs:
-      build_and_apply: false
-      config_path: https://raw.githubusercontent.com/kubeflow/manifests/master/kfdef/kfctl_istio_dex.yaml
+  # - py_func: kubeflow.kfctl.testing.ci.kfctl_e2e_workflow.create_workflow
+  #   name: e2e
+  #   job_types:
+  #     - presubmit
+  #     - postsubmit
+  #     - periodic 
+  #   include_dirs:
+  #     - admission-webhook/*
+  #     - application/*
+  #     - argo/*
+  #     - aws/*
+  #     - cert-manager/*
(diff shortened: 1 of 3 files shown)


prow_config.yaml
```diff
@@ -13,45 +13,45 @@ workflows:
...
-      - spark/*
-      - stacks/*
-      - tektoncd/*
-      - tf-training/*
-      - xgboost-job/*
-    kwargs:
-      build_and_apply: false
-      config_path: https://raw.githubusercontent.com/kubeflow/manifests/master/kfdef/kfctl_istio_dex.yaml
\ No newline at end of file
+  # - py_func: kubeflow.kfctl.testing.ci.kfctl_e2e_workflow.create_workflow
```
Jeffwan, line 16: Any reason to delete test case?

jlewi: I think we should delete this test for two reasons

1. We should have gotten rid of the E2E tests on this repo along time ago; we shouldn't be running kfctl on presubmits anymore

1. We should get rid of the use of py_func and Argo for testing

The only tests I think this repo should be running our go unittests
https://github.com/kubeflow/manifests/tree/master/tests

So I think we can just turn on the Tekton go unittests by adding a pipelinerun.yaml
https://github.com/kubeflow/testing/blob/b2b4b2f06ce48b3f18cf21bbc47b42efa9913a12/prow_config.yaml#L16

tests/workflows/components/workflows.libsonnet
```diff
@@ -138,7 +138,8 @@
         },
         spec: {
           entrypoint: "e2e",
-          ttlSecondsAfterFinished: 3600,
+          ttlSecondsAfterFinished: 8*60*60,
```
jlewi, line 141: I don't think we should be investing anymore in ksonnet/jsonnet

prow_config.yaml
jlewi: kubeflow/kfctl#57 is a related issue tracking our overreliance on expensive tests and the desire to get rid of them.

jlewi: #818 is the more specific issue for kubeflow/manifests

PatrickXYS: > Commenting E2E test without testing E2E workflow, hold until Unit Test succeed, will add E2E test back in the PR
Commenting this to skip E2E test for this PR, because I only want to improve unit test

tests/workflows/components/workflows.libsonnet
PatrickXYS: Agree, I should move to Tekton or python-based Argo workflow

PatrickXYS:
FYI, this is not high priority, feel to skip this for 1.2

prow_config.yaml
jlewi: @PatrickXYS Why not just get rid of the E2E test altogether? i.e. just delete it.

PatrickXYS: @jlewi From the issue you mentioned before,

I agree with `E2E Test is expensive and become flaky`, but I think we want a E2E test to test orchestration and conformance even though current E2E test is not perfect.

I think to make E2E test more modular and stable, think about `Does it really make sense when testing changes to kfctl to test that all the Kubeflow applications are deployed and healthy?`.

We need @kubeflow/wg-manifests-leads @kubeflow/wg-deployment-leads collaborate and discuss about next step after kubeflow 1.2. 

Can we wait until 1.2 finished, and let's discuss about the next step for E2E test on kfctl and manifests repo?

PatrickXYS:
@jlewi To answer your questions, I gathered all the testing pain points in this issue https://github.com/kubeflow/internal-acls/pull/393

We can talk in that issue to better track our discussion

prow_config.yaml
jlewi: Opened #1613 1613

PR #1611, opened 2020-11-05
Increase unit test argo workflow TTL

PatrickXYS:
Resolves #

Increase unit test TTL from 1h to 8h

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

### DIFF
--- a/tests/workflows/components/workflows.libsonnet
+++ b/tests/workflows/components/workflows.libsonnet
@@ -138,7 +138,7 @@
         },
         spec: {
           entrypoint: "e2e",
-          ttlSecondsAfterFinished: 3600,
+          ttlSecondsAfterFinished: 8*60*60,
           volumes: [
             {
               name: "github-token",


tests/workflows/components/workflows.libsonnet
```diff
@@ -138,7 +138,7 @@
         },
         spec: {
           entrypoint: "e2e",
-          ttlSecondsAfterFinished: 3600,
+          ttlSecondsAfterFinished: 8*60*60,
```
Jeffwan, line 141: Please calculate the number directly, leave `// xx days/hrs `

Jeffwan: BTW, is it better to change Argo configuration in CI cluster directly? Then we don't have change it per job

Jeffwan (changes requested):

PatrickXYS:
> BTW, is it better to change Argo configuration in CI cluster directly? Then we don't have change it per job

This feature is not supported in argo 2.4.3, but in later version, check here for more details (https://argoproj.github.io/argo/fields/#ttlstrategy)

tests/workflows/components/workflows.libsonnet
PatrickXYS: Emmm don't you think `8*60*60` might make more sense to users, otherwise, they need to calculate how long is 28800 seconds, I think it make sense to add comment like `//8hours`

Jeffwan: This is minor issue. I am fine.

Jeffwan:
> > BTW, is it better to change Argo configuration in CI cluster directly? Then we don't have change it per job
> 
> This feature is not supported in argo 2.4.3, but in later version, check here for more details (https://argoproj.github.io/argo/fields/#ttlstrategy)

I think manifest repo change will be good enough. Seems all others won't take that long.

Jeffwan (approved):

PR #1612, opened 2020-11-05
Add AWS Periodic Tests

PatrickXYS:
Resolves #

Add AWS Periodic Tests to test orchestration on Kubeflow AWS.

Also removes periodic tests for e2e and unit tests, since no one will check it daily and see if they succeed or fail, thus, it only consume resources without any contribution.

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

### DIFF
--- a/prow_config.yaml
+++ b/prow_config.yaml
@@ -12,13 +12,11 @@ workflows:
     job_types:
       - presubmit
       - postsubmit
-      - periodic  
   - py_func: kubeflow.kfctl.testing.ci.kfctl_e2e_workflow.create_workflow
     name: e2e
     job_types:
       - presubmit
       - postsubmit
-      - periodic 
     include_dirs:
       - admission-webhook/*
       - application/*
@@ -55,3 +53,13 @@ workflows:
     kwargs:
       build_and_apply: false
       config_path: https://raw.githubusercontent.com/kubeflow/manifests/master/kfdef/kfctl_istio_dex.yaml
+
+  ####################################### AWS Specific Tests
+  - py_func: kubeflow.kfctl.testing.ci.kfctl_e2e_workflow.create_workflow
+    name: aws-e2e
+    job_types:
+      - periodic
+    kwargs:
+      build_and_apply: false
+      config_path: https://raw.githubusercontent.com/kubeflow/manifests/master/kfdef/kfctl_aws.yaml
+    


terrytangyuan (approved):
/lgtm

prow_config.yaml
```diff
@@ -54,4 +52,14 @@ workflows:
       - xgboost-job/*
     kwargs:
       build_and_apply: false
-      config_path: https://raw.githubusercontent.com/kubeflow/manifests/master/kfdef/kfctl_istio_dex.yaml
\ No newline at end of file
+      config_path: https://raw.githubusercontent.com/kubeflow/manifests/master/kfdef/kfctl_istio_dex.yaml
```
Jeffwan, line 55: em. I know kubeflow minimum distribution is still being discussed, but I hope to keep it until we figure it out If this is not a problem at this moment.

Jeffwan:
> Also removes periodic tests for e2e and unit tests, since no one will check it daily and see if they succeed or fail, thus, it only consume resources without any contribution.

This doesn't make sense to me. It's ok to add AWS or other cloud provider cases here. That's totally separate story. AWS test won't guarantee kubeflow manifest wok for other platforms.

PatrickXYS:
> > Also removes periodic tests for e2e and unit tests, since no one will check it daily and see if they succeed or fail, thus, it only consume resources without any contribution.
> 
> This doesn't make sense to me. It's ok to add AWS or other cloud provider cases here. That's a separate story. AWS test won't guarantee kubeflow manifest wok for other platforms.

I agree with the idea AWS test is separate story. The reason I removed minimum distribution periodic job is presubmit should be able to detect any failures, only situation I can come up with is establishing Kubeflow-Kubernetes Version Compatibility Matrix, but I don't think we want to support before 1.2.

Jeffwan:
@PatrickXYS 
 
Kubeflow compability matrics is a separate story.. We will give some matrix in 1.2 but that's not related. 

I would suggest to bring it back since this is cloud agnostic and only have some core components and can reflect master/branch control plane status. Presubmit does help but this won't cover all the cases. For example, some dependency change (like kfctl, kuebrnetes cluster version), Github update (tar.gz or https issue in the past) or GCR outage, etc. This help us make quick response to Kubernete users. 

We already had the pain in 1.1 and hope at least we can run it during release cycle.

PatrickXYS:
@Jeffwan 

Thank you for explanations! That sounds good to me, I'll add those periodic test back

PatrickXYS:
Created PR https://github.com/kubeflow/manifests/pull/1615

PR #1630, opened 2020-11-10
Presubmit kicked off by istio_dex manifests

PatrickXYS:
Resolves #

Since cloud providers are changing kfdef/* manifests frequently, and we should only kick off E2E tests when `kfctl_istio_dex` manifests.

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

### DIFF
--- a/prow_config.yaml
+++ b/prow_config.yaml
@@ -33,7 +33,7 @@ workflows:
       - istio/*
       - jupyter/*
       - katib/*
-      - kfdef/*
+      - kfdef/kfctl_istio_dex*
       - kfserving/*
       - knative/*
       - kubebench/*


prow_config.yaml
```diff
@@ -33,7 +33,7 @@ workflows:
       - istio/*
       - jupyter/*
       - katib/*
-      - kfdef/*
+      - kfdef/kfctl_istio_dex*
```
Jeffwan, line 36: em. I thought `kfdef/*` is a super set of `kfdef/kfctl_istio_dex*`. Any reason to change it?

Jeffwan (approved):

prow_config.yaml
PatrickXYS: So if cloud providers are changing `kfctl_aws` / `kfctl_ibm`, `kfctl_openshift`, it should not kick off E2E test because it's not agnostic change.

And I found folks may change very frequently and this change can help speed up and reserve resources.

Jeffwan: Got it. Make sense.  `kfctl_istio_dex ` should be good enough.

PR #1625, opened 2020-11-10
Add kfdef 1.2 manifests for Azure

berndverst:
**Which issue is resolved by this Pull Request:**
Resolves #1597 for Azure

**Description of your changes:**
Copying the approach of @Jeffwan in #1620 for Azure: 
Adds two Azure manifests files for v1.2.0 identical to v1.1.0 except the manifest resource now points to the v.1.2.0 branch.

Note this merges several existing Azure specific commits from the v1.1.0 branch into Master
- Add overlay to support Azure Database for MySQL (#1587)
- Create v1.1.0 Azure Kfdef manifests with OIDC Azure AD  (#1576)
- Create v1.1.0 Azure Kfdef manifests (#1487)
-  Add Azure owners (#1524)

Eventually these commits will be cherry picked into the v1.2.0 branch.

**Checklist:**
- [] Unit tests have been rebuilt: 
    1. `cd manifests/tests`
    2. `make generate-changed-only`
    3. `make test`

Note that tests are currently broken in master it appears (or in v1.1.0) - I'm not introducing any testing changes. Could someone help look into this?

PR #1459, opened 2020-08-04
Remove tensorboard manifest

Jeffwan:
**Which issue is resolved by this Pull Request:**
Resolves #415 

**Description of your changes:**
Remove following items
 - tensorboard/
 - references in kfdef/. I didn't touch pinned manifest, just change master files.
 - stacks/  - aws and ibm has some references
 - tests/tests/legacy_kustomizations/tensorboard

**Checklist:**
- [x] Unit tests have been rebuilt: 
    1. `cd manifests/tests`
    2. `make generate-changed-only`
    3. `make test`

PR #1633, opened 2020-11-11
Replacement PR for PR #1625 - Azure kfdef manifests for 1.2 and previous

### LINKED ISSUE #1597: Manifests 1.2 release
Please help link all 1.2 release related issues here. 

- Openshift kubeflow/manifests#1567

berndverst:
Resolves #1597 for Azure

Please note that this PR replaces the previous PR #1625 discussed with @Jeffwan and #pvaneck


berndverst:
@Jeffwan this is the same PR as #1625 cleaned up. Commit Author changed now.  The reference to the original PRs still has the original authors. Also those commits remain unchanged in the v1.1 branch.

As discussed, there are no Azure changes -- this is what is already released as v1.1 and documented. Just moving it into Master. For v1.2 all I did was changing the target to the resource v1.2 branch just like you did.

@pvaneck Tensorboard is removed per your suggestion. It was not used in our existing kfdef manifests anyway.

yilun-msft:
Thanks for doing this @berndverst . Could you remind me why Tensorboard is removed?

berndverst:
@yilun-msft due to #1459, which originally is motivated by #415 which states Tensorboard should never have been deployed for Kubeflow by default. So due to #1459 our stack files caused test failures. Also, tensorboard isn't in our kfdef manifests anyway.

Jeffwan:
/lgtm
/approve

Please help update website later.

Jeffwan:
@PatrickXYS Can you help approve this change?

Jeffwan:
> Thanks for doing this @berndverst . Could you remind me why Tensorboard is removed?

That's not a mature solution and no one use it. there's a separate project going on to build a new one. Let's see

pvaneck (commented):
Since a stack was added, I think tests still need to be generated in this PR:
1. `cd manifests/tests`
3. `make test`

Just make sure you have golang and kustomize v3.2.1 installed. And if I remember correctly, periods in directory names cause issues with the test generation/testing, so you will likely hit an error when running the `make generate-changed-only` while having the `add-anonymous-user-filter-istio-1.6` directories in `stacks/azure` and `istio`. Can remedy this by changing the names to `add-anonymous-user-filter-istio-1-6`. This aligns with how other directories are named like `istio-1-3-1`. Then just update the reference to the new name in the v1.2.0 Kfdef yaml (the v1.1.0 Kfdef should still use the old name as that branch won't have the change).

berndverst:
Let me do this right away and get back to you.

I wasn't sure whether the generated tests should be added the PR.

Jeffwan:
@berndverst Any generate tests should be included. 

/cc @PatrickXYS Can you help approve this PR?

PatrickXYS:
/lgtm
/approve 

Thanks @berndverst !

PR #1638, opened 2020-11-12
feat: remove kf metadata server

Bobgy:
Part of https://github.com/kubeflow/metadata/issues/250

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`


Bobgy:
/assign @Jeffwan @PatrickXYS @neuromage 
I'm not sure we want to do this now, a problem is that when KF metadata UI virtual service is removed. I think users cannot access KF metadata server either, so I think there's no reason to keep metadata server too.

PatrickXYS:
I think removal sounds reasonable since we have already removed metadata UI virtualservice.

SGTM

/cc @kubeflow/wg-notebook-leads

Jeffwan:
@Bobgy @PatrickXYS I think it makes sense to remove all and only leave grpc deployment and server there. https://github.com/kubeflow/metadata/issues/250

Jeffwan:
https://github.com/kubeflow/kubeflow/pull/5404 PR to remove link in central dashboard

PR #1647, opened 2020-11-16
fix commonLabels

thesuperzapper:
Makes the following components have consistent `commonLabels` between `base` and `base_v3`:
- centraldashboard
- jupyter-web-app
- notebook-controller

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

### DIFF
--- a/jupyter/jupyter-web-app/base_v3/kustomization.yaml
+++ b/jupyter/jupyter-web-app/base_v3/kustomization.yaml
@@ -12,20 +12,12 @@
 # break the existing KFDef files. So we want to make the stacks work 
 # and then replace it.
 apiVersion: kustomize.config.k8s.io/v1beta1
-commonLabels:
-  app.kubernetes.io/component: jupyter-web-app
-  app.kubernetes.io/instance: jupyter-web-app-v1.0.0
-  app.kubernetes.io/managed-by: kfctl
-  app.kubernetes.io/name: jupyter-web-app
-  app.kubernetes.io/part-of: kubeflow
-  app.kubernetes.io/version: v1.0.0
 kind: Kustomization
 namePrefix: jupyter-web-app-
 namespace: kubeflow
 commonLabels:
   app: jupyter-web-app
   kustomize.component: jupyter-web-app
-namespace: kubeflow
 images:
 - name: gcr.io/kubeflow-images-public/jupyter-web-app
   newName: gcr.io/kubeflow-images-public/jupyter-web-app
--- a/tests/stacks/aws/test_data/expected/apps_v1_deployment_centraldashboard.yaml
+++ b/tests/stacks/aws/test_data/expected/apps_v1_deployment_centraldashboard.yaml
@@ -5,6 +5,7 @@ metadata:
     app: centraldashboard
     app.kubernetes.io/component: centraldashboard
     app.kubernetes.io/name: centraldashboard
+    kustomize.component: centraldashboard
   name: centraldashboard
   namespace: kubeflow
 spec:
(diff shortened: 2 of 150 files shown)


thesuperzapper:
@Jeffwan can you please review?

Jeffwan:
It looks good to me. I am not sure if there's still user on `base/` probably some 1.0 users. In the future, we may want to remove and deprecate legacy manifests

/lgtm

thesuperzapper:
@PatrickXYS can you approve this, as we need someone who is in the root OWNERS file

thesuperzapper:
@Bobgy could you also approve this?

jupyter/jupyter-web-app/base_v3/kustomization.yaml
```diff
@@ -12,20 +12,12 @@
 # break the existing KFDef files. So we want to make the stacks work 
 # and then replace it.
 apiVersion: kustomize.config.k8s.io/v1beta1
-commonLabels:
```
Bobgy, line 15: Why do we remove everything here?

tests/stacks/aws/test_data/expected/apps_v1_deployment_centraldashboard.yaml
```diff
@@ -5,6 +5,7 @@ metadata:
     app: centraldashboard
     app.kubernetes.io/component: centraldashboard
     app.kubernetes.io/name: centraldashboard
```
Bobgy, line 7: Looking at your PR, this label shouldn't exist now. Is this a mistake?

Bobgy (commented):
Generally updating labels is a breaking change to upgrade experience.
We can document to tell users delete old deployments when they see the label selector immutable error message, but it's not ideal.

So I'd want to make sure we should minimize such changes as much as possible in the future.

Bobgy:
/cc @PatrickXYS 
the tests didn't fail, so I guess we may have lost tests for https://github.com/kubeflow/manifests/issues/1131 or the tests were not added properly initially?

jupyter/jupyter-web-app/base_v3/kustomization.yaml
thesuperzapper: This is a duplication in the file, look a few lines down.

tests/stacks/aws/test_data/expected/apps_v1_deployment_centraldashboard.yaml
thesuperzapper: That will be coming from the `kustomization.yaml` in `overlay/application`:  https://github.com/kubeflow/manifests/blob/master/common/centraldashboard/overlays/application/kustomization.yaml

Bobgy: Sorry, still not fully clear.
Why do some test snapshots have the two labels, while some do not?
Is that expected?

thesuperzapper: If the stack is using the `application` overlay, there will be both, but not otherwise (because I removed them from the base).

Ones from `application` overlay: 
```yaml
    app.kubernetes.io/component: centraldashboard
    app.kubernetes.io/name: centraldashboard
```

Ones from base: 
```yaml
    app: centraldashboard
    kustomize.component: centraldashboard
```

thesuperzapper:
@Bobgy I agree that we need to confirm if we want to force users to remove their Deployment resources during the next upgrade.

I made these changes to make the labels consistent.

tests/stacks/aws/test_data/expected/apps_v1_deployment_centraldashboard.yaml
Bobgy: Thanks for clarification! Created an issue to track the problem: https://github.com/kubeflow/manifests/issues/1649.

PR #1645, opened 2020-11-16
update image tags

thesuperzapper:
**Which issue is resolved by this Pull Request:**

This resolves #1553 (for the components that belong to wg-notebook)

**Description of your changes:**

Updates the image tags for the following applications:
* admission-webhook
* centraldashboard
* jupyter-web-app
* notebook-controller
* kfam
* profile-controller

**Checklist:**
- [X] Unit tests have been rebuilt: 
    1. `cd manifests/tests`
    2. `make generate-changed-only`
    3. `make test`

Issue #1285, opened 2020-06-18
fix names for Application and VirtualService in jupyter-web-app/base_v3 and notebook-controller/base_v3 


adrian555:
As mentioned in the [comment](https://github.com/kubeflow/manifests/pull/1246#issuecomment-645103438), open this issue to track.

Basically the current `base_v3` for jupyter-web-app and notebook-controller applications are using `namePrefix` in kustomization.yaml. Hence the name for Application and VirtualService ends up as `jupyter-web-app-jupyter-web-app`. Would like to remove the namePrefix for these resources, but also make sure that the change should not break any current use of `base_v3`.

yanniszark:
Closing as there is no base_v3 any longer.

PR #1652, opened 2020-11-17
[IBM] Update notebooks and profiles image tags

pvaneck:
This PR updates the image tags that were listed in the IBM stack based on updates in https://github.com/kubeflow/manifests/pull/1645. 

Ideally, we get rid of the hard-coding here and use each component's `base_v3` directly, but I'm hesitant to do so based on https://github.com/kubeflow/manifests/issues/1285 . So for now, just update the relevant image tags in `stacks/ibm`.

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

### DIFF
--- a/stacks/ibm/application/jupyter-web-app/base/kustomization.yaml
+++ b/stacks/ibm/application/jupyter-web-app/base/kustomization.yaml
@@ -10,7 +10,7 @@ namespace: kubeflow
 images:
 - name: gcr.io/kubeflow-images-public/jupyter-web-app
   newName: gcr.io/kubeflow-images-public/jupyter-web-app
-  newTag: vmaster-gd9be4b9e
+  newTag: vmaster-g845af298
 resources:
 - ../../../../../jupyter/jupyter-web-app/base/cluster-role-binding.yaml
 - ../../../../../jupyter/jupyter-web-app/base/cluster-role.yaml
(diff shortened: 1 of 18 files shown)


animeshsingh:
thanks @pvaneck 

/lgtm
/approve

animeshsingh:
Does this need a cherry pick?

thesuperzapper:
@animeshsingh why is #1285 stopping you from removing the hard coding?

pvaneck:
Hey @thesuperzapper. If I remove just the `image:` sections, then it will just use the images that are listed in the `base/deployment.yaml` files which are old. (e.g. https://github.com/kubeflow/manifests/blob/master/jupyter/notebook-controller/base/deployment.yaml#L13)

Initially, I was changing everything to directly use the `base_v3` directories so as to pick up the new image tags, however, seems  previously there were some [name change concerns](https://github.com/kubeflow/manifests/pull/1246#issuecomment-644919271) raised by @adrian555 , and it seems like the [original intent](https://github.com/kubeflow/manifests/blob/master/stacks/ibm/application/jupyter-web-app/README.md) of the custom structure in `stacks/IBM` was to avoid the namePrefix being applied to something already with the prefix (e.g. `jupyter-web-app-jupyter-web-app`). Overall this may not be a big deal, but for now, since I was uncertain about all this and its side effects, I just opted to keep names of things consistent for this release by only updating the image tags.

PR #1657, opened 2020-11-18
Update jupyter web app image tag for Azure Stack

berndverst:
Resolves #1645

Updates hardcoded image tag for Jupyter Web App in Azure Stack

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

### DIFF
--- a/stacks/azure/application/jupyter-web-app/base/kustomization.yaml
+++ b/stacks/azure/application/jupyter-web-app/base/kustomization.yaml
@@ -10,7 +10,7 @@ namespace: kubeflow
 images:
 - name: gcr.io/kubeflow-images-public/jupyter-web-app
   newName: gcr.io/kubeflow-images-public/jupyter-web-app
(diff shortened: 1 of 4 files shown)


thesuperzapper:
@berndverst you might want to consider just not defining it in your `kustomization.yaml`, and instead delegating to the `base_v3`, so that you don't get out of sync in the future.

Jeffwan:
> @berndverst you might want to consider just not defining it in your `kustomization.yaml`, and instead delegating to the `base_v3`, so that you don't get out of sync in the future.

@thesuperzapper Agree. If platform like to do that way, they will need to be responsible for the reliability. As WG owners, it's hard to understand what version platform owners want. It's possible user want more reliable version. :D

Jeffwan:
@berndverst Do you need this in v1.2-branch? If so, can you help cherry-pick the changes?

PR #1669, opened 2020-11-20
fix katib application selectors

### LINKED ISSUE #1573: most Application Resource Selectors are Incorrect
Most of the Application `matchLabels` don't match the Resources they should.

This is partially caused by not merging auto-PRs as raised here: #1568

I think we can reduce the complexity of these label selectors significantly, but this is really a symptom of the mess in this Repo, and highlights a need to clean it up, as raised here: #1554

A few quick examples (NOTE: there are MANY more, as pretty much every Application is affected):
* jupyter-web-app:
   * https://github.com/kubeflow/manifests/blob/master/jupyter/jupyter-web-app/overlays/application/application.yaml#L7
* pipelines:
   * https://github.com/kubeflow/manifests/blob/master/pipeline/installs/generic/application.yaml#L11
* centraldashboard:
   * https://github.com/kubeflow/manifests/blob/master/common/centraldashboard/overlays/application/application.yaml#L7
* kfserving-install:
   * https://github.com/kubeflow/manifests/blob/master/kfserving/kfserving-install/overlays/application/application.yaml#L7

thesuperzapper:
Fixes #1573 (for katib)

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

### DIFF
--- a/katib/installs/katib-external-db/kustomization.yaml
+++ b/katib/installs/katib-external-db/kustomization.yaml
@@ -11,6 +11,12 @@ secretGenerator:
   - name: katib-mysql-secrets
     envs:
       - secrets.env
-commonLabels:
-  app.kubernetes.io/component: katib
-  app.kubernetes.io/name: katib-controller
+patches:
+  - patch: |-
+      apiVersion: v1
+      kind: Secret
+      metadata:
+        name: katib-mysql-secrets
+        labels:
+          app.kubernetes.io/component: katib
+          app.kubernetes.io/name: katib-controller
--- a/katib/katib-controller/overlays/application/application.yaml
+++ b/katib/katib-controller/overlays/application/application.yaml
@@ -6,11 +6,17 @@ spec:
   addOwnerRef: true
   componentKinds:
     - group: core
-      kind: Service
+      kind: ConfigMap
     - group: apps
       kind: Deployment
     - group: core
       kind: Secret
+    - group: core
+      kind: Service
+    - group: core
+      kind: ServiceAccount
+    - group: networking.istio.io
+      kind: VirtualService
     - group: core
       kind: ServiceAccount
     - group: kubeflow.org
@@ -59,7 +65,4 @@ spec:
   selector:
     matchLabels:
       app.kubernetes.io/component: katib
-      app.kubernetes.io/instance: katib-controller
-      app.kubernetes.io/managed-by: kfctl
       app.kubernetes.io/name: katib-controller
-      app.kubernetes.io/part-of: kubeflow
--- a/katib/katib-crds/overlays/application/application.yaml
+++ b/katib/katib-crds/overlays/application/application.yaml
@@ -5,18 +5,8 @@ metadata:
 spec:
   addOwnerRef: true
   componentKinds:
-    - group: core
-      kind: Service
-    - group: apps
-      kind: Deployment
-    - group: core
-      kind: ServiceAccount
-    - group: kubeflow.org
-      kind: Experiment
-    - group: kubeflow.org
-      kind: Suggestion
-    - group: kubeflow.org
-      kind: Trial
+    - group: apiextensions.k8s.io
+      kind: CustomResourceDefinition
   descriptor:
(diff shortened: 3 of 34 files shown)


Jeffwan:
@thesuperzapper Can you help cherry-pick this PR to v1.2-branch? I would like to cut official release Nov 20 morning. Once @andreyvelich sign off the change. I will publish the official release

katib/installs/katib-external-db/kustomization.yaml
```diff
@@ -11,6 +11,12 @@ secretGenerator:
   - name: katib-mysql-secrets
     envs:
       - secrets.env
-commonLabels:
-  app.kubernetes.io/component: katib
-  app.kubernetes.io/name: katib-controller
+patches:
```
andreyvelich, line 14: Why do we have this patch?
I tested locally with `commonLabels` and it also adds label to the Katib mysql Secret.

katib/katib-controller/overlays/application/application.yaml
```diff
@@ -6,11 +6,17 @@ spec:
   addOwnerRef: true
   componentKinds:
     - group: core
-      kind: Service
+      kind: ConfigMap
```
andreyvelich, line 9: Do we need to also add other components from base: https://github.com/kubeflow/manifests/tree/master/katib/katib-controller/base?
ClusterRole, ClusterRoleBinding and PersistentVolumeClaim ?

andreyvelich (commented):
Thank you for doing this @thesuperzapper!
I left few comments.

katib/installs/katib-external-db/kustomization.yaml
thesuperzapper: @andreyvelich I did this because if I put `commonLabels` in this one, it will override the `commonLables` in `../../components/katib-crds`, which would make them incorrect. (As `app.kubernetes.io/name: katib-controller` is the wrong label for the CRDs)

katib/katib-controller/overlays/application/application.yaml
thesuperzapper: @andreyvelich you can NOT put cluster level resources in the Application selector, as applications are Namespaced resources, everything would break.

Re: PersistentVolumeClaim, it depends on if we want to allow users to accidentally delete their PVCs by deleting their Application.

katib/installs/katib-external-db/kustomization.yaml
andreyvelich: Got it, thanks @thesuperzapper

katib/katib-controller/overlays/application/application.yaml
andreyvelich: In that case, I think we can skip them.

andreyvelich:
Thank you for doing this @thesuperzapper!
/lgtm
/cc @gaocegege @johnugeorge

thesuperzapper:
@andreyvelich you have to /approve this

andreyvelich:
> @andreyvelich you have to /approve this

I know, I am just waiting for another review from @gaocegege or @johnugeorge and we can approve it.

thesuperzapper:
@kubeflow/wg-automl-leads anyone want to review this?

gaocegege:
/lgtm

Thanks for your contribution! :tada: :+1:

thesuperzapper:
@gaocegege it needs to be an `/approve`

yanniszark:
@thesuperzapper this is great work! Given the recent discussions on the role of manifests and separation of responsibilities, I believe this PR is better suited for the Katib repo, which maintains its own manifests. The responsibility of manifests is to sync them from there. Does it make sense to open this PR in the Katib repo?

thesuperzapper:
@yanniszark unless there are immediate plans to move all manifests out of `kubeflow/manifests` we should still merge this.

The issue needs to be fixed either way, and any future manifests repo will probably take the YAML which already exists in `kubeflow/manifests`, so its easier if this fix is already applied.

yanniszark:
Closing since we are moving away from using the Application CR:
https://github.com/kubeflow/manifests/issues/1715

PR #1686, opened 2020-12-03
fix jupyter-web-app image tag

thesuperzapper:
This fixes the version of the `jupyter-web-app` image to the version which is correct for Kubeflow 1.2.

Without this, the features from https://github.com/kubeflow/kubeflow/pull/5237 are NOT present.

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`

### DIFF
--- a/jupyter/jupyter-web-app/base/kustomization.yaml
+++ b/jupyter/jupyter-web-app/base/kustomization.yaml
@@ -22,7 +22,7 @@ commonLabels:
 images:
 - name: gcr.io/kubeflow-images-public/jupyter-web-app
   newName: gcr.io/kubeflow-images-public/jupyter-web-app
-  newTag: vmaster-g845af298
+  newTag: vmaster-ge4456300
 # We need the name to be unique without the suffix because the original name is what
 # gets used with patches
 configMapGenerator:
--- a/jupyter/jupyter-web-app/base_v3/kustomization.yaml
+++ b/jupyter/jupyter-web-app/base_v3/kustomization.yaml
@@ -21,7 +21,7 @@ commonLabels:
 images:
 - name: gcr.io/kubeflow-images-public/jupyter-web-app
   newName: gcr.io/kubeflow-images-public/jupyter-web-app
-  newTag: vmaster-g845af298
+  newTag: vmaster-ge4456300
 resources:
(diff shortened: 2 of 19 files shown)


thesuperzapper:
@Jeffwan, I think it's probably fine to merge this, and then cherry pick it into `v1.2-branch`, so that people running `kfctl build ...` in future get the correct version for Kubeflow 1.2.

thesuperzapper:
I also need `/approve` from @berndverst (azure stack) and @animeshsingh (IBM stack), because they hard-code the `jupyter-web-app` image tag into their stack.

berndverst:
@thesuperzapper I don't think there is a reason for the image to be hardcoded in the Azure Stack. Let me work on removing that quickly.

berndverst:
@thesuperzapper Out of curiosity, have we considered having additional tags corresponding to Kubeflow versions we can update?

Assume `gcr.io/kubeflow-images-public/jupyter-web-app:vmaster-g845af298` was also tagged `image: `gcr.io/kubeflow-images-public/jupyter-web-app:kf1.2.0` or something like that. Then we could simply remove the tag now and update the new image as such.

`image: gcr.io/kubeflow-images-public/jupyter-web-app:vmaster-ge4456300`

That would also make it clearer which image is recommended for which Kubeflow version.

Are there any downsides to this?

PatrickXYS:
> @thesuperzapper Out of curiosity, have we considered having additional tags corresponding to Kubeflow versions we can update?
> 
> Assume `gcr.io/kubeflow-images-public/jupyter-web-app:vmaster-g845af298` was also tagged `gcr.io/kubeflow-images-public/jupyter-web-app:kf1.2.0` or something like that. Then we could simply remove the tag now and update the new image as such.
> 
> `image: gcr.io/kubeflow-images-public/jupyter-web-app:vmaster-ge4456300`
> 
> That would also make it clearer which image is recommended for which Kubeflow version.
> 
> Are there any downsides to this?

It depends on wg-notebooks' release process, what falls back is our CD for releasing images.

thesuperzapper:
This is extremely important, I need one of  @adrian555 @animeshsingh @tomcli to approve this ASAP

adrian555 (approved):

PR #1718, opened 2021-01-15
chore: delete outdated metadata google cloudsql overlay

Bobgy:
Resolves kubeflow/gcp-blueprints#182

Delete metadata google cloudsql overlay, because it's no longer our recommended way to deploy cloudsql.
My suggestion would be reusing the same CloudSQL instance between metadata and KFP and it can be implemented in https://github.com/kubeflow/pipelines/issues/4356. It can use workload identity instead of service account key.

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`


Bobgy:
/cc @chensun @capri-xiyue 
Can you help me LGTM?

PR #1719, opened 2021-01-15
chore(kfp): upgrade istio resources -- part 1

### LINKED ISSUE #1707: Upgrade KFP Istio configuration to v1.6
Currently KFP uses Istio RBAC policies that is going to be deprecated in v1.6. The policies needs to be upgraded to supported Istio 1.6 in future KF releases.

james-jwu:
Part of #1707 
Part of https://github.com/kubeflow/pipelines/issues/4768

Switched KFP from using Istio v1alpha1 ServieRole/ServiceRoleBinding to v1beta1 AuthorizationPolicy

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`


PatrickXYS:
Looks like `kfctl_go_unit_test` pod got OOMKilled

PatrickXYS:
```
INFO     root:util.py:72 kfctl exited with error: failed to apply:  (kubeflow.error): Code 500 with message: kfApp Apply failed for kustomize:  (kubeflow.error): Code 500 with message: A
pply.Run : unable to recognize "/tmp/kout135427630": no matches for kind "AuthorizationPolicy" in version "security.istio.io/v1beta1"
```
It's because E2E tests are still using istio-1-3-1, which doesn't have `AuthorizationPolicy` I think

yanniszark:
@PatrickXYS the E2E tests are going to break anyway, due to the incoming refactoring of the manifests repo.
Does it make sense to disable the test for now, until we have a manifests repo in its new form?

DavidSpek (commented):
In multiple places`networking.istio.io/v1alpha3` is used, rather than `networking.istio.io/v1beta1` for the Destination rule. As far as I can tell from https://istio.io/latest/docs/reference/config/networking/virtual-service/ nothing would need to be changed other than the apiVersion.

Bobgy:
Agree, we can break the e2e kubeflow-manifests-presubmit test.

The only remaining change is as suggested by @DavidSpek, we need to update the resource versions for other resources.

james-jwu:
@DavidSpek Istio 1.4 does not support networking.istio.io/v1beta1/DestinationRule. If I update DestinationRule now the product will break. Either we update Istio to v1.5+ (which supports networking.istio.io/v1beta1), or keep networking.istio.io/v1alpha3 and make another update after upgrading to Istio 1.6. I suggest we keep the v1alpha3 version now.

Bobgy:
/lgtm
/approve
Makes sense to me, I believe we can follow up with that after moving KFP manifests back to KFP repo.
Let's get the changes in first.

Bobgy:
Strange, tide still blocks on kubeflow-manifests-presubmit test, even though it's not managed by this prow instance.

@PatrickXYS can you help disable the test temporarily until we can migrate it to support the upgraded manifest structure?

DavidSpek:
@james-jwu @Bobgy If I remember correctly Kubeflow can use up to istio 1.5 already, and issue mentioned and PRs for KFAM and profile controller are for istio 1.6+. So targetting compatibility with istio 1.4 which is already possible seems like a strange choice.

DavidSpek:
/lgtm cancel

Given that the linked issues in for this PR are regarding Istio 1.6+ support, I think it is a mistake to try and have support for Istio 1.4 in this PR. I would assume that, given the linked issues, the target is to use Istio 1.6+ and not 1.4.

yanniszark:
> If I update DestinationRule now the product will break.

What does product refer to in this sentence? I'm trying to understand what will break.

In general, I agree with @DavidSpek that unless there is a compelling reason, we should use `v1beta1` here. Of course, we can leave it for later as @Bobgy says. But I would like to understand the reason.

Bobgy:
Because Kubeflow on GCP is using Anthos Service Mesh 1.4 (compatible with Istio 1.4), we'll work on an upgrade before next release too, but that makes upgrading other resources not possible to test for us, temporarily.

Bobgy:
I updated PR title and removed `resolves xxx` relationship to issues to better reflect the intention.

DavidSpek:
@Bobgy I assumed something like this might be the case. As `networking.istio.io/v1alpha3` is available in Istio 1.6 and 1.8 this shouldn't be a blocking issue and I believe it can move forward. 

/lgtm

PatrickXYS:
@Bobgy  @yanniszark 

I sent this PR to disable tests temporarily. https://github.com/kubeflow/testing/pull/854

yanniszark:
@PatrickXYS is that necessary? Can't we edit the `prow_config.yaml` in this repo?
https://github.com/kubeflow/manifests/blob/master/prow_config.yaml

PatrickXYS:
@yanniszark I think both ways work, we can also comment the prow_config.yaml to temporarily disable tests

yanniszark:
> @yanniszark I think both ways work, we can also comment the prow_config.yaml to temporarily disable tests

Let's do that then. The reason is because we are OWNERS of the kubeflow/manifests repo, so we can achieve our goal without requiring external help from kubeflow/testing.

@Bobgy @james-jwu you can disabled the test in this PR by commenting out this file. Please add a comment on the reason and a link to the PR:
https://github.com/kubeflow/manifests/blob/master/prow_config.yaml

PR #1721, opened 2021-01-18
Adding KFserving and kfp tekton to Openshift Stack

nakfour:
This PR adds KFServing  and KFP tekton to Openshift stack. To test follow these instructions:
KFServing
```
oc new-project kfserving-test
oc create -f https://raw.githubusercontent.com/kubeflow/kfserving/master/docs/samples/v1alpha2/sklearn/sklearn.yaml
oc port-forward --namespace istio-system svc/istio-ingressgateway 8080:80
export SERVICE_HOSTNAME=sklearn-iris.kfserving-test.example.com
export INGRESS_HOST=localhost
export INGRESS_PORT=8080
curl -v -H "Host: ${SERVICE_HOSTNAME}" http://${INGRESS_HOST}:${INGRESS_PORT}/v1/models/sklearn-iris:predict -d @data.json
{"predictions": [1, 1]}
```
**Note:** If your cluster is using a loadbalancer and assigns a global IP address to the ingress gateway service you can also try the curl command explained in these instructions: https://github.com/kubeflow/kfserving/blob/master/docs/OPENSHIFT_GUIDE.md
If your ingress gateway uses nodeport and does not have a global ip address only port forwarding works at this point. Even if the service is assigned a global DNS name, it requires configuring the knative configmap and exposed the knative route for that service. Some documentation is available here: https://github.com/kubeflow/kfserving/tree/master/docs/samples/custom-domain#modify-the-config-domain-configmap.

KFP Pipeline just run the example coin flip test included
**Checklist:**
    1. `cd manifests/tests`
    3. `make test`


animeshsingh (commented):
Thanks @nakfour  qq - what's the stance with KFP-Argo and Seldon vis a vis this? Do the manifests need to be edited to enable one or the other, or its all in one go?

animeshsingh:
cc @moficodes @pvaneck @Tomcli

nakfour:
@animeshsingh at this time we provided an example kfserving (with seldon included) in "kfctl_openshift.v1.2.0.yaml" and an example kfp-tekton (no argo/pipeline) with "kfctl_openshift.master.kfptekton.yaml" . I am open to suggestions, to be honest these example kfdef need re-organization it is very confusing to users. But for ODH, we will explain each example in the operator description page so users know which example to use and how to customize it.

animeshsingh:
yes looks like this is too much code duplication - can't we just have one master kfdef (or two if multi-user is needed) with KFServing and KFP-Tekton commented out, and then the instructions can say that if you need them, uncomment those sections?

nakfour:
@animeshsingh my issue with that is, not only does the user have to uncomment the kfp-tekton but they have to comment out the pipeline app or else it will cause a conflict, correct. Thats why we have two kfdef examples to avoid this issue.

kfdef/kfctl_openshift.master.kfptekton.yaml
```diff
@@ -0,0 +1,111 @@
...
+    name: tektoncd-install
+  - kustomizeConfig:
+      repoRef:
+        name: manifests
+        path: tektoncd/tektoncd-dashboard/base
+    name: tektoncd-dashboard
+  - kustomizeConfig:
+      repoRef:
+        name: manifests
+        path: stacks/ibm/application/kfp-tekton
```
pvaneck, line 106: Perhaps better to make a copy of the `kfp-tekton` app folder under `stacks/openshift/application` instead of using the `ibm` stack one.

nakfour: @pvaneck good catch, actually we did create one under openshift but missed to fix it in kfdef. Done

kfdef/kfctl_openshift.v1.2.0.yaml
```diff
@@ -14,8 +14,13 @@ spec:
   - kustomizeConfig:
       repoRef:
         name: manifests
-        path: stacks/openshift/application/istio-stack
+        path: stacks/openshift/application/istio-1-3-1-stack
```
LaVLaS, line 17: Do you plan to add this `istio-1-3-1-stack` to the `v1.2.0` branch?

nakfour: @LaVLaS good question, right now it will be in master.

nakfour: @LaVLaS based on your comment, I reverted kfctl for v1.2.0 back to original and added kfserving to the new master kfctl. /cc @pvaneck @animeshsingh

pvaneck (commented):
/lgtm

Issue #538, opened 2019-10-16
kustomize syntax "env: params.env" deprecated in next release of kustomize; kubeflow stuck on kustomize 3.2.1


jbrette:
Heads up, just learned the hard way:

the following syntax is not supported anymore (since 10/15/2019)

```yaml
configMapGenerator:
- env: params.env
  name: parameters
````

It needs to be:

```yaml
configMapGenerator:
- envs:
  - params.env
  name: parameters
```

Will have to enforce that the user uses a very specific version of kustomize until the manifests are updated.

yanniszark:
@jbrette thanks for the heads up! 
I think that means that we have to pin to a certain version in kfctl, correct?

thesuperzapper:
@yanniszark, I am now unable to use v1.0.2 of this repo with Kustomize 3.5.4 due to this issue.

jlewi:
Before we close this out it would be good to add a unittest ala
https://github.com/kubeflow/manifests/blob/master/tests/validate_resources_test.go

To verify we don't use the deprecated syntax in any kustomization.yamls

jlewi:
Writing the tests as a follow on to #1204 would be a good first issue for anyone interested

Bobgy:
A problem from pipelines is that, pipelines' own kustomize manifests have tried to be compatible with kustomize v2 in kubectl. We are syncing it into kubeflow/manifests/pipeline/upstream as the base for kubeflow manifest to support easy upgrade later.

However, supporting kustomize latest v3 causes incompatibility for `env` vs `envs`. I'm thinking about a way forward.
Maybe we can write a script using yq to automatically transform to the new syntax when pulling updates from kfp repo. I'll investigate.

Bobgy:
I implemented a simple `sed` command to patch `env` to `envs` when pulling in KFP manifests: https://github.com/kubeflow/manifests/commit/f5150c5f222ce47e87375cb455fc5f51d4dda2e5#diff-0f3a19cf72b902b67aa1e4b5e85f9662. So far this works without problem.

jlewi:
Per comment: https://github.com/kubeflow/website/pull/2118#issuecomment-673603772

It looks like we are still using old style envs in some packages and this is blocking use of newer versions of kustomize.

I think we should add a test to verify we are not using the deprecated syntax as mentioned above so that we can use the latest kustomize.

PR #1736, opened 2021-02-08
Re-add registration flow to centraldashboard via stacks 

tumido:
~~Resolves #~~

1. Reintroduce `REGISTRATION_FLOW` environment variable for the `centraldashboard`  via stacks.
2. Adds a new Openshift stack with this value set to `false`

**Checklist:**
    1. `cd manifests/tests`
    3. `make test`


tumido:
@googlebot I signed it!

PatrickXYS:
> Nope, this is pulling some unrelated changes and failing even when run against master branch

Please make sure you have kustomize <= 3.2.1 as described in the README

https://github.com/kubeflow/manifests/blob/master/README.md

tumido:
@PatrickXYS done, thanks! :+1:

DavidSpek:
@PatrickXYS The README states that kustomize <= 3.2.1 needs to be used until https://github.com/kubeflow/manifests/issues/538 is fixed. Given that the issue is closed, shouldn't the latest Kustomize work and should the README then be updated? If there is still an issue with using the latest Kustomize an issue should be created so this can be addressed and fixed (upstream).

tumido:
Replaced by https://github.com/kubeflow/manifests/pull/1783

PR #1743, opened 2021-02-15
Update OWNERS

animeshsingh:
Has been maintaining IBM/OpenShift manifests. 
Additionally I believe Jeremy is off the project, so removing for now.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -2,7 +2,6 @@ approvers:
   - animeshsingh
   - Bobgy
   - elikatsis
-  - jlewi
   - PatrickXYS
   - StefanoFioravanzo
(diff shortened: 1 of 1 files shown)


animeshsingh:
cc @moficodes

yanniszark:
@animeshsingh thanks for the PR. Since @moficodes is maintaining the IBM/Openshift distribution, they should be added in the relevant OWNERS file. That would be: 
https://github.com/kubeflow/manifests/blob/master/distributions/stacks/openshift/OWNERS
 https://github.com/kubeflow/manifests/tree/master/distributions/stacks/ibm
Could you amend the PR to edit these files instead?

PR #1758, opened 2021-02-25
remove spartakus

### LINKED ISSUE #1757: spartakus-volunter uses incomaptible image
```
  Warning  Failed     10s (x5 over 48s)  kubelet, minikube  Error: container has runAsNonRoot and image has non-numeric user (nobody), cannot verify user is non-root (pod: "spartakus-volunteer-558f8bfd47-v56qd_kubeflow(06a28722-0634-446b-96c0-1e3ee3ea0d54)", container: volunteer)
```
This happens because https://github.com/kubeflow/manifests/blob/master/contrib/spartakus/base/deployment.yaml uses gcr.io/google_containers/spartakus-amd64:v1.1.0

The fix is very simple. You just have to add USER xxxx where xxxx != 0 at the end of the dockerfile.
The spartakus project seems to be not longer maintained https://github.com/kubernetes-retired/spartakus
Is there a new upstream project?

If not we could also just a securitycontext to the deployment
```
kubectl -n kubeflow edit deployment spartakus-volunteer 
...
        image: gcr.io/google_containers/spartakus-amd64:v1.1.0
        imagePullPolicy: IfNotPresent
        name: volunteer
        securityContext:
          runAsUser: 1007550000
...

Please tell me whether i should create a merge request.

juliusvonkohout:
Resolves #1757 

remove every file with a filename that contains the word spartakus
remove spartakus from all current distributions in distributions/kfdef/ (eg no specific version like 1.2.0)
Please tell me whether i should remove it from older releases too.

@PatrickXYS 

**Checklist:**
- [ ] Unit tests pass:
    2. `make test`


PatrickXYS (changes requested):
Let's leave spartakus application manifests alone, and only remove stack manifests' spartakus installation, e.g, 

https://github.com/kubeflow/manifests/blob/master/distributions/stacks/azure/kustomization.yaml#L11

Though I'm not sure if other cloud providers want to delete spartakus or not, but we can first finish PR and wait for their approvals.

Issue #1740, opened 2021-02-09
Step 2: Move manifests development upstream


yanniszark:
Part-of: https://github.com/kubeflow/manifests/issues/1735
Depends-on: https://github.com/kubeflow/manifests/issues/1738

### Background

As part of the work of wg-manifests for 1.3 (https://github.com/kubeflow/manifests/issues/1735), we are moving manifests development in upstream repos. This is the second step of our plan for 1.3.

### Current State

Applications have manifests in two places:
- In their upstream repos.
- In the manifests repo.

This is a table of every KF app in manifests, along with its upstream repo and the folder in the upstream repo where authors currently store manifests:

| Application | Manifests Path | Upstream Path | Upstream Repo |
| --- | --- | --- | --- |
| TFJob Operator | apps/tf-training/upstream | manifests | https://github.com/kubeflow/tf-operator |
| PyTorch Operator | apps/pytorch-job/upstream | manifests | https://github.com/kubeflow/pytorch-operator |
| MPI Operator | apps/mpi-job/upstream | manifests | https://github.com/kubeflow/mpi-operator |
| MXNet Operator | apps/mxnet-job/upstream | manifests | https://github.com/kubeflow/mxnet-operator |
| XGBoost Operator | apps/xgboost-job/upstream | manifests | https://github.com/kubeflow/xgboost-operator |
| Kubebench | apps/kubebench/upstream | manifests | https://github.com/kubeflow/kubebench |
| Notebook Controller | apps/jupyter/notebook-controller/upstream | components/notebook-controller/config | https://github.com/kubeflow/kubeflow |
| Central Dashboard | apps/centraldashboard/upstream | components/centraldashboard/manifests | https://github.com/kubeflow/kubeflow |
| Profiles + KFAM | apps/profiles/upstream | components/profile-controller/config | https://github.com/kubeflow/kubeflow |
| PodDefaults Webhook | apps/admission-webhook/upstream | components/admission-webhook/manifests | https://github.com/kubeflow/kubeflow |
| Jupyter Web App | apps/jupyter/jupyter-web-app/upstream | components/crud-web-apps/jupyter/manifests | https://github.com/kubeflow/kubeflow |
| Katib | apps/katib/upstream | manifests/v1beta1 | https://github.com/kubeflow/katib |
| KFServing | apps/kfserving/upstream | config | https://github.com/kubeflow/kfserving |
[...]

PR #1765, opened 2021-03-15
chore: delete kubeflow/metadata manifests

Bobgy:
Part of #1740 

Remove kubeflow/metadata manifests, MLMD manifests should come from KFP's upstream manifests now.

**Checklist:**
- [ ] Unit tests pass:
    2. `make test`


PatrickXYS (approved):
/lgtm
/approve

/hold 
Hold for others to review

Bobgy:
@yanniszark can you confirm this change for manifest WG?

elikatsis:
Hi @Bobgy,

I'm a bit concerned with purging this manifests in favor of standalone KFP ones. The reason why is that KF distributions have been deploying using these ones which come with a distinct and dedicated database (`overlays/db/` kustomization) whereas the KFP manifests use the MySQL backing KFP itself.

So if we remove these manifests, users that upgrade their clusters will lose their data.

Bobgy:
That's a fair point.

Do you have any suggestions? Move these to KFP upstream?

Bobgy:
@yanniszark @elikatsis with the manifests moved to upstream, shall we proceed with this PR?

yanniszark:
Yes, thank you @Bobgy !
/lgtm
/approve

PR #1767, opened 2021-03-15
update dex and dex-authenticator

DavidSpek:
Updates the manifests to the new `ghcr.io/dexidp/dex` registry and updates dex and `dex-k8s-authenticator` to the latest version.


PatrickXYS:
Does it works fine with manifests binded istio?

/cc @yanniszark

DavidSpek:
I couldn't find the dex-authenticator in my deployment, but I switched the dex deployment to this container and logging in and out still works, so I don't think anything breaking has happened. There is a new API though, but the old one is still there for compatibility and isn't receiving new features.

PR #1768, opened 2021-03-16
Update Knative serving/eventing to 0.17

### LINKED ISSUE #1754: Update Knative version
The current version of Knative is 0.14.3. With Istio being updated soon (https://github.com/kubeflow/manifests/issues/1753), we should also update the versions for Knative serving/eventing to ensure that they work with the new Istio once it lands.

Not sure which version yet, as it will likely depend on the minimum Kubernetes version we decide to support. This issue can be used to track progress on the Knative front.

pvaneck:
Resolves #1754 

- Updates Knative serving and eventing to the latest patch versions of 0.17. (i.e. 0.17.4 for serving and 0.17.9 for eventing)
- Knative 0.17 was chosen to ensure compatibility with K8s 1.16+. An  `upstream` folder was made in respective folders to contain the resource YAMLS that were retrieved from the upstream release assets.
- YAML anchors and aliases are also removed/expanded as they sometimes cause breakages in kustomize v3.9+.
- Top level directory structure is the same, so installation commands will remain the same.

I validated this with Kubeflow 1.3 manifests on K8s 1.18 using Minikube. Was able to spin up a KFServing InferenceService and curl a prediction using the X-Auth-Token header (because of Dex). I also tested Knative Eventing by creating trigger and broker to log Inference requests which worked fine.

Also Validated on OpenShift 4.5 using the Istio 1.9.0 and KFServing manifests in the master branch of this repo.

**Checklist:**
- No unit tests generated as unit test generation is currently broken.


Tomcli:
@pvaneck quick question, is the new manifests recommend to have most stuff in one yaml? I want to understand is there any new practice in the design.

pvaneck:
@Tomcli I do not believe there is any specific recommendation, though it seems like the current practice is to split up by type of resource. In this case, I just thought it would be more straighforward to keep the upstream YAML in a structure that mirrors the YAML files Knative provides for those installing Knative manually. This way, one could presumably just download newer versions of `serving-core.yaml` and `net-istio.yaml`, then put them into the upstream folder if a newer version of Knative with Kubeflow is desired.

common/knative/serving/install/upstream/serving-core.yaml
```diff
@@ -0,0 +1,1615 @@
...
+    # These sample configuration options may be copied out of
+    # this example block and unindented to be in the data block
+    # to actually change the configuration.
+
+    # List of repositories for which tag to digest resolving should be skipped
+    registriesSkippingTagResolving: "ko.local,dev.local"
+
+    # ProgressDeadline is the duration we wait for the deployment to
+    # be ready before considering it failed.
+    progressDeadline: "120s"
```
yuzisun, line 384: Can you help bump this to 600s to match what we have in KFServing [e2e test](https://github.com/kubeflow/kfserving/blob/master/test/scripts/run-e2e-tests.sh#L102), sometimes the models are big and takes longer than 2 min to start the containers.

pvaneck: Sure, I added a patch to set `progressDeadline`  to `600s`.

pvaneck:
Validated this with the Kubeflow 1.3 manifests on K8s 1.18 using Minikube. Was able to spin up a KFServing InferenceService and curl a prediction using the X-Auth-Token header (because of Dex).

pvaneck:
Validated Knative Eventing install by creating trigger and broker to log Inference requests on same K8s 1.18 minikube environment as above.

Validated on OpenShift 4.5 using the Istio 1.9.0 and KFServing manifests in the master branch of this repo.

@yanniszark Would be great to have this in for the Kubeflow 1.3 release if possible. The updated Knative brings additional configuration and features that would be useful in the context of KFServing.

Issue #1775, opened 2021-03-19
Release 1.3 - Sync manifests from upstream repos


yanniszark:
In order to cut the 1.3 release, we need to sync manifests from the upstream repos. For this reason, the manifests wg needs to gather the git revisions from each upstream repo, in order to sync them here.

| Application | Manifests Path | Upstream Path | Upstream Repo | Revision |
| --- | --- | --- | --- | --- |
| TFJob Operator | apps/tf-training/upstream | manifests | https://github.com/kubeflow/tf-operator | `v1.1.0 ` |
| PyTorch Operator | apps/pytorch-job/upstream | manifests | https://github.com/kubeflow/pytorch-operator | `v0.7.0` |
| MPI Operator | apps/mpi-job/upstream | manifests | https://github.com/kubeflow/mpi-operator | `b367aa55886d2b042f5089df359d8e067e49e8d1` |
| MXNet Operator | apps/mxnet-job/upstream | manifests | https://github.com/kubeflow/mxnet-operator | `v1.1.0` |
| XGBoost Operator | apps/xgboost-job/upstream | manifests | https://github.com/kubeflow/xgboost-operator | `v0.2.0` |
| Notebook Controller | apps/jupyter/notebook-controller/upstream | components/notebook-controller/config | https://github.com/kubeflow/kubeflow | `origin/v1.3.0-rc.0` |
| Central Dashboard | apps/centraldashboard/upstream | components/centraldashboard/manifests | https://github.com/kubeflow/kubeflow | `origin/v1.3.0-rc.0` |
| Profiles + KFAM | apps/profiles/upstream | components/profile-controller/config | https://github.com/kubeflow/kubeflow | `origin/v1.3.0-rc.0` |
| PodDefaults Webhook | apps/admission-webhook/upstream | components/admission-webhook/manifests | https://github.com/kubeflow/kubeflow | `origin/v1.3.0-rc.0` |
| Jupyter Web App | apps/jupyter/jupyter-web-app/upstream | components/crud-web-apps/jupyter/manifests | https://github.com/kubeflow/kubeflow | `origin/v1.3.0-rc.0` |
| Volumes Web App | apps/volumes-web-app | components/crud-web-apps/volumes/manifests | https://github.com/kubeflow/kubeflow |  `origin/v1.3.0-rc.0` |
[...]

PR #1783, opened 2021-03-30
feat: Reintroduce REGISTRATION_FLOW variable for central dashboard

tumido:
Replaces: https://github.com/kubeflow/manifests/pull/1736 (I need to keep the original branch of the previous PR since I'm deploying from it, hence I'm creating a new PR, sorry)

Add the `REGISTRATION_FLOW` for central dashboard and makes the flag available in stacks

**Checklist:**
- [ ] Unit tests pass:  I'm unable to execute the `make generate-changes-only` due to `Error: accumulating resources: accumulating resources from '../../../pipeline/pipelines-ui/overlays/istio/virtual-service.yaml': open /home/tcoufal/Programming/AI-CoE/kubeflow/manifests/pipeline/pipelines-ui/overlays/istio/virtual-service.yaml: no such file or directory`, the tests were not yet adjusted to the new directory schema probably.

### DIFF
--- a/apps/centraldashboard/upstream/overlays/stacks/deployment_kf_config.yaml
+++ b/apps/centraldashboard/upstream/overlays/stacks/deployment_kf_config.yaml
@@ -18,3 +18,8 @@ spec:
                 configMapKeyRef:
                   name: kubeflow-config
                   key: userid-prefix
+            - name: REGISTRATION_FLOW
+              valueFrom:
+                configMapKeyRef:
+                  name: kubeflow-config
+                  key: registration-flow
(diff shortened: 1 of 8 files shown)


DavidSpek:
I think you are looking for this: https://github.com/kubeflow/kubeflow/blob/98fc2dcb57fe0a5311abed28510a4df58a506971/components/centraldashboard/manifests/base/kustomization.yaml#L59

tumido:
@DavidSpek you are 100% correct! Does that mean that this spec will get sync here as part of https://github.com/kubeflow/manifests/issues/1775 so it is available to be used in stacks?

DavidSpek:
Yes, the upstream repositories are responsible for their manifests. This repository will just pull from the upstream ones, so if something is not correct in a manifest it needs to be addressed in the upstream repository.

tumido:
Thank you @DavidSpek for your explanation. Learning something new every day. :+1:

That means this PR can be closed then :smile:

PR #1786, opened 2021-04-01
Fix default dex, authservice config

yanniszark:
Fix default dex, authservice config for 1.3


elikatsis:
Thanks @yanniszark ! I'll look into this

PR #1788, opened 2021-04-01
istio: Change Istio IngressGateway Service to NodePort

yanniszark:
Change Istio IngressGateway Service to NodePort as it was for the previous installation.


PatrickXYS (approved):
/lgtm

PR #1789, opened 2021-04-02
oidc-authservice: fix secret generator envs

DavidSpek:
Fix a small issue in the oidc-authservice manifests where `env: secret_params.env` is used instead of 
```yaml
envs:
  - secret_params.env
```


DavidSpek:
@yanniszark related to [your comment](https://github.com/kubeflow/manifests/pull/1805#issuecomment-814859695), this fix seems necessary for newer versions of kustomize. However, it also ensures consistency even within this file.

PR #1795, opened 2021-04-03
Fix using newer kustomize for knative-eventing-install

DavidSpek:
While testing with the latest version of kustomize the build for `common/knative/knative-eventing-install/base` fails due to the usage of `&ref_0` and `*ref_0` in the `matchLabels` and `labels` sections respectively. While investigating I compared the `deployment.yaml` of knative-eventing to that of knative-serving and noticed in the knative-serving manifests that the labels were explicitly stated. In the spirit of consistency and making the manifests usable with the latest version of kustomize this PR edits the `deployment.yaml` of knative-eventing to use the same syntax as that of knative-serving. 

/cc @yanniszark @PatrickXYS

### DIFF
--- a/common/knative/knative-eventing-install/base/cluster-role.yaml
+++ b/common/knative/knative-eventing-install/base/cluster-role.yaml
@@ -315,7 +315,7 @@ rules:
   - endpoints
   - events
   - serviceaccounts
-  verbs: &ref_0
+  verbs:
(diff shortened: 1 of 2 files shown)


yanniszark:
@DavidSpek thanks for your contribution!
/lgtm
/approve

PR #1809, opened 2021-04-07
Dex: Fix user must be valid email address

DavidSpek:
To be able to add contributors to your namespace the `name` value in the profile CR must be a valid email address. Currently, the default is set to `user` instead of `user@kubeflow.org`, which will break adding it as a contributor to other namespaces. This PR fixes the dex config so the `email` value is a "valid" email address and the associated Profile CR contains this email address. 

/cc @yanniszark 
Also /cc @berndverst as I think you might run into this as well as you are going through the deployment manifests now.

**Checklist:**
- [ ] Unit tests pass:
    2. `make test`

### DIFF
--- a/common/dex/base/config-map.yaml
+++ b/common/dex/base/config-map.yaml
@@ -18,7 +18,7 @@ data:
       skipApprovalScreen: true
     enablePasswordDB: true
     staticPasswords:
-    - email: user@kubeflow.org
+    - email: user@example.com
       hash: $2y$12$4K/VkmDd1q1Orb3xAt82zu8gk7Ad6ReFR4LCP9UeYE90NLiN9Df72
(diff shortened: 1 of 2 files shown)


yanniszark:
@DavidSpek thanks for the PR. Overall I agree.
I propose the following changes:
- Change user to `user@example.com`, which is a standard domain for examples (see example.com)
- Change Profile name to `kubeflow-user-example-com` accordingly

DavidSpek:
@yanniszark the profile name is equal to the namespace. When the profile controller creates this it is the email with the domain removed. As I think `user` is a bit to generic for a namespace, I used the `kubeflow-user@example.com` email address with the appropriate username and namespace.

Issue #1812, opened 2021-04-08
Using external Auth (APPID) get 403 on notebook and tensorboard creation


moficodes:
issue from kubeflow: https://github.com/kubeflow/kubeflow/issues/5803

The `APP_SECURE_COOKIES` is not set and is set [in code](https://github.com/kubeflow/kubeflow/blob/master/components/crud-web-apps/common/backend/kubeflow/kubeflow/crud_backend/settings.py#L3) to true by default.

This means until we setup https we wont be able to create notebook or tensorboard on non dex auth. 

The error looks like 
```
Could not find CSRF cookie XSRF-TOKEN in the request. http://10.14.0.121/jupyter/api/namespaces/kubeflow-user/notebooks
```

We should make this into a `params.env` variable and make it configurable.

DavidSpek:
@moficodes Yeah I was expecting a lot of people would run into this. For my distribution I fixed this problem by having the `kubeflow-self-signing-issuer` used by the components to create certificates for the istio ingress-gateway as well and defaulting to HTTPS. 

@yanniszark Given that many people are going to run into this issue and most likely create a lot of issues, should I open up a PR for defaulting the ingress gateway to HTTPS with a certificate from the `kubeflow-self-signing-issuer`?

moficodes:
@DavidSpek thanks.

yanniszark:
@moficodes @DavidSpek I see two distinct issues here:

- One for making changes to individual app settings.
- One for using a self-signed certificate by default.

Let me answer each one:

> - One for making changes to individual app settings.

@moficodes to make changes to the default Tensorboard Web App (or other web app), you can use standard kustomize patches. 
Here are some example:
https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/patchesstrategicmerge/
https://github.com/kubeflow/manifests/blob/master/apps/profiles/upstream/overlays/kubeflow/patches/kfam.yaml

If you need some help on using kustomize patches, feel free to ping me and we can arrange a quick meet.
Please tell me if anything is not clear.

> - One for using a self-signed certificate by default.

For this issue, I am inclined to decide against including such a thing.
Exposing Services correctly to the outside world is a deeply platform/environent-specific process and each distribution should include instructions if they want to guide their users through exposing their Kubeflow installations correctly with HTTPS.
This is the reason why we have only included the port-forward method in the installation guide.
[...]

PR #1819, opened 2021-04-08
Make ingress gateway HTTPS by default

DavidSpek:
Closes: https://github.com/kubeflow/manifests/issues/1812 https://github.com/kubeflow/kubeflow/issues/5803

Currently the Jupyter, Tensorboards and Volumes Web Apps are setup to require secure cookies. However, as the istio ingress-gateway that Kubeflow uses is setup for HTTP, by default users will run into errors causing these web apps to not work. To mitigate this (and promote the use of HTTPS), this PR sets the ingress-gateway to use HTTPS by default with a certificate created by the `kubeflow-self-signing-issuer` that other kubeflow components already use.

/cc @yanniszark 

**Checklist:**
- [x] Unit tests pass:
    2. `make test`

### DIFF
--- a/common/istio-1-9-0/kubeflow-istio-resources/base/certificate.yaml
+++ b/common/istio-1-9-0/kubeflow-istio-resources/base/certificate.yaml
@@ -1,0 +1,13 @@
+apiVersion: cert-manager.io/v1alpha2
+kind: Certificate
+metadata:
+  name: istio-ingressgateway-certs
+  namespace: istio-system
+spec:
+  secretName: istio-ingressgateway-certs
+  issuerRef:
+    name: kubeflow-self-signing-issuer
+    kind: ClusterIssuer
+  commonName: kubeflow.example.com
+  dnsNames:
+    - kubeflow.example.com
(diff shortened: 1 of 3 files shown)


moficodes (approved):
/lgtm

DavidSpek:
@moficodes You're missing a `/`, it's `/lgtm` :).

moficodes:
I doubt my `/lgtm` does anything. But fixed it. 😄
Thanks once again.

thesuperzapper:
@yanniszark we need you to approve (if you agree)

yanniszark:
@thesuperzapper I am not convinced that this change is something we want to do. We can continue the discussion in https://github.com/kubeflow/manifests/issues/1812 and open a PR (or not) when we reach a conclusion. I believe @moficodes is unblocked to continue the integration with IBM, so I will come back to this after making a pass at other blocking issues.

DavidSpek:
@yanniszark [Here](https://github.com/kubeflow/kubeflow/issues/5803#issuecomment-817310731) is another example of a user that is struggling with this problem even after following the instructions to setup HTTPS. If this was already part of the kustomize manifests he would not have run into this problem.

yanniszark:
> @yanniszark Here is another example of a user that is struggling with this problem even after following the instructions to setup HTTPS. If this was already part of the kustomize manifests he would not have run into this problem.

Agreed, users will try to export Kubeflow over HTTP with NodePort / LoadBalancer / Ingress, that much is evident. For these reasons, I proposed: https://github.com/kubeflow/manifests/issues/1812#issuecomment-818117944

Let's take the discussion into the issue if you agree.

Issue #1234, opened 2020-06-08
Upgrade cert-manager to 0.15; add probes to deal with readiness failures during deployment


jlewi:
It looks like we are currently on cert-manager 0.11. The latest is 0.15. It looks like 0.12 added readiness and liveness probes for the webhook.
https://github.com/jetstack/cert-manager/pull/2245

This might fix issues during deployment. Currently, if we try to create cert-manager resources (e.g. certificates for the kubeflow admission-webhook) then we will get an error like

```
Error from server (InternalError): error when creating ".build/kubeflow-apps/cert-manager.io_v1alpha2_certificate_admission-webhook-cert.yaml": Internal error occurred: failed calling webhook "webhook.cert-manager.io": the server is currently unable to handle the request
```

Adding kubectl wait statements like the following doesn't work

```
kubectl --context=$(KFCTXT) -n cert-manager wait --for=condition=Available --timeout=600s deploy cert-manager-webhook
	kubectl --context=$(KFCTXT) -n cert-manager wait --for=condition=Available --timeout=600s deploy cert-manager
	kubectl --context=$(KFCTXT) -n cert-manager wait --for=condition=Available --timeout=600s deploy cert-manager-cainjecto
```

This is likely because our deployment for the webhook doesn't include readiness and liveness probes.
https://github.com/jetstack/cert-manager/blob/master/deploy/charts/cert-manager/templates/webhook-deployment.yaml#L68

I think if we upgrade to 0.15 then we should be able to add readiness and liveness probes.

@krishnadurai looks like you did the upgrade to 0.11. Do you think it would be worth trying to upgrade to 0.15 for KF 1.1?

krishnadurai:
@jlewi it looks like this will help the situation with cert-manager readiness. We should look to upgrade to 0.15.
Can kfctl leverage waiting for availability to deploy cert-manager resources?

jlewi:
@krishnadurai in principle kfctl can run any arbitrary code.

The approach we are taking with GCP blueprints is much simpler. We just use `kubectl`
https://github.com/kubeflow/gcp-blueprints/blob/5734f2693f89ea45fa0c404bd3330caf70e4cba8/kubeflow/Makefile#L203

So we can call `kubectl wait`. I'm hoping that if the deployment includes readiness probes then `kubectl wait` will wait until its actually ready to serve.

PatrickXYS:
[...]

PR #1820, opened 2021-04-08
Update cert-manager manifests to 1.3.1 version

DavidSpek:
Closes: https://github.com/kubeflow/manifests/issues/1234 https://github.com/kubeflow/manifests/issues/1838

This PR updates the cert-manager manifests to the latest version 1.3.1, as the current manifests are from October 2019 and there have probably been many fixes since v0.11. The above mentioned issue is related to OpenShift, according to the cert-manager [documentation](https://cert-manager.io/docs/installation/openshift/#installing-with-regular-manifests) the latest (the one in this PR) manifest can also be deployed on OpenShift so this likely resolves the above mentioned issue.

/cc @yanniszark @nakfour 

**Checklist:**
- [x] Unit tests pass:
    2. `make test`


common/cert-manager/cert-manager/base/kustomization.yaml
```diff
@@ -1,41 +1,16 @@
...
-- cluster-role-binding.yaml
-- cluster-role.yaml
-- deployment.yaml
-- mutating-webhook-configuration.yaml
-- service-account.yaml
-- service.yaml
-- validating-webhook-configuration.yaml
-commonLabels:
-  kustomize.component: cert-manager
+- https://github.com/jetstack/cert-manager/releases/download/v1.3.0/cert-manager.yaml
```
zijianjoy, line 5: Should we use the download URL for cert-manager, or storing the actual files as before? It is easy to write kustomization if we can access these files in local IDE.

If we are going to use the external github link, should we move cert-manager to `contrib/` parent directory instead? Because it is third party tool.

DavidSpek: Personally I believe users shouldn't need to edit the actual cert-manager manifest, but if for some reason they need to customize it they should add an overlay or a PatchStrategicMerge, etc. I don't think the original YAML should be touched though, so linking to it upstream removes clutter and makes it easy to see what version is being used and to upgrade it.

This is also the setup I am using for all the Kubeflow components in the Argo CD deployment, which makes it a lot easier for users as they don't need to deal with the huge amount of YAML. https://github.com/kubeflow-onprem/ArgoFlow

zijianjoy: Thank you David for the explanation! I understand the benefit. I agree that there are pros and cons choosing one approach versus the other. So I will leave it to you to decide.

Besides, I cannot access this link https://github.com/kubeflow-onprem/ArgoFlow, is this the right link or is it private repo?

cc @Bobgy

DavidSpek: @zijianjoy Yeah the repo needed to change homes suddenly. It's now https://github.com/argoflow/argoflow.

zijianjoy:
I tried to install this cert-manager

```
kustomize build --load-restrictor LoadRestrictionsNone -o .build/cert-manager ./common/cert-manager/kubeflow-issuer
kubectl apply -f ./cert-manager
```

But it failed with the following error:

```
Error from server (InternalError): error when creating ".build/cert-manager/cert-manager.io_v1alpha2_clusterissuer_kubeflow-self-signing-issuer.yaml": Internal error occurred: failed calling webhook "webhook.cert-manager.io": Post https://cert-manager-webhook.cert-manager.svc:443/mutate?timeout=10s: x509: certificate signed by unknown authority
```

How do I resolve this installation issue?

DavidSpek:
@zijianjoy Not sure why you are using the `-o` flag, the load restrictor flag is definitely not necessary. To build the manifests from this PR you just need to run `kustomize build <cert_manager_directory>`.

zijianjoy:
> @zijianjoy Not sure why you are using the `-o` flag, the load restrictor flag is definitely not necessary. To build the manifests from this PR you just need to run `kustomize build <cert_manager_directory>`.

The `-o` flag means `output`, which allows me to check the resource file I am going to apply.

You can ignore `loadRestrictor` flag, it doesn't have effect in this case.

Would you like to be specific on <cert_manager_directory> path? I don't see a kustomization.yaml file under ./common/cert-manager path

DavidSpek:
Sorry, on my phone so couldn't check very easily. Looks the issuer and cert-manager folder seem to be separate here. So for cert-manager the folder is `./common/cert-manager/base`.

zijianjoy:
> Sorry, on my phone so couldn't check very easily. Looks the issuer and cert-manager folder seem to be separate here. So for cert-manager the folder is `./common/cert-manager/base`.

Right, I think the `kubeflow-issuer/kustomization.yaml` depends on `overlays/self-signed/kustomization.yaml`, which in turns depends on `base`. So cert-manager and issuer are not separated in this PR. I am not sure if it is the root cause of issue, but `kubeflow/gcp-blueprints` and the current `kubeflow/manifests` separate _kubeflow-issuer_ and _cert-manager_ targets.

zijianjoy:
@DavidSpek Thank you David! Please see my comment in https://github.com/kubeflow/gcp-blueprints/pull/212#issuecomment-823482800 for a successful cert-manager and kubeflow-issuer deployment.

zijianjoy:
/lgtm

Thank you David for the change and cert-manager upgrade!

zijianjoy:
NIT: You might want to update README after your upgrade: https://github.com/kubeflow/manifests#cert-manager

DavidSpek:
Good catch, will need to see what the plan is to implement this after the release.

berglh (commented):
[cert-manager upgrading 0.16 to 1.0](https://cert-manager.io/docs/installation/upgrading/upgrading-0.16-1.0/)

I might be mistaken, but I think all the `apiVersions` for cert-manager resources need to reference `cert-manager.io/v1` instead of `cert-manager.io/v1alpha2` in versions greater than `cert-manager v1.0` which I think is at a minimum in `cluster-issuer.yaml`.

DavidSpek:
@berglh From what I have noticed the `cert-manager.io/v1alpha2` api still works with the latest version of cert-manager, but indeed the upstream component manifests should also be updated to use the latest version of the API.

yanniszark:
@DavidSpek thanks for the PR. A few notes:
- I have only managed to tests and confirm cert-manager 1.3.1, because 1.4 was added recently. As such, let's include 1.3.1 for now and update further in later releases.
- I have made a small patch to explicitly disable `preserveUnknownFields`, otherwise upgrade with 'kubectl apply' fails when upgrading.
- I have moved the kubeflow-issuer into its own kustomization, same as other software in the repo.
- I have updated README and the example kustomization.

Please take a look at https://github.com/arrikto/kubeflow-manifests/tree/update-cert-manager and push it to this PR if you agree, in order to move forward.

DavidSpek:
@yanniszark I've made the suggested changes, but kept cert-manager version 1.4 instead of 1.3.1. I've tested this extensively with multiple fresh cluster deployments, and many large organization are using this in conjunction with [ArgoFlow](https://argoflow.ai/). Looking through the [release notes](https://github.com/jetstack/cert-manager/releases) there aren't any breaking changes that would affect us, but there are bug fixes and new features that might be important for users. Another good reason to use 1.4.0 instead of 1.3.1, is that it enabled  `RunAsNonRoot` by default, which should help improve comparability for OpenShift.
As such, I don't see a reason to not upgrade cert-manager to the latest version. We should avoid having a reluctance to upgrade components, since this is what caused the 1.3.0 release to have an incompatible cert-manager version for KFServing despite me suggesting to update this component months before, and during the RC period of, the previous release.

yanniszark:
@DavidSpek I'm sorry but I am not going to include something I have not tested.
Review testing allows us to catch bugs early, such as the patch to support the cert-manager upgrade, which was caught during review testing. 
Because this is was a change close to the release, it couldn't make it in 1.3.1.
I am very open to upgrading after that, which should be even easier since the versions are so close.
Would it be possible to update your PR to include 1.3.1 as it originally did, since that's the version I tested in the review?

DavidSpek:
I'll change this PR to 1.3.1. However, I would like to highlight here that the inclusion of manifest features or updates should not, and cannot rely on a single person testing such an update or feature for hopefully obvious reasons. Also, a single person/entity should not be able to make these decisions independently, as this goes against neutral governance within the community.

DavidSpek:
@yanniszark I just pushed the commit to change to cert-manager 1.3.1.

yanniszark:
> I'll change this PR to 1.3.1. I would like to know more about what review testing is done and what procedure is followed, since I also wonder how the numerous issues and incompatibilities surrounding KFServing and KNative were not caught. It seems like the testing procedure needs to be improved.
> However, I would also like to highlight here that the inclusion of manifest features or updates should not, and cannot rely on a single person testing such an update or feature for hopefully obvious reasons. Also, a single person/entity should not be able to make these decisions independently, as this goes against neutral governance within the community.

Thanks a lot! Yes, our testing processes should be improved and we will be improving as we evolve in the future. Note that before 1.3, where the Manifests WG didn't even exist, the process was even worse to start with. And we have gone a long way since then.
Let's continue the discussion on this in a separate issue.

Continuing on the topic of the review, there are still some differences. Mainly:
- The PR still sets the 1.4 images. Could you remove the images part, since it's not needed because they correct images are already set. Let's keep this simple.
- The PR does not have the comments on the kube-system namespace and the use of `preserveUnknownFields`.

Here is the full diff for convenience:
```diff
diff --git a/common/cert-manager/cert-manager/base/kustomization.yaml b/common/cert-manager/cert-manager/base/kustomization.yaml
index 4361a8450..76e89b747 100644
--- a/common/cert-manager/cert-manager/base/kustomization.yaml
+++ b/common/cert-manager/cert-manager/base/kustomization.yaml
@@ -3,16 +3,20 @@ kind: Kustomization

 resources:
 # Manifests downloaded from:
-# https://github.com/jetstack/cert-manager/releases/download/v1.3.1/cert-manager.yaml
+# - https://github.com/jetstack/cert-manager/releases/download/v1.3.1/cert-manager.yaml
 - cert-manager.yaml

+images:
+- name: quay.io/jetstack/cert-manager-controller
+  newName: quay.io/jetstack/cert-manager-controller
+  newTag: v1.4.0
+- name: quay.io/jetstack/cert-manager-cainjector
+  newName: quay.io/jetstack/cert-manager-cainjector
+  newTag: v1.4.0
+- name: quay.io/jetstack/cert-manager-webhook
+  newName: quay.io/jetstack/cert-manager-webhook
+  newTag: v1.4.0

-# XXX: Do NOT use the namespace transformer, as cert-manager defines resources
-# in two namespaces, 'cert-manager' and 'kube-system'.
-# For more information, see https://github.com/jetstack/cert-manager/issues/4102.
-
-# Patch upstream manifests to explicitly disable 'preserveUnknownFields',
-# otherwise upgrade with 'kubectl apply' fails.
 patches:
 - path: patches/crd-preserve-unknown-fields.yaml
   target:
```

Could you make these final changes?

yanniszark:
@DavidSpek kind ping

DavidSpek:
@yanniszark Thanks for the ping, I missed your last message. Looking into it now.

DavidSpek:
@yanniszark Just pushed the commit to fix the tags and add the comment about `preserveUnknownFields`.

yanniszark:
@DavidSpek thanks a lot!
Is it possible to remove the imageTransformer and keep the kustomization simpler?
Also (minor), could you remove the '-' in the cert-manager link comment?

Including the desired diff for convenience:
```diff
diff --git a/common/cert-manager/cert-manager/base/kustomization.yaml b/common/cert-manager/cert-manager/base/kustomization.yaml
index 7d1bf5f0e..4361a8450 100644
--- a/common/cert-manager/cert-manager/base/kustomization.yaml
+++ b/common/cert-manager/cert-manager/base/kustomization.yaml
@@ -3,19 +3,9 @@ kind: Kustomization

 resources:
 # Manifests downloaded from:
-# - https://github.com/jetstack/cert-manager/releases/download/v1.3.1/cert-manager.yaml
+# https://github.com/jetstack/cert-manager/releases/download/v1.3.1/cert-manager.yaml
 - cert-manager.yaml

-images:
-- name: quay.io/jetstack/cert-manager-controller
-  newName: quay.io/jetstack/cert-manager-controller
-  newTag: v1.3.1
-- name: quay.io/jetstack/cert-manager-cainjector
-  newName: quay.io/jetstack/cert-manager-cainjector
-  newTag: v1.3.1
-- name: quay.io/jetstack/cert-manager-webhook
-  newName: quay.io/jetstack/cert-manager-webhook
-  newTag: v1.3.1
```

DavidSpek:
@yanniszark Just pushed the commit. In the future, the images section should be added I think since it is used by dependency automation and it allows users to easily switch the registry if they are in an airgapped environment.

yanniszark:
Thanks!
/lgtm
/approve

PR #1826, opened 2021-04-09
Make note of knative-eventing being optional in README 

pvaneck:
Only Knative Serving is required for KFServing to work. Knative Eventing is used for specific logger/metrics tracking use cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,13 +128,17 @@ kustomize build --load_restrictor=none common/oidc-authservice/base | kubectl ap
 
 Knative is used by the KFServing official Kubeflow component.
 
-Install Knative:
+Install Knative Serving:
 ```sh
 kustomize build --load_restrictor=none common/knative/knative-serving-crds/base | kubectl apply -f -
 kustomize build --load_restrictor=none common/knative/knative-serving-install/base | kubectl apply -f -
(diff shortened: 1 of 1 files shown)


yanniszark:
Thanks a lot @pvaneck. I have some questions about the PR.

First of all, since the example kustomization is about installing all Kubeflow components and common services, should we leave eventing inside?

Then, we can use the detailed component-by-component guide to describe its optional nature. In the docs, instead of commenting out the optional part, I would suggest making them two individual code blocks. Something like the following:

Install Knative Serving:

```sh
kustomize build --load_restrictor=none common/knative/knative-serving-crds/base | kubectl apply -f -
kustomize build --load_restrictor=none common/knative/knative-serving-install/base | kubectl apply -f -
kustomize build --load_restrictor=none common/istio-1-9-0/cluster-local-gateway/base | kubectl apply -f -
```

Optionally, you can install Knative Eventing which is needed for inference request logging:
```sh
kustomize build --load_restrictor=none common/knative/knative-eventing-crds/base | kubectl apply -f -
kustomize build --load_restrictor=none common/knative/knative-eventing-install/base | kubectl apply -f -
```

What do you think?

pvaneck:
@yanniszark That sound reasonable. I will follow your suggestion about the code blocks and will leave eventing inside the example reference deployment.

PR #1827, opened 2021-04-09
knative: Remove use of yaml ref

yanniszark:
Remove use of yaml ref as it is not supported by later kustomize
versions.

This finishes the work started by:
https://github.com/kubeflow/manifests/pull/1795


yanniszark:
@DavidSpek could you take a quick look since you submitted the original PR?

DavidSpek:
@yanniszark looks good to me. Should we maybe extend this PR with a search through all the manifests to remove them everywhere if they are still in other places?

yanniszark:
> @yanniszark looks good to me. Should we maybe extend this PR with a search through all the manifests to remove them everywhere if they are still in other places?

Right, with a quick look, I could find an extra use in Tekton's manifests. However, I think we should merge this PR and then let the maintainers of the tekton kustomization, which is under contrib, know about the incompatibility.

DavidSpek:
That’s good to know. I think the Tekton maintainers should take care of that. Should I lgtm this PR?

yanniszark:
Yeap!

Issue #1798, opened 2021-04-05
Distributions: Readiness for 1.3 Kubeflow Release


yanniszark:
### Problem Statement

We have a 6 step plan for releasing Kubeflow 1.3: https://github.com/kubeflow/manifests/issues/1777
This issue is for the 6th step: distributions use instructions from wg-manifests for how to use kustomizations and create distributions for 1.3.

| Distribution | Owners | Readiness for 1.3 |
| - | - | - |
| Arrikto EKF | @kubeflow/arrikto | ✔️  |
| Arrikto MiniKF | @kubeflow/arrikto | ✔️  |
| Azure | @kubeflow/azure | :x: |
| AWS | @kubeflow/aws | ✔️  |
| Charmed Kubeflow | @RFMVasconcelos | ✔️  |
| Google Cloud | @kubeflow/google | ✔️ |
| IBM | @kubeflow/ibm | ✔️  |
| Kubeflow on MicroK8s | @RFMVasconcelos | ✔️  |
| Kubeflow Operator | @kubeflow/red-hat | ✔️  |
| Kubeflow with Argo CD | @DavidSpek | ✔️  |
| kfctl_istio_k8s | | Support dropped, remove from docs |
| kfctl_istio_dex | | Support dropped, remove from docs |
| Openshift | @kubeflow/red-hat | ✔️  |

Distributions should have owners and update their process/docs for 1.3 installations.

Distributions can use the instructions for installing Kubeflow 1.3 components and common services, provided by wg-manifests, to perform their integration:
https://github.com/kubeflow/manifests/tree/v1.3.0-rc.0#readme

I would like to ask all distribution owners to check-in in this issue to confirm that they will support their distributions for Kubeflow 1.3.

Current Issues:
- [x] https://github.com/kubeflow/kubeflow/issues/5813
- [x] https://github.com/kubeflow/manifests/issues/1810
- [x] https://github.com/kubeflow/kubeflow/issues/5821
- [x] https://github.com/kubeflow/kubeflow/issues/5814
- [x] https://github.com/kubeflow/manifests/issues/1812

DavidSpek:
Guess you can add Kubeflow On-Prem/ArgoFlow/Argo CD to this list with https://github.com/kubeflow-onprem/ArgoFlow and status Ready. 
/cc @jtfogarty @tbaums @RFMVasconcelos

nakfour:
[...]

PR #1832, opened 2021-04-09
Add AWS 1.3 distribution

PatrickXYS:
Helps #1798

AWS-EKS manifest for kubeflow 1.3

**Checklist:**
- [ ] Unit tests pass:
    2. `make test`


karlschriek:
@kubeflow/aws I have a few questions about this release. When we deploy to AWS we offload as much as possible to managed AWS services as we possibly can, meaning in particular:

1. Using Cognito for user pool management
2. Using RDS for all metadata storage (pipelines, metadb, cachedb, katibdb) instead of an on-cluster MySQL db
3. Using S3 for all pipeline artifact storage instead of using Minio for on-cluster storage
4. Using Secret Manager in combination with https://github.com/external-secrets/kubernetes-external-secrets in order to manage secrets (for example RDS database credentials)
5. Using IRSA as far as possible

From what I can tell, this distribution aims to roll out on EKS but does not attempt to integrate with any further AWS services. Is that the intention?

soleares:
I thought that one of the goals in 1.3 was to move away from kfdef and use straight kustomize. See: https://github.com/kubeflow/manifests/issues/1062. Is kfdef going to continue to be used for distributions like this?

karlschriek:
> I thought that one of the goals in 1.3 was to move away from kfdef and use straight kustomize. See: #1062. Is kfdef going to continue to be used for distributions?

As far as I know the idea is to make the manifests independent of kfdef/kfctl and to cut the distributions loose to do what they want. Eventually the "distributions" folder in the manifests should disappear as well.

Whether kfdef/kfctl continues to exist I guess will depend on whether distributions continue to adopt them. I personally would prefer the AWS distro to be pure kustomize. We would then wrap that within ArgoCD as Gitops operator. With kfdef/kfctl inbetween that is a bit difficult.

DavidSpek:
@karlschriek I have plans to add an AWS (and other vendors) version(s) of my Argo CD distribution that add the integrations to the manifests. If this is also the route you or other cloud vendors want to take please message me on Slack so we can try and coordinate this effort.

PatrickXYS:
> @kubeflow/aws I have a few questions about this release. When we deploy to AWS we offload as much to managed AWS services as we possibly can, meaning in particular:
> 
> 1. Using Cognito for user pool management
> 2. Using RDS for all metadata storage (pipelines, metadb, cachedb, katibdb) instead of an on-cluster MySQL db
> 3. Using S3 for all pipeline artifact storage instead of using Minio for on-cluster storage
> 4. Using Secret Manager in combination with https://github.com/external-secrets/kubernetes-external-secrets in order to manage secrets (for example RDS database credentials)
> 5. Using IRSA as far as possible
> 
> From what I can tell, this distribution aims to roll out on EKS but does not attempt to integrate with any further AWS services. Is that the intention?
> 
> In the past (in particular with respect to Pipelines and KFServing) we've also had to do quite a bit of hacking to get the authentication via Cognito to work for users.

Hi @karlschriek can you create an issue in the https://github.com/aws/containers-roadmap/issues, and it will attract PM and Managers, and we could also look into how many customers are interested in this.

Thanks!

karlschriek:
Hi @PatrickXYS, sure I can do that. With respect to the current release you are working on, can you confirm if you plan to still and a kfdef with Cognito as was the case in previous releases?

We are also already looking at it and it is not yet evident how to make Cognito work with Istio 1.9.

PatrickXYS:
@karlschriek Our intial plan is to adopt community's manifest at this stage.

After 1.3 formal release, we'll explore the functionality of authenticating Cognito with istio 1.9

DavidSpek:
@PatrickXYS I’m working together with @karlschriek and @matty1979 to try and get Cognito integrated with the Argo CD deployment (which is also just using the upstream manifests). Would you or somebody else from AWS be interested in working together on this?

PatrickXYS:
@DavidSpek This is not related to current PR right? Can we create another PR to keep track?

yanniszark:
/lgtm
/hold
@PatrickXYS please remove the hold when convenient for you!

soleares:
Hi @PatrickXYS it looks like the following from `kfctl_aws.v1.2.0.yaml` are not included in this manifest

- aws-alb-ingress-controller
- nvidia-device-plugin
- istio-ingress
- aws-istio-authz-adaptor

Do you to know if they're compatible with the 1.3 release? I'd like to spin up a 1.3 cluster on EKS but it's probably not worth it yet if I can't use the ingress controller and GPU nodes.

DavidSpek:
@soleares The alb ingress controller is replaced by the [loadbalancer controller](https://github.com/kubernetes-sigs/aws-load-balancer-controller).
As an alternative to the Nvidia device plug-in I can recommend the [NVIDIA GPU operator](https://github.com/NVIDIA/gpu-operator) as this also allows for Prometheus monitoring of GPUs. 
The AWS Istio authz adaptor is not compatible with newer versions of Istio so that is why it isn’t included. 

If you are wanting a deployment that is integrated with AWS services such as RDS and S3 and also contains the gpu operator and other features necessary for a production deployment we almost have this finalized with [ArgoFlow](https://github.com/argoflow/argoflow-aws). This uses Argo CD for all the deployments making it easier to maintain you Kubeflow deployment over time and update it’s components. You can ping me on Slack if you need any help setting things up.

PR #1714, opened 2021-01-13
Update oidc-authservice

yhwang:
Use the newer image of oidc-authservice, add mandatary
parameter(AUTHSERVICE_URL_PREFIX) and change to
`istio/base_v3`.

Signed-off-by: Yihong Wang <yh.wang@ibm.com>

**Which issue is resolved by this Pull Request:**
Resolves #

**Description of your changes:**

**Checklist:**
- [x] Unit tests have been rebuilt: 
    1. `cd manifests/tests`
    2. `make generate-changed-only`
    3. `make test`

PR #1839, opened 2021-04-13
Update oidc-authservice image tag

DavidSpek:
While going through the image tags for the Kubeflow deployment I noticed that the image tag for the OIDC-authservice is from December 2019. Looking at the [registry](https://console.cloud.google.com/gcr/images/arrikto/GLOBAL/kubeflow/oidc-authservice?gcrImageListsize=30) there are some images that were created 12 days ago with a `-dirty` suffix, the image tag in this PR is the newest with this suffix.

@yanniszark I'm not sure what has changed with the OIDC-authservice, since this is an Arrikto project. Should this be updated? Do you know the difference between the regular and the `-dirty` image tags for the OIDC-authservice? I assume the newer versions contain bug/security fixes that Kubeflow distributions would then want to use.

### DIFF
--- a/common/oidc-authservice/base/kustomization.yaml
+++ b/common/oidc-authservice/base/kustomization.yaml
@@ -43,4 +43,4 @@ configurations:
 images:
 - name: gcr.io/arrikto/kubeflow/oidc-authservice
   newName: gcr.io/arrikto/kubeflow/oidc-authservice
-  newTag: 28c59ef
+  newTag: fef11c3


yanniszark:
@DavidSpek unfortunately it is not possible to do this yet. Please see:
https://github.com/kubeflow/manifests/pull/1714#issuecomment-803106340

DavidSpek:
@yanniszark Thanks for informing me. I figured there was likely a reason this was not done yet. Also, sorry for creating the duplicate PR. I doubt I will have the time for updating the Central Dashboard to be able to use newer images of the OIDC authservice, but I'll keep it in mind if I don't feel like working on other stuff one day.

Issue #1797, opened 2021-04-05
Upgrade to use Kustomize v4+


zijianjoy:
From the README of kubeflow/manifests: https://github.com/kubeflow/manifests#prerequisites, it is a prerequisite to use Kustomize `v3.2.0` for building resources. However, Kustomize v4+ is available. Can we upgrade prerequisite to use new version of Kustomize?

One thing that needs to be changed is that: flag name `load_restrictor` is changed in Kustomize v4+, we need to make the following updates in docs and build commands:

Change from `--load_restrictor=none` to `--load-restrictor LoadRestrictionsNone`. Note the difference between `_` and `-`.

yanniszark:
@zijianjoy thanks a lot for opening this issue to switch our target version to latest Kustomize v4.
This was not in our plan for 1.3.0, the reason being that we didn't want to change too many things together.
But it can very well be in our plans for the point release or 1.4. Let's use this issue to track the effort.

DavidSpek:
I am using Kustomize v4.0.5 for https://github.com/kubeflow-onprem/ArgoFlow and everything works as expected. The `--load_restrictor` flag is also not necessary except for the Katib manifests, which should be fixed and I have a PR open for this already. For the Argo installation I am not using `load_restrictor` (for Katib I am using my fixed version of the manifests) so this can be seen as proof that this flag is not necessary for deploying Kubeflow.

berndverst:
This issue burns me over and over. I am used to installing the latest version of tools and I ran into this again trying to work out the latest

As far as I can tell 3.9.2+ are broken with Knative also:
`kustomize build --load_restrictor=none common/knative/knative-eventing-install/base`

errors with

```
... at path 'spec/template/metadata/labels': wrong Node Kind for spec.template.metadata.labels expected: MappingNode was AliasNode: value: {*ref_0}
```

The latest version that works for me is 3.9.1 or 3.8.10. I installed and tested about 20 different versions.

berndverst:
FYI @DavidSpek with 4.0.5 Knative is also broken as is:

```
 ~/src/kubeflow/manifests-1.3-branch:  kustomize build --load-restrictor LoadRestrictionsNone common/knative/knative-eventing-install/base
```

```
wrong Node Kind for spec.template.metadata.labels expected: MappingNode was AliasNode: value: {*ref_0}
```

@yanniszark 
[...]

PR #1840, opened 2021-04-13
Sync Katib manifests with Kustomize v4 support

andreyvelich:
**Which issue is resolved by this Pull Request:**
Related PR: https://github.com/kubeflow/katib/pull/1498.

**Description of your changes:**
After these changes user can deploy Katib without `load_restrictor` flag.
I verified `katib-with-kubeflow-cert-manager` and the new `katib-with-kubeflow` install.
The output manifests should be equal.

/assign @DavidSpek @yanniszark @gaocegege @johnugeorge 

**Checklist:**
- [ ] Unit tests pass:
  **Make sure you have installed kustomize == 3.2.1**
    1. `make generate-changed-only`
    2. `make test`

PR #1844, opened 2021-04-14
Remove load_restrictor kustomize flag and lint README markdown

DavidSpek:
Related to: https://github.com/kubeflow/manifests/issues/1797

This PR removes the `load_restrictor` flag from the README and the testing scripts as this flag should not be needed since https://github.com/kubeflow/manifests/pull/1840 has been merged. This allows Kustomize v3 and v4 to be used by users. This PR also fixes a few style inconsistencies in the README.

/cc @yanniszark I'm unsure if the approach for the Python and Go is correct. Also, the `load_restrictor` flag is still used in 2 places in the pipelines manifests which should be fixed upstream.

**Checklist:**
- [ ] Unit tests pass:
    2. `make test`

### DIFF
--- a/docs/KustomizeBestPractices.md
+++ b/docs/KustomizeBestPractices.md
@@ -39,38 +39,6 @@ installs/
 Defining separate packages for each component makes it easier to use composition to define new configurations; e.g. using an external database as opposed
 to a database running in cluster.
 
-## Reuse patches
-
-We encourage reusing patches across kustomize packages when it makes sense. For example suppose we
-have an onprem and standalone version of our application but both of them want to reuse
-a common patch to use an external database. We could lay the packages out like so
-
-```
-components/
-          /patches/
-                  /deployment-external-db.yaml
-installs/
-        /app-standalone
-        /app-onprem
-```
-
-The kustomization files for app-standalone could then look like the following
-
-```
-apiVersion: kustomize.config.k8s.io/v1beta1
-...
-patchesStrategicMerge:
-- ../../components/patches/deployment-external-db.yaml
-```
-
-### Disable security check for file outside of directory root
-
-To support the above layout we need to disable [kustomizes' security check](https://github.com/kubernetes-sigs/kustomize/blob/master/docs/FAQ.md#security-file-foo-is-not-in-or-below-bar) by running with the `load_restrictor` flag: 
-
-```
-kustomize build --load_restrictor none $target
-```
-
 ## Command Line substitution
 
 To make it easy for users to override command line arguments use the following pattern.
--- a/docs/KustomizeBestPractices.md
+++ b/docs/KustomizeBestPractices.md
@@ -39,6 +39,42 @@ installs/
 Defining separate packages for each component makes it easier to use composition to define new configurations; e.g. using an external database as opposed
 to a database running in cluster.
 
+## Reuse patches
+
+**Note:** We are in the process of moving to Kustomize v4, see [this](https://github.com/kubeflow/manifests/issues/1797).
+This method of reusing patches is outdated and will likely be replaced by
+[kustomize components](https://github.com/kubernetes-sigs/kustomize/blob/master/examples/components.md).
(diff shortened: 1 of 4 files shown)


docs/KustomizeBestPractices.md
```diff
@@ -39,38 +39,6 @@ installs/
 Defining separate packages for each component makes it easier to use composition to define new configurations; e.g. using an external database as opposed
 to a database running in cluster.
 
-## Reuse patches
```
yanniszark, line 42: Can we keep this until we find a tested replacement?

hack/generate_tests.py
```diff
@@ -58,7 +58,7 @@ def run_kustomize_build(repo_root, package_dir):
   logging.info("Creating directory %s", output_dir)
   os.makedirs(output_dir)
 
-  subprocess.check_call([os.environ.get("KUSTOMIZE_BIN", "kustomize"), "build", "--load_restrictor", "none",
+  subprocess.check_call([os.environ.get("KUSTOMIZE_BIN", "kustomize"), "build",
```
yanniszark, line 61: Can we keep this until we have found a tested replacement to use instead of load_restrictor none?

tests/test_util.go
```diff
@@ -57,10 +58,7 @@ func RunTestCase(t *testing.T, testCase *KustomizeTestCase) {
 	}
 
 	fsys := fs.MakeRealFS()
-	// We don't want to enforce the security check.
-	// This is equivalent to running:
-	// kustomize build --load_restrictor none
-	lrc := loader.RestrictionNone
+	lrc := loader.RestrictionRootOnly
```
yanniszark, line 61: Same here, can we keep this until we have found a tested replacement to use instead of load_restrictor none?

hack/generate_tests.py
DavidSpek: Just pushed to commit to reverse this change

tests/test_util.go
DavidSpek: Just pushed to commit to reverse this change

docs/KustomizeBestPractices.md
DavidSpek: I thought it would be best not to encourage people to do exactly what we have now spent time on removing from the manifests. We wouldn't want people creating manifests that essentially cause a regression before people have had the time to play with and test [kustomize components]( https://github.com/kubernetes-sigs/kustomize/blob/master/examples/components.md).

yanniszark: I agree that we want to move to components. The problem we've had issues moving to kustomize v4: https://github.com/kubeflow/manifests/issues/1797
Does it make sense to leave a note here? That we want to move to `components` as an alternative to the `load_restrictor` flag, but can't do it yet because of https://github.com/kubeflow/manifests/issues/1797.

DavidSpek: I believe https://github.com/kubeflow/manifests/issues/1797 can be closed as the Katib manifests have been merged. But I just pushed a commit that adds this section again with a note.

yanniszark: @DavidSpek we can't close #1797 yet. Please see:
https://github.com/kubeflow/manifests/issues/1797#issuecomment-817017412

DavidSpek: Right, my mistake. I'll look into that a bit more as well. How does the note look?

yanniszark:
@DavidSpek thanks a lot!
/lgtm
/approve

I also have an off-topic question. Did you use a linter for the markdown newlines or just elbow grease?
If it was a linter, it would be great to add it to our CI.

DavidSpek:
@yanniszark For the Markdown linting I use the `markdownlint` VS-Code extension by David Anson. I went through it manually though to make sure it didn't make very large changes. We should see if this can be added to CI for this repo and others as I think that would give a much more consistent reading experience.

PR #1857, opened 2021-04-20
Update user references in README to use email

pvaneck:
Some updates to the README to account for the change of the default user login from `user` to `user@example.com`.

/cc @yanniszark


yanniszark:
Thanks @pvaneck!
/lgtm
/approve

PR #1902, opened 2021-06-11
Change user name to user@example.com in README

andreyvelich:
To continue this PR: https://github.com/kubeflow/manifests/pull/1857 we should also update few user names in the README.

/cc @DavidSpek @yanniszark @pvaneck 

**Checklist:**
- [ ] Unit tests pass:
    2. `make test`

### DIFF
--- a/README.md
+++ b/README.md
@@ -361,7 +361,7 @@ kustomize build apps/xgboost-job/upstream/overlays/kubeflow | kubectl apply -f -
 
 #### User Namespace
 
-Finally, create a new namespace for the the default user (named `user@example.com`).
+Finally, create a new namespace for the the default user (named `kubeflow-user-example-com`).
 
 ```sh
 kustomize build common/user-namespace/base | kubectl apply -f -
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ Option 2 targets customization and ability to pick and choose individual compone
 
 The `example` directory contains an example kustomization for the single command to be able to run.
 
-:warning: In both options, we use a default username (`user`) and password (`12341234`). For any production Kubeflow deployment, you should change the default password by following [the relevant section](#change-default-user-password).
+:warning: In both options, we use a default username (`user@example.com`) and password (`12341234`). For any production Kubeflow deployment, you should change the default password by following [the relevant section](#change-default-user-password).
 
 ### Prerequisites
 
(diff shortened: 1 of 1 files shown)


DavidSpek:
@andreyvelich I don't think this PR is correct. There is some (understandable) confusion surrounding when the `username` and when the `email` is used.

andreyvelich:
> @andreyvelich I don't think this PR is correct. There is some (understandable) confusion surrounding when the `username` and when the `email` is used.

Why ?
Aren't we confusing user if in the installation we are using [`user`](https://github.com/kubeflow/manifests/blob/master/README.md#installation) username and `12341234` password, but in the [port-forward](https://github.com/kubeflow/manifests/blob/master/README.md#port-forward) section we are using `user@example.com` name and `12341234` password ?

DavidSpek:
The confusing part is that Dex login requires the use of the email address, and not the user name. But in this PR you are stating:

```
we use a default username (`user@example.com`)
```
and
```
it includes a static user named `user@example.com`
```
Which is incorrect, as that isn't the username, but rather the user email address.

andreyvelich:
> The confusing part is that Dex login requires the use of the email address, and not the user name. But in this PR you are stating:
> 
> ```
> we use a default username (`user@example.com`)
> ```
> 
> and
> 
> ```
> it includes a static user named `user@example.com`
> ```
> 
> Which is incorrect, as that isn't the username, but rather the user email address.

Got it. What about namespace name. Should I leave the change with the correct one ?
```
Finally, create a new namespace for the the default user (named `kubeflow-user-example-com`).
```

yanniszark:
I agree with the spirit of this PR. Yes, we should mention the same thing everywhere, to not confuse the user.
Moreover, we can also make the wording more accurate, as @DavidSpek suggests.
How about:

> we use a default username (`user@example.com`)

we use a default email (`user@example.com`)

> a static user named `user@example.com`

a static user with email `user@example.com`

DavidSpek:
@yanniszark That looks good to me.
@andreyvelich I think the namespace is still correct yes.

andreyvelich:
Thanks @DavidSpek and @yanniszark. I will make this change.

andreyvelich:
@DavidSpek @yanniszark Are we ready to merge this PR?

DavidSpek:
Looks good to me!

/lgtm

yanniszark:
Thanks @DavidSpek @andreyvelich 
/lgtm
/approve

PR #1905, opened 2021-06-15
Add linguist configuration

davidspek:
By default, GitHub does not detect YAML files in a repository. As a result, this repository shows as being mostly Python, Go and Shell scripts. However, this is not an accurate representation of the repository for users (see below).

image

This PR adds a `.gitattributes` file that enables the detection of YAML and JSON by GitHub and takes these into account in the Languages section. Below is an example of what this would look like (from a different repository).

image

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+*.yaml linguist-detectable=true
+*.json linguist-detectable=true


yanniszark:
Thanks @DavidSpek, this lgtm.
/lgtm
/approve

PR #1917, opened 2021-06-24
Knative 0.17

yanniszark:
Refs https://github.com/kubeflow/manifests/issues/1754

Follow-up from https://github.com/kubeflow/manifests/pull/1768:
- Document commands for all changes, with `yq`, so we can deterministically and programmatically arrive at Knative manifests.
- Stop using 2 kustomizations per serving/eventing. Keep 1 kustomization for installing knative-serving and 1 for knative-eventing.
- Include the post-install jobs for people upgrading.
- Add a copyright notice since we remove comments, because `yq` does not handle them correctly.

cc @kubeflow/wg-manifests-leads @pvaneck


pvaneck (commented):
Thanks for the refactoring! I tested this out locally, testing out some inferenceservices, and everything worked as expected.

/lgtm

Issue #1933, opened 2021-07-06
KF 1.3.1 changes on v1.3-branch broke Distribution Openshift 1.3.0 release


nakfour:
Our example [kfdef](https://github.com/kubeflow/manifests/blob/master/distributions/kfdef/kfctl_openshift_v1.3.0.yaml) for KF 1.3.0 release pointed to v1.3-branch. When new changes were added to v1.3-branch for the KF 1.3.1 release this broke our distribution installation. We did open a PR to switch the uri pointing to the tag v1.3.0 instead of a branch. However, I would like to discuss if this is best practice to change a stable branch, example v1.3-branch after a release. I think it will be good to describe best practices for a Kubeflow release.
@kimwnasptd @yanniszark

PR #1934, opened 2021-07-06
Openshift distribution: Fixing the uri to point to a tag 1.3.0 and not a branch 1.3.0

nakfour:
#1933
Changing uri to point to a tag and not a branch. Also fixing the istio folder to the right name in the 1.3.0 tag

**Checklist:**
- [ ] Unit tests pass:
    2. `make test`

### DIFF
--- a/distributions/kfdef/kfctl_openshift_v1.3.0.yaml
+++ b/distributions/kfdef/kfctl_openshift_v1.3.0.yaml
@@ -13,7 +13,7 @@ spec:
   - kustomizeConfig:
       repoRef:
         name: manifests
-        path: distributions/stacks/openshift/application/istio-1-9-Openshift
+        path: distributions/stacks/openshift/application/istio-1-9-0-Openshift
     name: istio-stack
   - kustomizeConfig:
       repoRef:
@@ -80,5 +80,5 @@ spec:
     name: kubeflow-apps
   repos:
   - name: manifests
-    uri: https://github.com/kubeflow/manifests/archive/v1.3-branch.tar.gz
-  version: v1.3-branch
+    uri: https://github.com/kubeflow/manifests/archive/v1.3.0.tar.gz
+  version: v1.3-tag
--- a/distributions/kfdef/kfctl_openshift_v1.3.0.yaml
+++ b/distributions/kfdef/kfctl_openshift_v1.3.0.yaml
@@ -81,4 +81,4 @@ spec:
   repos:
   - name: manifests
     uri: https://github.com/kubeflow/manifests/archive/v1.3.0.tar.gz
-  version: v1.3-tag
+  version: v1.3.0-tag


distributions/kfdef/kfctl_openshift_v1.3.0.yaml
```diff
@@ -80,5 +80,5 @@ spec:
     name: kubeflow-apps
   repos:
   - name: manifests
-    uri: https://github.com/kubeflow/manifests/archive/v1.3-branch.tar.gz
-  version: v1.3-branch
+    uri: https://github.com/kubeflow/manifests/archive/v1.3.0.tar.gz
+  version: v1.3-tag
```
vpavlin, line 84: Should this be `v1.3.0`?

distributions/kfdef/kfctl_openshift_v1.3.0.yaml
```diff
@@ -13,7 +13,7 @@ spec:
   - kustomizeConfig:
       repoRef:
         name: manifests
-        path: distributions/stacks/openshift/application/istio-1-9-Openshift
+        path: distributions/stacks/openshift/application/istio-1-9-0-Openshift
```
vpavlin, line 16: Should we rename the dir referenced here from `istio-1-9-0-Openshift` to `istio-1-9-5-Openshift` since that is what is actually being deployed?

nakfour: @vpavlin thats for KF 1.3.1, this is specific for release 1.3.0 which is 1-9-Openshift

distributions/kfdef/kfctl_openshift_v1.3.0.yaml
nakfour: This is not really used anywhere, but fixed it

distributions/kfdef/kfctl_openshift_v1.3.0.yaml
```diff
@@ -80,5 +80,5 @@ spec:
     name: kubeflow-apps
   repos:
   - name: manifests
-    uri: https://github.com/kubeflow/manifests/archive/v1.3-branch.tar.gz
-  version: v1.3-branch
+    uri: https://github.com/kubeflow/manifests/archive/v1.3.0.tar.gz
+  version: v1.3.0-tag
```
crobby, line 84: Is "-tag" (or -branch as it was before) the common way to reference things?  Not terribly important, I guess, but it's probably good to go along with whatever the standard is across the project.

crobby (approved):

distributions/kfdef/kfctl_openshift_v1.3.0.yaml
nakfour: @crobby  so it was branch before because we were pulling from a branch and this broke our distribution because branch is not stable. We moved it to tag, and I placed tag in the version to indicate that. This```version``` is not used anywhere and I don't think we have standard for the format.

PR #1965, opened 2021-08-13
Update Knative to 0.22.1

### LINKED ISSUE #1957: Update Knative for 1.4
Right now the manifests are providing the `0.17` version of Knative [[source]](https://github.com/kubeflow/manifests/tree/master/common/knative).

@kubeflow/wg-serving-leads @pvaneck which version of Knative would you like to have for the 1.4 Kubeflow release? IIRC we are planning to include the [`0.6`](https://github.com/kubeflow/kfserving/tree/v0.6.0) version of KFServing, which has as a minimum requirement `0.17`. Should we stick with this version for 1.4?

kimwnasptd:
Resolves #1957 

Update the Knative to 0.22.1

I also made some changes to the `common/knative/knative-serving/kustomization.yaml` to:
1. Don't apply a namespace to all resources, since there are objects that should be applied in other namespaces like the `knative-local-gateway` service
2. Don't apply extra commonLabels to all the components

I've tried to create an InferenceService and it became Ready

/cc @pvaneck 
/cc @yuzisun


pvaneck (commented):
Tested this out and was able to deploy an inferenceservice and perform a curl request with the `X-Auth-Token` header (because of dex). :)

Initially, I was running with Kustomize v4. However, this fails with kustomize complaining about `Error: json: unsupported type: map[interface {}]interface {}` which is caused by the integer keys of the `PeerAuthentication` resources in `net-istio.yaml` (See: https://github.com/kubernetes-sigs/kustomize/issues/3446) . The workaround apparently is to put the integer key in quotes (`8443` -> `"8442"`).

Can this workaround also be applied and documented here so that those using later versions of kustomize won't hit issues when applying these manifests?

kimwnasptd:
@pvaneck sure! I changed the keys from `8443` to `8442` and the manifests can be generated with kustomize v4.2. I also updated the README.

PTAL

pvaneck (commented):
/lgtm

PR #1988, opened 2021-08-23
Update manifests for AutoML WG 0.12-rc.0

kimwnasptd:
closes #1970 

I copied the manifests from https://github.com/kubeflow/katib/tree/v0.12.0-rc.0/manifests/v1beta1

/cc @andreyvelich @johnugeorge @gaocegege

Issue #1961, opened 2021-08-10
Notebooks WG and Kubeflow 1.4 release


kimwnasptd:
@kubeflow/wg-notebooks-leads let's use this tracking issue to coordinate the integration of Notebooks with the Kubeflow 1.4 release.

First off a heads up that the feature freeze phase will start this Monday (16th August). Before then I'd like to have updated this repo with the manifests of the kubeflow/kubeflow repo, in order to be able to cut the first RC tag in this repo.

Since the Notebooks WG is following the same versioning with the Kubeflow release I expect that there will also be a `v1.4-branch` branch and corresponding RC tags.

So what I'd like to ask as a first step before the feature freeze is:
1. Are there any open issues/work in progress that you will be working on for your version as the KF release process will be progressing?
2. What will the K8s supported versions be for kubeflow/kfserving?

kimwnasptd:
Small heads, we are now in the Feature Freeze phase and updating the docs in kubeflow.org
https://github.com/kubeflow/manifests/tree/master/docs/releases/release-1.4#timeline

If you have any issues you'd like to work on for updating the docs, for the KF 1.4 release, please add a comment in kubeflow/website#2879 so we can track them.

kimwnasptd:
closing this, since KF 1.4 has been released

/close

Issue #1960, opened 2021-08-10
Serving WG and Kubeflow 1.4 release


kimwnasptd:
@kubeflow/wg-serving-leads @pvaneck  let's use this tracking issue to coordinate the integration of KFServing with the Kubeflow 1.4 release.

First off a heads up that the feature freeze phase will start this Monday (16th August). Before then I'd like to have updated this repo with the manifests of the kubeflow/kfserving repo, in order to be able to cut the first RC tag in this repo.

So what I'd like to ask as a first step before the feature freeze is:
1. What version of KFServing would you like to include for the 1.4 release?
2. Could you provide me with a branch/tag for this version? It doesn't have to be final. The branch/tag provided can keep on getting fixes through out the release process, but not new features
3. Are there any open issues/work in progress that you will be working on for your version as the KF release process will be progressing?
3. What will the K8s supported versions be for kubeflow/kfserving?

Also note, that we are already tracking the update of Knative to v0.22 in https://github.com/kubeflow/manifests/issues/1957

kimwnasptd:
@yuzisun @pvaneck I'll go on and use the `0.6` version from the https://github.com/kubeflow/kfserving/tree/v0.6.0 tag.

pvaneck:
Hey @kimwnasptd , yea let's go with the 0.6 release.  I just created a PR to pin the image tags in the release-0.5 branch: https://github.com/kubeflow/kfserving/pull/1763 . So once that merges, the PR of that commit can be used to sync with manifests.

kimwnasptd:
Thank you very much @pvaneck! That should be more than enough for now to sync manifests and create the first RC here.

kimwnasptd:
Small heads, we are now in the Feature Freeze phase and updating the docs in kubeflow.org
https://github.com/kubeflow/manifests/tree/master/docs/releases/release-1.4#timeline

If you have any issues you'd like to work on for updating the docs, for the KF 1.4 release, please add a comment in kubeflow/website#2879 so we can track them.

kimwnasptd:
closing this, since KF 1.4 has been released

/close

Issue #1959, opened 2021-08-10
Pipelines WG and Kubeflow 1.4 release


kimwnasptd:
@kubeflow/wg-pipeline-leads let's use this tracking issue to coordinate the integration of Pipelines with the Kubeflow 1.4 release.

First off a heads up that the feature freeze phase will start this Monday (16th August). Before then I'd like to have updated this repo with the manifests of the kubeflow/pipelines repo, in order to be able to cut the first RC tag in this repo.

So what I'd like to ask as a first step before the feature freeze is:
1. What version of Pipelines would you like to include for the 1.4 release?
2. Could you provide me with a branch/tag for this version? It doesn't have to be final. The branch/tag provided can keep on getting fixes through out the release process, but not new features
3. Are there any open issues/work in progress that you will be working on for your version as the KF release process will be progressing?
3. What will the K8s supported versions be for kubeflow/pipelines?

kimwnasptd:
@Bobgy @zijianjoy should I use the https://github.com/kubeflow/pipelines/tree/v1.7.0-alpha.3 tag as a starting point for the RC of the manifests?

kimwnasptd:
@Bobgy @zijianjoy kind ping for this

Bobgy:
1. We want to use 1.7.0
2. We can start with 1.7.0-rc.3, final release ETA next week. We will cut a release branch if patch versions are required.
3. We will finish the release.
4. I am interested in how many distributions can run docker container runtime. For the other distributions, we would recommend emissary executor now. https://www.kubeflow.org/docs/components/pipelines/installation/choose-executor/. I think both will support 1.18~ 1.21. 1.22 support is pending https://github.com/kubeflow/pipelines/issues/6014.

kimwnasptd:
Small heads, we are now in the Feature Freeze phase and updating the docs in kubeflow.org
https://github.com/kubeflow/manifests/tree/master/docs/releases/release-1.4#timeline

If you have any issues you'd like to work on for updating the docs, for the KF 1.4 release, please add a comment in kubeflow/website#2879 so we can track them.

Bobgy:
We just released 1.7.0.
It fixes a blocker issue: https://github.com/kubeflow/manifests/issues/1985.

kimwnasptd:
closing this, since KF 1.4 has been released

/close

PR #1989, opened 2021-08-23
Update the versions of components in README

kimwnasptd:
Makes sure the versions in the README are up to date. In the future we should enforce that whenever a PR updates the version of a component it will also need to update the README. Thanks to @andreyvelich for also bringing it up https://github.com/kubeflow/manifests/pull/1988#pullrequestreview-736331707.

cc @pvaneck @Bobgy @kubeflow/wg-notebooks-leads [since this PR touches these components]
/cc @elikatsis


annajung:
Versions match what was provided in 1.4 release tracking for [kfserving](https://github.com/kubeflow/manifests/issues/1960), [kfp](https://github.com/kubeflow/manifests/issues/1959), and [notebook](https://github.com/kubeflow/manifests/issues/1961)

/lgtm

PR #1974, opened 2021-08-16
Add 1.4 release information

annajung:
Signed-off-by: Anna Jung (VMware) <antheaj@vmware.com>

**Description of your changes:**
Based on [discussion of improving file structure for the release](https://kubeflow.slack.com/archives/C9V2WT2KV/p1628727885011700),
- rename directory `release` to `releases`
- add `release-1.4` directory
- add `release-1.4/README.md` with release timeline and relevant links
- add `release-1.4/release-team.md` with name of all release members

**Checklist:**
- [x] Unit tests pass:
  **Make sure you have installed kustomize == 3.2.1**
    1. `make generate-changed-only`
    2. `make test`

cc @RFMVasconcelos @DavidSpek @mkbhanda @kimwnasptd

PR #1996, opened 2021-08-30
Allocate 3 weeks for distribution testing

kimwnasptd:
Right now we only give two weeks for distribution testing, but we should include 3 https://github.com/kubeflow/manifests/blob/master/docs/releases/handbook.md#timeline.

The manifests testing should happen 2 weeks after the feature freeze phase has started and in parallel with the docs effort. 

I missed this in the review of the initial PR https://github.com/kubeflow/manifests/pull/1974

/cc @annajung @kubeflow/release-team


annajung:
Thanks for correcting this!
/lgtm

PR #2044, opened 2021-10-05
Fix knative image tags and runasnonroot

juliusvonkohout:

* Fixes Bug: kubeflow#2039
* Fix nonroot user to 65532 to properly support runasnonroot in distroless containers according to upstream https://github.com/GoogleContainerTools/distroless/blob/9287db1ed403f8e0494503759d5656bdfefec374/base/base.bzl#L9

I replaced the SHA256 digests with proper tags from the repository and added the right runasnonroot user

Sadly there are not yet upstream tags for all images:
```
For the following 4 packages used in the manifests (8 character digest shown), the digests do not match any existing tags.

gcr.io/knative-releases/knative.dev/eventing/cmd/apiserver_receive_adapter: a05b1db9
gcr.io/knative-releases/knative.dev/eventing/cmd/in_memory/channel_controller: a6983f71
gcr.io/knative-releases/knative.dev/eventing/cmd/in_memory/channel_dispatcher: 904f42a7
gcr.io/knative-releases/knative.dev/eventing/cmd/mtchannel_broker: a2678934
```

@Bobgy

**Checklist:**
- [ ] Unit tests pass:
    2. `make test`

### DIFF
--- a/common/knative/knative-eventing/base/upstream/eventing-core.yaml
+++ b/common/knative/knative-eventing/base/upstream/eventing-core.yaml
@@ -472,7 +472,7 @@ spec:
       containers:
         - name: eventing-controller
           terminationMessagePolicy: FallbackToLogsOnError
-          image: gcr.io/knative-releases/knative.dev/eventing/cmd/controller@sha256:6ddffbc286a84048cfd090193d00b4ecda25a3a7bf2de1a8e873f8b3755cc913
+          image: gcr.io/knative-releases/knative.dev/eventing/cmd/controller:v0.22.1
           resources:
             requests:
               cpu: 100m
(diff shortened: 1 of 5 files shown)


juliusvonkohout:
@kimwnasptd
@StefanoFioravanzo
Is there something wrong with my pull request?

umka1332:
> Using sha digests is actually a best practice since they are immutable, so it would be better to not change those.
> 
> /ok-to-test

Yes and no - they are pretty good in terms of no one can re-upload a different image with the same sha digest, but it is pretty inconvenient, when you try to deploy KF in an air gap environment - because those sha digest also changes when you renaming images and you need to also change tags, not only docker registry.

PR #2052, opened 2021-10-07
Update the example installation

kimwnasptd:
Since the example installation is using Dex/AuthService it should also use the new overlay introduced from #2048

/cc @elikatsis


kimwnasptd:
Thanks for the ping @sebastien-prudhomme! I'll be doing a last round of PRs to ensure everything is in sync

PR #2073, opened 2021-11-01
Fix envoy-filter indent

### LINKED ISSUE #1947: Invalid EnvoyFilter in OIDC-AuthService
`EnvoyFilter` manifest included in [/common/oidc-authservice/base/envoy-filter.yaml](https://github.com/kubeflow/manifests/blob/master/common/oidc-authservice/base/envoy-filter.yaml) is not valid. 

`listener` section needs to be indented one level higher.
`subFilter.name` must have a valid name (and not be empty).

and-1:
Resolves #1947

Added correct indent and filled missing subFilter name

### DIFF
--- a/common/oidc-authservice/base/envoy-filter.yaml
+++ b/common/oidc-authservice/base/envoy-filter.yaml
@@ -10,12 +10,12 @@ spec:
     - applyTo: HTTP_FILTER
       match:
         context: GATEWAY
-      listener:
-        filterChain:
-          filter:
-            name: "envoy.http_connection_manager"
-            subFilter:
-              name: ""
+        listener:
+          filterChain:
+            filter:
+              name: "envoy.filters.network.http_connection_manager"
+              subFilter:
+                name: "envoy.filters.http.router"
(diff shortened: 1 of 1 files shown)


common/oidc-authservice/base/envoy-filter.yaml
```diff
@@ -10,12 +10,12 @@ spec:
...
-          filter:
-            name: "envoy.http_connection_manager"
-            subFilter:
-              name: ""
+        listener:
+          filterChain:
+            filter:
+              name: "envoy.filters.network.http_connection_manager"
+              subFilter:
+                name: "envoy.filters.http.router"
```
junaid18183, line 18: We can still add subFilter like 
```
              subFilter:
                name: "envoy.router"
```

and-1: Yes, It will be correct filter also, but in runtime config i see name of that filter is envoy.filters.http.router and maybe in feature istio's release envoy.router name will stop working

PR #1787, opened 2021-04-01
namespace-related kustomizations

yanniszark:
- Add a kustomization for the kubeflow namespace
- Add a kustomization for the default user Profile

/assign @elikatsis

PR #2079, opened 2021-11-22
Remove Katib label from Kubeflow namespace

andreyvelich:

I notice that we added Katib Metrics Collector injection label to `kubeflow` namespace in this PR: https://github.com/kubeflow/manifests/pull/1787.

We should not do that since users have to submit Katib Experiments only in Profile's namespaces.

/assign @gaocegege @johnugeorge @yanniszark @kimwnasptd 

**Checklist:**
- [ ] Unit tests pass:
    2. `make test`


gaocegege (commented):
/lgtm

PR #2092, opened 2021-12-21
knative-serving: Workaround Istio race wrt Gateway creation

### LINKED ISSUE #2082: Istio and Knative problematic Gateway ports mapping in KF 1.4
After a handful of installations we bumped into an upstream Istio issue, that is triggered by Knative https://github.com/knative/serving/issues/10160#issuecomment-788669314, regarding how the ports are opened in the Gateway Pod. Istio has created a hotfix https://github.com/istio/istio/pull/33021 which was back-ported all the way to Istio 1.9.6, that comes along KF 1.4 https://github.com/istio/istio/commits/1.9.6. 

The above issue, that gets triggered by Knative's Gateways and Services, can result in **404 from InfereceServices**. We've seen users report 404s from KFServing even in the first RC of 1.4, but hadn't gotten to the bottom of it at that time https://github.com/kubeflow/manifests/issues/2007.

I'll provide some more technical details in the comments below on how to diagnose the specific issue, the root cause as well as how to fix it, but the key points I want to raise are:
1. This is triggered by the ordering that Knative and Istio's resources are applied
3. This affects all installations that are using manifests from Knative we provide, and can result in unusable installations

Because of the above, I'd like us to actually consider having a KF 1.4.1 release that will include the fix for the above problem https://github.com/knative-sandbox/net-istio/pull/636 in our manifests. cc @kubeflow/wg-manifests-leads @kubeflow/release-team

kimwnasptd:
Resolves #2082

When deploying knative-serving there a race due to an Istio issue that ends up having cluster-local-gateway misconfigured because of conflicting ports. In a nutshell Istio creates listeners based on Gateway ports and the corresponding Services (if any) and in our case we have two Services (cluster-local-gateway, knative-local-gateway) with port 80 and different targetPort (8080, 8081). In normal operation the pod gets the following listeners:
```
ADDRESS PORT  MATCH DESTINATION
0.0.0.0 8080  ALL   Route: http.80
0.0.0.0 8081  ALL   Route: http.8081
```

We have seen on fresh deployment to end up with only 8081:8081 and istio complaining with:
```
Error adding/updating listener(s) 0.0.0.0_8081: duplicate listener 0.0.0.0_8081 found
```

To fix that add a label in knative-local-gateway service so that it gets ignored by istio when translating gateway ports.

/cc @elikatsis

### DIFF
--- a/common/knative/knative-serving/base/kustomization.yaml
+++ b/common/knative/knative-serving/base/kustomization.yaml
@@ -8,6 +8,7 @@ patchesStrategicMerge:
 - patches/configmap-patch.yaml
 - patches/namespace-injection.yaml
 - patches/clusterrole-patch.yaml
+- patches/service-labels.yaml
 patches:
 - path: patches/sidecar-injection.yaml
   target:
--- a/common/knative/knative-serving/base/patches/service-labels.yaml
+++ b/common/knative/knative-serving/base/patches/service-labels.yaml
@@ -1,0 +1,7 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: knative-local-gateway
+  namespace: istio-system
+  labels:
+    experimental.istio.io/disable-gateway-port-translation: "true"


elikatsis:
Thanks @kimwnasptd !

/lgtm

PR #2121, opened 2022-01-27
Add networkpolicies under /contrib/networkpolicies

juliusvonkohout:
Resolves https://github.com/kubeflow/kubeflow/issues/6228#issuecomment-1021230549
@kimwnasptd 

Add networkpolicies under /contrib/networkpolicies

**Checklist:**
- [ ] Unit tests pass:
    2. `make test`


kimwnasptd:
Thank you for the time and work @juliusvonkohout! I'll look into them today and include them either in RC0 or RC1.

As a first comment, from a quick glance, could you also add a small README file and an OWNERS file? For the README I can think of the following information to expose:
* Why would a user apply the extra policies 
* Effects they will have in the cluster [how does this affect communication and between which services?]
* A small note that ideally we should achieve the same with AuthorizationPolicies, if not then this is something we must work on. If there's extra functionality that can not be achieved with the AuthorizationPolices then lets discuss and expose this as well

Regarding the OWNERS file can you add yourself as an approver? I can also be a reviewer to help if you want.

juliusvonkohout:
@kimwnasptd is this sufficient?

juliusvonkohout:
Now it is way more secure. One could limit even further how the pods in the kubeflow namespace can communicate with each other, but this would be quite cumbersome and a task for the future

juliusvonkohout:
@kimwnasptd i could also add podsecuritypolicies here and add networkpolicies and podsecuritypolicies to https://github.com/kubeflow/manifests/blob/master/example/kustomization.yaml

kimwnasptd:
The changes look good! One last nit is to remove me from the `approvers`, I want to try and make the OWNERS files to be as accurate as possible. Once in the future I also help with some reviews in the networking policies then let's add me as an `approver`.

juliusvonkohout:
> The changes look good! One last nit is to remove me from the `approvers`, I want to try and make the OWNERS files to be as accurate as possible. Once in the future I also help with some reviews in the networking policies then let's add me as an `approver`.

done

kimwnasptd:
Thanks!

/lgtm
/approve

Issue #2099, opened 2022-01-04
Basic tests for manifests


kimwnasptd:
We've seen in the KF 1.4 release (Manifests Testing phase), as well in the 1.4.1 patch release https://github.com/kubeflow/manifests/issues/2084#issuecomment-1000539950 that testing the manifests currently takes some manual cycles.

It will help us to reduce time on testing, and thus releasing faster, by defying some basic unit and integration tests that run on each PR, to ensure that at least:
1. The manifests an be generated with `kustomize`
2. The underlying Pods can be created and become Ready
3. The components can work together alongside the common dependencies (Istio, Knative, K8s version)

The aim of these tests should not be to run exhaustive tests on specific components, since this falls to the WGs' territory and responsibilities. The tests should provide some very basic checks that each component is deploy-able and can work, alongside the other components.

I'd like to create a proposal for this effort since it touches a lot of parts and needs a clear definition of goals and non-goals. I think it will be a good time to also start having a slightly more formal process for more advanced features.

cc @kubeflow/wg-manifests-leads

kimwnasptd:
Before starting with the proposal I want to evaluate where these tests should run.

Traditionally we have been running tests on Prow, by utilizing infra provided from the [kubeflow/testing](https://github.com/kubeflow/testing) repo and the Optional-testing infra that is backed by AWS. So I can see the following approaches:
1. Stick with Prow, that spins up AWS clusters, and run tests using Argo (or Tekton)
2. Go with GitHub Actions to trigger the tests and use something like KinD to setup a K8s cluster

Both have pros and cons, like being closer to a production environment, ease of writing the tests, time to setup a K8s cluster etc. But before going down the comparison road I'd first want to ensure that both options are viable **in the longs run**. 

Specifically, I'd like to have some assurance on the first approach which will help in evaluating the best long term approach. @surajkota can you provide some more insight on whether the Optional-testing will be supported in the future? 

Lastly, since we are getting close to the feature freeze and manifests testing phase of the [release](https://github.com/kubeflow/community/tree/master/releases/release-1.5) I'd like to submit the proposal early next week, to get this effort into motion. @surajkota If we could have this information, regarding the Optional-testing infra, by end of the week it would really help us for sticking with the timeline.

jaypipes:
[...]

PR #2128, opened 2022-02-07
tests: Scripts for e2e tests

kimwnasptd:
This is a first step for #2099 

The PR reworks the entire `tests` repo to include a simple e2e test that:
1. Installs all the KF manifests
2. Runs the E2E Notebooks we have in https://github.com/kubeflow/pipelines/blob/master/samples/contrib/kubeflow-e2e-mnist/kubeflow-e2e-mnist.ipynb

While developing the tests locally I was using KinD to spin up clusters and test everything from scratch. With this mentality, and along the fact that we don't have many maintainers in the `kubeflow/testing` repo, I tried to use GH actions to run these E2E tests.

But the performance was abysmal. Even with optimizations, like loading etcd in memory https://github.com/kubernetes-sigs/kind/issues/845, I couldn't get the tests to succeed. It took more than 45 mins to apply all the manifests and then the test failed due to etcd timeouts.

I've included the GH workflow files, to give it one last try, but I don't have many expectations. The next step for this is to see how could we use the AWS infra/prow to spin up the clusters and try our luck with there. The current blocking PR for this is https://github.com/kubeflow/testing/pull/972.

cc @kubeflow/release-team
/cc @StefanoFioravanzo @elikatsis


kimwnasptd:
The PR also broke the unit tests. I'll send a fix for them. We should be able to run unit tests for GH actions though, since this is not resource intensive.

/assign @kimwnasptd

kimwnasptd:
After inspecting the VM sizes of the GH actions I realized that the machines were running out of memory when I was installing all the manifests. It's a [7Gb VM](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#cloud-hosts-for-github-hosted-runners).

I managed to created a "light" version of the manifests that only install the barebone components, to get the tests running. This time it took **6mins** to install, but in the end got timeouts during training. This is most probably due to CPU limitations.

@StefanoFioravanzo @elikatsis I'm going to remove the GH action and fix unit tests. Then as a next step I'll implement this by using Prow.

kimwnasptd:
@elikatsis @StefanoFioravanzo this should be ready for review. To summarize:

1. I've removed old the deprecated files in the `tests` dir
2. I've updated the unit tests to build the `example` kustomization, as a quick sanity check that most of the components's files can be generated
3. I've prepared some helpers in `hack` for spinning up a KinD cluster **locally**, so that someone can test the e2e scripts
4. I've created `tests/e2e` scripts, which we can later use with Prow

The next step will be to use Prow, with workflows similar to what we do for Notebooks, in order to run our e2e tests there. https://github.com/kubeflow/kubeflow/blob/master/prow_config.yaml

Issue #2110, opened 2022-01-24
KServe and Kubeflow 1.5 release


kimwnasptd:
@yuzisun @animeshsingh @pvaneck starting this tracking issue to coordinate the integration of KServe with the Kubeflow 1.5 release.

First off a heads up that the feature freeze phase will start Tuesday (25th January). Before then I'd like to have updated this repo with the manifests of the [kserve/kserve](https://github.com/kserve/kserve) repo, in order to be able to cut the first RC tag in this repo.

From Paul's Community presentation I understand we are targeting KServe 0.7 and the corresponding git tag. 

We will also target Knative 0.22 and K8s 1.20 for the KF 1.5 release. Do the above sound good to you as well? I'll also take care of including the models web app as well, now that it has its own repo

cc @kubeflow/release-team

pvaneck:
Thanks, @kimwnasptd. I think that sounds good, though I wonder if a newer Knative Version would be preferred (e.g. v0.26). @yuzisun Do you have thoughts on the Knative version, or is 0.22.1 (the current version being used) still fine?

Tomcli:
It looks like Knative v0.22.1 might have some mutation errors on Kubernetes v1.22. I was digging into the release comments and some of the Kubernetes 1.22 fixes are patched in Knative 0.22.3. It is possible to upgrade the knative version?
https://github.com/knative/serving/releases/tag/v0.22.3

kimwnasptd:
Thanks for the heads up @Tomcli! 

> It looks like Knative v0.22.1 might have some mutation errors on Kubernetes v1.22

Can you also point some of the opened issues, so that users can be aware of them as well?

> It is possible to upgrade the knative version?

Considering that we are targeting K8s 1.20 and 1.21 for the KF 1.5 release, I'll put a slightly less priority in this update. But, since it's a very small bump I'll get to it once I complete the basic tests in manifests which will also help evaluating quickly such small updates #2099

Tomcli:
Here's the issue on k8s 1.22+
https://github.com/knative/serving/issues/11448

The new minor release is to address the new APIs in k8s 1.22+

yhwang:
@kimwnasptd are you going to provide `kustomization.yaml` under `contrib/kserve` and also the corresponding `models web app` under that folder for kubeflow 1.5?

kimwnasptd:
[...]

PR #2135, opened 2022-02-14
Sync kserve/models-web-app manifests

kimwnasptd:
Refs #2110 

The PR:
1. Adds the `contrib/kserve/models-web-app` manifests, from https://github.com/kserve/models-web-app/tree/v0.7.0/config
2. Rename `contrib/kserve/upstream` to `contrib/kserve/kserve`, so that the directory structure mirrors KServe org's repos


kimwnasptd:
Had to also update the `example/kustomization.yaml` accordingly. This should be ready for review

PR #2162, opened 2022-03-08
docs: Update KServe/KFServing section in README

pvaneck:

Added some info to the individual component section about KServe/KFServing.


pvaneck:
FYI @kimwnasptd 

Also, I see that the example kustomization.yaml still installs both KServe and KFServing. Should we comment out KFServing to avoid users hitting issues with two crds with the same name but different API groups?  (e.g. a user doing `kubectl get isvc` might return kfserving inferenceservices when they intended to just get kserve inferenceservices, so users would have to use fully qualified name `kubectl get inferenceservices.serving.kserve.io`).

kimwnasptd:
Thanks for updating this part @pvaneck, I had missed this. The instructions are very clear, and LGTM.

> Also, I see that the example kustomization.yaml still installs both KServe and KFServing. Should we comment out KFServing to avoid users hitting issues with two crds with the same name but different API groups?

The kubectl argument is a valid one but we'd still like to include both of them, at least for this release. Of course manifests adopters can use any of the two or just one altogether.

/lgtm
/approve

PR #2164, opened 2022-03-08
Use KServe component in tests instead of KFServing

js-ts:
The upstream example https://github.com/kubeflow/pipelines/blob/master/samples/contrib/kubeflow-e2e-mnist/kubeflow-e2e-mnist.ipynb has been updated to use kserve component instead of the kfserving one but users can still
use the kfserving component if they want
https://github.com/kubeflow/manifests/compare/master/tests/e2e/utils/kserve.py#L6-L11


js-ts:
cc @kimwnasptd @pvaneck

js-ts:
cc @kimwnasptd

kimwnasptd:
Apologies for the huge delay here, thanks for the PR @js-ts!

/lgtm
/approve

PR #2170, opened 2022-03-14
(bug) Increase Resources to Tensboard Controller

### LINKED ISSUE #2101: Tensorboards-controller crash loop in OOM Killed
Hello there,

We have encountered a problem about the default resources of tensorboards controller.

We deployed kubeflow following [this section](https://github.com/kubeflow/manifests#install-individual-components) about 60 days ago, starting from v1.3 and upgraded to v1.4.1

We did not keep observing all the service status, but as far as I can remember, all pods were in Running state when we first deploy kubeflow v1.3

Currently we notice that the tensorboard-controller pod keeps crashing because of OOM Killed.

We expanded the memory requests to 40mi and limits to 60mi, and the error is still there.

But both to 100mi the error disappeared.

Is the default resource limit too small?

vinayan3:
Resolves #2101

Increase the amount of CPU and memory to the Tensorboard controller otherwise it will be in a crash backoff loop.

**Checklist:**
- [ ]  Unit tests pass: -  Could not find the make file to run this.
    2. `make test`

### DIFF
--- a/apps/tensorboard/tensorboard-controller/upstream/manager/manager.yaml
+++ b/apps/tensorboard/tensorboard-controller/upstream/manager/manager.yaml
@@ -31,9 +31,9 @@ spec:
         name: manager
         resources:
           limits:
-            cpu: 100m
-            memory: 50Mi
+            cpu: 300m
+            memory: 200Mi
           requests:
             cpu: 100m
-            memory: 30Mi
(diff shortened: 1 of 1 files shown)


vinayan3:
Closing PR because my email was kept private and needs to be public

PR #2173, opened 2022-03-15
(bug) Increase Resources to Tensboard Controller

### LINKED ISSUE #2101: Tensorboards-controller crash loop in OOM Killed

vinayan3:
Resolves #2101 

* Increase resources so that the controller stops crashing.

**Checklist:**
- [ ] Unit tests pass: - no make file.
    2. `make test`

### DIFF
--- a/apps/tensorboard/tensorboard-controller/upstream/manager/manager.yaml
+++ b/apps/tensorboard/tensorboard-controller/upstream/manager/manager.yaml
@@ -31,9 +31,9 @@ spec:
         name: manager
         resources:
           limits:
-            cpu: 100m
-            memory: 50Mi
+            cpu: 300m
+            memory: 200Mi
           requests:
             cpu: 100m
-            memory: 30Mi
+            memory: 100Mi
       terminationGracePeriodSeconds: 10


Arhell (commented):
Please sign CLA
https://github.com/kubeflow/manifests/pull/2173/checks?check_run_id=5561270783

vinayan3:
@Arhell Ii have to work out with the corporate signer what is going on with the CLA. I won't be able to commit until that is sorted out.

juangon:
I am reproducing this problem too

kimwnasptd:
Thanks for the PR @vinayan3! This is definitely something to include.

But, can you make the PR on the `kubeflow/kubeflow` repo where the source of the manifests is? We only copy manifests in this repo from the components' repos.

vinayan3:
> Thanks for the PR @vinayan3! This is definitely something to include.
> 
> But, can you make the PR on the `kubeflow/kubeflow` repo where the source of the manifests is? We only copy manifests in this repo from the components' repos.

Yes I'll get up a PR for this in the main repo. Thanks for letting me know how this works.

Issue #2165, opened 2022-03-09
cache-deployer Pod fails for kubeflow 1.5 


akartsky:
When installing kubeflow from [latest master](https://github.com/kubeflow/manifests/tree/039442a950916ad9c28a327bfa78a16f022d9939) and rc2 release `cache-deployer-deployment` fails
```
kubeflow       cache-deployer-deployment-6f4bcc969-jxq2f        1/2     CrashLoopBackOff   23         102m
kubeflow       cache-server-575d97c95-md8bg                     0/2     Init:0/1           0          102m
```
Error logs for `cache-deployer` : 
```
+ for x in $(seq 10)
++ kubectl get csr cache-server.kubeflow -o 'jsonpath={.status.certificate}'
+ serverCert=
+ [[ '' != '' ]]
+ sleep 1
+ [[ '' == '' ]]
+ echo 'ERROR: After approving csr cache-server.kubeflow, the signed certificate did not appear on the resource. Giving up after 10 attempts.'
ERROR: After approving csr cache-server.kubeflow, the signed certificate did not appear on the resource. Giving up after 10 attempts.
+ exit 1
```

`cache-server.kubeflow` looks like this :
```
$ kubectl get csr cache-server.kubeflow -o yaml

apiVersion: certificates.k8s.io/v1
kind: CertificateSigningRequest
metadata:
  creationTimestamp: "2022-03-09T03:05:10Z"
  name: cache-server.kubeflow
  resourceVersion: "306173"
  selfLink: /apis/certificates.k8s.io/v1/certificatesigningrequests/cache-server.kubeflow
  uid: f6179659-fa58-4c3b-a9af-d69543667415
spec:
  groups:
  - system:serviceaccounts
  - system:serviceaccounts:kubeflow
  - system:authenticated
[...]

PR #2186, opened 2022-04-08
fix(kfp_cache): Use cert manager for kfp cache

akartsky:
Resolves https://github.com/kubeflow/manifests/issues/2165
First merge https://github.com/kubeflow/pipelines/pull/7538 and use new cache server image

Use cert manager instead of cache deployer for KFP

### DIFF
--- a/example/kustomization.yaml
+++ b/example/kustomization.yaml
@@ -26,7 +26,7 @@ resources:
 
 
 # Kubeflow Pipelines
-- ../apps/pipeline/upstream/env/platform-agnostic-multi-user
+- ../apps/pipeline/upstream/env/aws/multi-user
 # KFServing
 - ../apps/kfserving/upstream/overlays/kubeflow
 # Katib
--- a/apps/pipeline/upstream/env/aws/multi-user/base/kustomization.yaml
+++ b/apps/pipeline/upstream/env/aws/multi-user/base/kustomization.yaml
@@ -1,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+bases:
+  - ../../../platform-agnostic-multi-user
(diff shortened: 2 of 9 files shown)


zijianjoy:
I think the whole folder of `apps/pipeline/upstream` are copied from https://github.com/kubeflow/pipelines/tree/master/manifests/kustomize. You will need to create an overlay, either in https://github.com/kubeflow/pipelines/tree/master/manifests/kustomize, or reorganize the current repo to be able to inject your overlay under  `apps/pipeline/upstream` (Example: https://github.com/kubeflow/manifests/tree/master/contrib/kserve/models-web-app)

zijianjoy:
cc @kimwnasptd

kimwnasptd:
>I think the whole folder of apps/pipeline/upstream are copied from https://github.com/kubeflow/pipelines/tree/master/manifests/kustomize. You will need to create an overlay, either in https://github.com/kubeflow/pipelines/tree/master/manifests/kustomize, or reorganize the current repo to be able to inject your overlay under apps/pipeline/upstream (Example: https://github.com/kubeflow/manifests/tree/master/contrib/kserve/models-web-app) 

Just wanted to comment the exact same thing, thanks @zijianjoy!

zijianjoy:
Note: `contrib/` is for Kubeflow component add-on like kserve. Since your overlay patch is for pipelines. You should probably still add it under `apps/pipeline/`, just not `apps/pipeline/upstream`. But I might be wrong because I am not managing this repo. @kimwnasptd please advise on the right location to put this overlay.

kimwnasptd:
Until now we expected all overlays to be part of the components' manifests. And this repo only had a copy of these manifests.

@zijianjoy I'd like to first evaluate if we can include this in the Pipelines manifests, and if not possible then add a commit here.

Since this is something that depends on the platform, would it make sense to include the cache-deployer patch inside https://github.com/kubeflow/pipelines/tree/master/manifests/kustomize/env/aws?

akartsky:
as discussed in the KF community meeting today 
I will be moving these manifests to KFP repo

example/kustomization.yaml
```diff
@@ -26,7 +26,7 @@ resources:
 
 
 # Kubeflow Pipelines
-- ../apps/pipeline/upstream/env/platform-agnostic-multi-user
```
zijianjoy, line 29: Can we keep this file unchanged?

PR #2204, opened 2022-05-02
Add patches to remove namespaces and edit kustomization

keithmattix:
 When applying kustomize generated kubeflow manifests to a k8s cluster without kubectl, the k8s apiserver will error on several of them due to certain cluster-scoped (i.e. no namespace) manifests containing namespace metadata. To fix this, I've created JSON patches for the resources that I encountered issues with.

**Checklist:**
- [ ] Unit tests pass:
    2. `make test`

### DIFF
--- a/apps/kfp-tekton/upstream/base/installs/multi-user/pipelines-profile-controller/patches/remove-namespace.yaml
+++ b/apps/kfp-tekton/upstream/base/installs/multi-user/pipelines-profile-controller/patches/remove-namespace.yaml
@@ -1,0 +1,2 @@
+- op: remove
+  path: /metadata/namespace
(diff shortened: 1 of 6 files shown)


juliusvonkohout:
This is the wrong repository. You need to fix it in the upstream repositories.

keithmattix:
Thanks for pointing that out @juliusvonkohout; I didn't realize that. Some follow-up questions:
1. The cert-manager kubeflow-issuer doesn't have an upstream repo, correct? Since it's a ClusterIssuer `kind`, the kustomization.yaml actually doesn't need a namespace declaration at all. 
2. The only way (that I'm aware of) to remove namepaces from certain cluster-scoped resources would be to remove[ top-level namespace declarations](https://github.com/kubeflow/pipelines/blob/master/manifests/kustomize/env/platform-agnostic-multi-user/kustomization.yaml#L23) from each kustomize.yaml and explicitly add the namespace to each manifest. Does that seem more desirable to you than patches?

Arhell (commented):
/ok-to-test

juliusvonkohout:
Please put this on the agenda of the next manifest wg meeting in two weeks and join us for a discussion

keithmattix:
I'm no longer working on this

Issue #2209, opened 2022-05-17
ml-pipeline-apiserver pods can fail to come up when there's a lengthy database migration


droctothorpe:
When we pivoted from an in-cluster mysql deployment to an externally hosted RDS instance for the metadata store, we ran into a scenario where the ml-pipeline-apiserver pod couldn't start because it didn't have a long enough grace period to complete its database migration.

We fixed this with a startup probe. I'll submit the associated PR momentarily. Filing this issue for proper book-keeping.

PR #2210, opened 2022-05-17
Add startup probe to pipeline API deployment

droctothorpe:
Co-authored-by: moorthy156 <narayanamoorthy.mari@gmail.com>

Resolves # https://github.com/kubeflow/manifests/issues/2209

When we pivoted from an in-cluster mysql deployment to an externally hosted RDS instance for the metadata store, we ran into a scenario where the ml-pipeline-apiserver pod entered a crash loop because it didn't have a long enough grace period to complete its database migration.

This pod is unique in that its startup time is variable depending on whether or not a migration needs to occur. A [startup probe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/) is perfect for this, since it offers a grace period in the event of a migration, but doesn't delay pod readiness when a migration does not need to happen.

**Checklist:**
- [x] Unit tests pass:
    2. `make test`

### DIFF
--- a/apps/pipeline/upstream/base/pipeline/ml-pipeline-apiserver-deployment.yaml
+++ b/apps/pipeline/upstream/base/pipeline/ml-pipeline-apiserver-deployment.yaml
@@ -105,6 +105,21 @@ spec:
           initialDelaySeconds: 3
           periodSeconds: 5
           timeoutSeconds: 2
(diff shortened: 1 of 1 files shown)


juliusvonkohout:
Please fix this in the kfp repository. The manifests here are just copied.

PR #2230, opened 2022-06-30
Test that each component can be applied in a K8s 1.22 cluster

NickLoukas:
Add GH actions for applying manifests at any pull request with the steps bellow:
1. Install KinD and configure a cluster with a specific K8s version
2. Install `kustomize`
3. Install `Istio`
4. Build and apply the manifests

### DIFF
--- a/.github/workflows/centraldb_kind_test.yaml
+++ b/.github/workflows/centraldb_kind_test.yaml
@@ -21,14 +21,11 @@ jobs:
       run: ./testing/gh-actions/install_kustomize.sh
 
     - name: Install Istio
-      run: |
-        cd common/istio-1-11
-        kustomize build istio-crds/base | kubectl apply -f -
-        kustomize build istio-namespace/base | kubectl apply -f -
-        kustomize build istio-install/base | kubectl apply -f -
+      run: ./testing/gh-actions/install_istio.sh
 
     - name: Build & Apply manifests
       run: |
         cd apps/centraldashboard/upstream
         kubectl create ns kubeflow
         kustomize build overlays/kserve | kubectl apply -f -
+        kubectl wait --for=condition=Ready pods --all --all-namespaces --timeout 180s
--- a/.github/workflows/kserve_kind_test.yaml
+++ b/.github/workflows/kserve_kind_test.yaml
@@ -21,20 +21,10 @@ jobs:
       run: ./testing/gh-actions/install_kustomize.sh
 
     - name: Install Istio
-      run: |
-        cd common/istio-1-11
-        kustomize build istio-crds/base | kubectl apply -f -
-        kustomize build istio-namespace/base | kubectl apply -f -
-        kustomize build istio-install/base | kubectl apply -f -
+      run: ./testing/gh-actions/install_istio.sh
 
     - name: Install cert-manager
-      run: |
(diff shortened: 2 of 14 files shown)


.github/workflows/kserve_kind_test.yaml
```diff
@@ -0,0 +1,44 @@
...
+    - name: Install Istio
+      run: |
+        cd common/istio-1-11
+        kustomize build istio-crds/base | kubectl apply -f -
+        kustomize build istio-namespace/base | kubectl apply -f -
+        kustomize build istio-install/base | kubectl apply -f -
+
+    - name: Install cert-manager
+      run: |
+        cd common/cert-manager
```
kimwnasptd, line 32: Since we use this in a couple of places let's also create script

.github/workflows/centraldb_kind_test.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+    - name: Create KinD Cluster
+      run: kind create cluster --config testing/gh-actions/kind-1-22.yaml
+
+    - name: Install kustomize
+      run: ./testing/gh-actions/install_kustomize.sh
+
+    - name: Install Istio
+      run: |
+        cd common/istio-1-11
```
kimwnasptd, line 25: let's have a common script for istio install as well. In the future we'll move to 1.14, so it'll be easier to just do one change in one file rather than changing all the actions

.github/workflows/centraldb_kind_test.yaml
```diff
@@ -0,0 +1,34 @@
...
+        kustomize build istio-crds/base | kubectl apply -f -
+        kustomize build istio-namespace/base | kubectl apply -f -
+        kustomize build istio-install/base | kubectl apply -f -
+
+    - name: Build & Apply manifests
+      run: |
+        cd apps/centraldashboard/upstream
+        kubectl create ns kubeflow
+        kustomize build overlays/kserve | kubectl apply -f -
```
kimwnasptd, line 34: Let's have an extra check here with `kubectl wait --for=condition=Ready pods --all --all-namespaces --timeout 180s` and ensure that all pods get created. We can set a timeout of 3 mins for now. This way we also ensure that the Pods can run as well

kimwnasptd (commented):
Let's use `kubectl wait` to also ensure the Pods of the components get created as well. We should do this for all tests we have right now

.github/workflows/kserve_kind_test.yaml
NickLoukas: I added the script.

.github/workflows/centraldb_kind_test.yaml
NickLoukas: I made a script for this too.

.github/workflows/centraldb_kind_test.yaml
NickLoukas: I added this in all the gh actions.

NickLoukas:
@kimwnasptd I addressed the review comments above. Take a look!

kimwnasptd:
looks good @NickLoukas! We will also battle test these once we'll sync the manifests of the apps 

/lgtm
/approve

PR #2232, opened 2022-06-30
Sync kubeflow training operator manifests v1.5.0 rc.0

kimwnasptd:
Used the sync script in `hack`.

I also updated the README to use the `training-operator` GH link, instead of the `tf-operator` one. Lastly I updated the sync script to automatically fix the README as well.

cc @johnugeorge


kimwnasptd:
All tests also pass, which means:
1. Training Operator manifests can be installed in a (KinD) K8s 1.22 cluster
2. All pods can become ready

which is great! cc @kubeflow/release-team @annajung

PR #2234, opened 2022-06-30
Update kubeflow/katib manifests from v0.14.0-rc.0

kimwnasptd:
Updates Katib for commit `v0.14.0-rc.0`. Just used the sync script for this, which was up to date.

cc @johnugeorge


kimwnasptd:
All tests also pass, which means:

1. Katib manifests can be installed in a (KinD) K8s 1.22 cluster
2. All pods can become ready

which is great! cc @annajung

PR #2237, opened 2022-06-30
Move all tests under tests dir

kimwnasptd:
Initially we introduced the GH Action helpers under `testing/gh-actions`. But we already have a `tests` folder, so I'm moving everything there.

To change the files I:
1. Found all references of `testing/gh-actions`
    ```bash
    cd .github
    grep "testing/gh-actions" -r *
    ```
2. Replaced all instances
    ```bash
    find . -type f -exec sed -i "s#testing/gh-actions#tests/gh-actions" {} +
    ```

cc @NickLoukas


kimwnasptd:
@NickLoukas after that, I noticed that we have configuration yamls for KinD in both `hack` and `test/gh-actions`. 

Let's:
1. Move the config files from hack to testing/gh-actions
2. Add a comment in each file that it needs to work with kind v0.14
3. Ensure we use the corresponding images they propose for each K8s version https://github.com/kubernetes-sigs/kind/releases/tag/v0.14.0

NickLoukas:
Everything looks good!

/lgtm

PR #2238, opened 2022-06-30
Change images of KinD configuration files to be compatible with v0.14.0

NickLoukas:
As requested here https://github.com/kubeflow/manifests/pull/2237#issuecomment-1171430252 we're moving all the KinD configuration files in one place and make sure that they all use the images from v0.14.0.

### DIFF
--- a/tests/gh-actions/kind-cluster-1-22.yaml
+++ b/tests/gh-actions/kind-cluster-1-22.yaml
@@ -1,3 +1,4 @@
+# This configuration file should work with KinD v0.14.0
 apiVersion: kind.x-k8s.io/v1alpha4
 kind: Cluster
 # Configure registry for KinD.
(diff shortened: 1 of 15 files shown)


tests/gh-actions/kind-cluster-1-22.yaml
```diff
@@ -1,3 +1,4 @@
+# This configuration file should work with KinD v0.14.0
```
kimwnasptd, line 1: Since you changed the name from `kind-1-22.yaml` to `kind-cluster-1-22.yaml` we'll need to update the actions that use this config.

NickLoukas: Fixed it.

NickLoukas:
@kimwnasptd I addressed the review comments above and rebased on top of the recent changes in master to resolve all the conflicts. Please take a look.

PR #2242, opened 2022-07-02
Update kubeflow/kubeflow manifests from v1.6.0-rc.0

kimwnasptd:
Updates the Notebooks WG manifests from `kubeflow/kubeflow` from the [`v1.6.0-rc.0`](https://github.com/kubeflow/kubeflow/releases/tag/v1.6.0-rc.0) tag.

Major changes:
1. We are using the [`docker.io/kubeflownotebookswg`](https://hub.docker.com/u/kubeflownotebookswg), we'll also provide instructions on how users will need to modify their overlays to target the newer images
2. Changes for K8s 1.22


kimwnasptd:
The tests are passing which means:
1. All manifests can be applied in a (KinD) K8s cluster
2. All pods can become ready

cc @annajung

Issue #2200, opened 2022-04-26
Manifests WG and Kubeflow 1.6 release


annajung:
Hello @kubeflow/wg-manifests-leads 

The [1.6 release team](https://github.com/kubeflow/community/blob/master/releases/release-1.6/release-team.md) will use this tracking issue to coordinate the integration of the Manifests with the Kubeflow 1.6 release.

Heads up that the **feature freeze** is scheduled for **Wednesday, Jun 1st, 2022**. For the full release schedule, refer to the [1.6 release timeline](https://github.com/kubeflow/community/blob/master/releases/release-1.6/README.md).

_(note: links are pending https://github.com/kubeflow/community/pull/558 and https://github.com/kubeflow/community/pull/560)_

To prepare for the 1.6 release, please provide answers to the following questions:
1. What version of Manifests would you like to include for the 1.6 release?
2. Could you provide a branch/tag for this version? _(note: It doesn't have to be final)_
3. Can you share a link to all open / work in progress issues that you're expecting to include in the 1.6?
4. If you're using the following dependencies, could you provide the versions you'll be supporting for each dependency?
  - Knative
  - Istio
  - Kubernetes
  - Cert-Manager
  - Kustomize

The release team WG liaison for Manifests WG will be @kimwnasptd @DomFleischmann @DnPlas nand you can keep track of high-level release updates in the [1.6 tracking](https://github.com/kubeflow/manifests/issues/2194) issue.

kimwnasptd:
Thanks for opening the issue @annajung! We are still evaluating these versions, but I'll be back very soon with a definitive answer.

What we are currently targeting is:
1. Knative 0.23.2
2. For Istio we are evaluating 1.13, since the upgrade path seems simple enough https://istio.io/latest/news/releases/1.13.x/announcing-1.13/upgrade-notes/ https://istio.io/latest/news/releases/1.12.x/announcing-1.12/upgrade-notes/, but we need to thoroughly test this
3. For K8s we want to start looking into 1.23, for the common components like Istio, Knative and Cert-Manager
4. For Kustomize we are still trying to get to 4+, with main blocker being https://github.com/kubeflow/manifests/issues/1797#issuecomment-861491762

juliusvonkohout:
I have two important things for 1.6

1. https://github.com/kubeflow/manifests/pull/2205 - updating the networkpolicies
2. https://github.com/kubeflow/manifests/pull/2150 - new authentication methods in oidc-authservice

[...]

PR #2243, opened 2022-07-02
Update Dex for K8s 1.22

kimwnasptd:
Refs #2200

Dex had some `CRD`, `ClusterRole` and `ClusterRoleBinding` resources that were using the deprecated `apiextensions.k8s.io/v1` API.

This PR:
1. Updates these APIs
2. Add a GH Action for testing the Dex manifests as well

### DIFF
--- a/.github/workflows/dex_kind_test.yaml
+++ b/.github/workflows/dex_kind_test.yaml
@@ -1,0 +1,30 @@
+name: Build & Apply Dex manifests in KinD
+on:
+  pull_request:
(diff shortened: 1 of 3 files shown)


DomFleischmann:
We have tried dex version v2.31.2 for our next release to ensure support for later  k8s versions in our distribution. Maybe we could consider using this version in upstream too? Additionally newer versions for dex aren't uploaded in quay anymore so we might want to update the image anyways.

Issue #2248, opened 2022-07-13
Simplistic integration testing for components


kimwnasptd:
We've done some first steps on having [tests](https://github.com/kubeflow/manifests/tree/master/.github/workflows) that:
1. Apply all components in a KinD K8s 1.22 cluster
2. Wait for all Pods to become ready

This is done with GH Actions, since these tests are lightweight and can complete relatively quickly.

As a next step we can extend these tests to also:
1. Apply a CR for each component
2. Wait for that object to become Ready/Succeed

This way we add yet another small layer of testing and ensure the components and their functionality is working as expected.

Once we fully move to AWS accounts we can deprecate these in favor of the [e2e test](https://github.com/kubeflow/manifests/tree/master/tests/e2e), which will be doing the same but by deploying a whole Kubeflow installation and ensuring that everything plays along together as expected.

cc @kubeflow/wg-manifests-leads

juliusvonkohout:
I think this is done now for a few components with machine to machine tokens.

Please reopen if necessary.

PR #2249, opened 2022-07-13
Extend Katib GH action to also create an experiment and assert it succeeds

NickLoukas:
Refs #2248


kimwnasptd:
cc @kubeflow/wg-automl-leads the goal is to just sanity check that Experiments can work as expected alongside the common dependencies of Kubeflow. 

In the future once we'll move to AWS we'll just run the E2E tests for these, but for now we'll be testing each component separately

tests/gh-actions/katib_test.yaml
```diff
@@ -0,0 +1,78 @@
+---
```
kimwnasptd, line 1: only nit I have would be to put the files/yamls we'll need for these tests into a distinct folder and not have everything flat under `tests/gh-actions`. Let's do `tests/gh-actions/kf-objects` WDYT?

NickLoukas: Yeap, I agree. I'll add a commit to arrange the folders like that.

NickLoukas:
@kimwnasptd I added an extra commit here for the review comment above.

kimwnasptd:
Thanks! 

/lgtm
/approve

PR #2250, opened 2022-07-13
Update kubeflow/pipelines manifests from 2.0.0-alpha.3

annajung:
Signed-off-by: Anna Jung (VMware) <antheaj@vmware.com>

File updated using the `./hack/sync-pipelines-manifests.sh`
- Updates KFP to 2.0.0-alpha.3

cc @kimwnasptd @surajkota @zijianjoy @chensun @kubeflow/release-team


.github/workflows/pipeline_kind_test.yaml
```diff
@@ -30,5 +30,7 @@ jobs:
       run: |
         cd apps/pipeline/upstream
         kubectl create ns kubeflow
+        kubectl apply -f third-party/metacontroller/base/crd.yaml
+        kubectl wait --for condition=established crd compositecontrollers.metacontroller.k8s.io
```
annajung, line 34: In addition to syncing pipelines manifest, I had to add additional commands in the github workflow for pipeline test to wait for the CRD creation before creating the `CompositeController`. Without it, it runs into the [no matches for kind "CompositeController" in version "metacontroller.k8s.io/v1alpha1" error](https://github.com/kubeflow/manifests/runs/7329985403?check_suite_focus=true)

.github/workflows/pipeline_kind_test.yaml
```diff
@@ -30,5 +30,7 @@ jobs:
       run: |
         cd apps/pipeline/upstream
         kubectl create ns kubeflow
+        kubectl apply -f third-party/metacontroller/base/crd.yaml
```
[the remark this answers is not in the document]
annajung: Sure, I can modify it to follow the same format as the Kserve installation

kimwnasptd:
Thanks for the PR @annajung and for creating the tag @chensun!

The changes in the workflows LGTM and the tests are passing so we are good to go. Small mention that we will soon add a small pipeline test as an extra step #2248 

/lgtm
/approve

.github/workflows/pipeline_kind_test.yaml
surajkota: Can you help point out how this CRD is installed when using this? https://github.com/kubeflow/manifests/blob/master/example/kustomization.yaml

annajung: Hi Suraj, I believe it comes from here https://github.com/kubeflow/manifests/blob/847faeb5533b5f76423f756c75a41452f302929f/example/kustomization.yaml#L29 which leads to https://github.com/kubeflow/manifests/blob/847faeb5533b5f76423f756c75a41452f302929f/apps/pipeline/upstream/env/platform-agnostic-multi-user/kustomization.yaml#L13

surajkota: Hi Anna, thanks. I missed the `cd apps/pipeline/upstream` in the script before when I posted the comment but forgot to resolve it afterwards.

Issue #2194, opened 2022-04-26
KF Release 1.6 tracking


annajung:
This issue will be used to provide high-level updates for the Kubeflow 1.6 release.

- [Release Timeline](https://github.com/kubeflow/community/blob/master/releases/release-1.6/README.md)
- [Release Team](https://github.com/kubeflow/community/blob/master/releases/release-1.6/release-team.md)
- [Release Board](https://github.com/orgs/kubeflow/projects/49/views/1)

cc @kubeflow/release-team

annajung:
- [Docs tracking issue](https://github.com/kubeflow/website/issues/3239)
- WG tracking issues
  - [AutoML](https://github.com/kubeflow/manifests/issues/2195)
  - [Training Operator](https://github.com/kubeflow/manifests/issues/2196)
  - [KServe](https://github.com/kubeflow/manifests/issues/2197)
  - [Pipelines](https://github.com/kubeflow/manifests/issues/2198)
  - [Notebook](https://github.com/kubeflow/manifests/issues/2199)
  - [Manifests](https://github.com/kubeflow/manifests/issues/2200)

annajung:
Discussion around dependency versions for 1.6 https://github.com/kubeflow/manifests/issues/2207

Tomcli:
For kfp-tekton project, we want to use kfp-tekton  [v1.2.1](https://github.com/kubeflow/kfp-tekton/releases/tag/v1.2.1) release for KF v1.6. The update script should be available at https://github.com/kubeflow/manifests/blob/master/hack/sync-kfp-tekton-manifests.sh. Thanks.

annajung:
Due to a blocker that was identified last week, we are delaying the release for 2 weeks, with a possibility of another extension

- Tracking issue: https://github.com/kubeflow/testing/issues/1006
- Email announcement: https://groups.google.com/u/2/g/kubeflow-discuss

annajung:
After checking with all WGs, the release team is delaying the release for 2 more weeks. The new feature freeze date is **Wednesday, Jun 29th, 2022**.

- Email announcement: https://groups.google.com/g/kubeflow-discuss/c/I4l97HvrGEA/m/227aCe_mCgAJ
- New schedule PR: https://github.com/kubeflow/community/pull/562

annajung:
Kubeflow 1.6.0-rc.0 is now available!
- release tag: https://github.com/kubeflow/manifests/releases/tag/v1.6.0-rc.0
[...]

PR #2255, opened 2022-07-20
Update metacontroller

gkcalat:

- KF 1.6 release target k8s v 1.22: #2194
- apiextensions.k8s.io/v1beta1 has been removed in k8s 1.22.

- Upgraded metacontroller to stay in sync for Kubeflow Pipelines.
- Added README with short instructions on how to manually upgrade it in the future.

**Checklist:**
- [ ] Unit tests pass:
    2. `make test`

### DIFF
--- a/contrib/metacontroller/README.md
+++ b/contrib/metacontroller/README.md
@@ -5,5 +5,12 @@
 
 ## Upgrade
 
-To update this component pull the latest content from [Kubeflow Pipelines third-party folder](https://github.com/kubeflow/pipelines/tree/master/manifests/kustomize/third-party/metacontroller).
+Metacontroller is pulled from [Kubeflow Pipelines third-party folder](https://github.com/kubeflow/pipelines/tree/master/manifests/kustomize/third-party/metacontroller). To update this component specify the desired KFP version and run the following command in console from the root directory **/**:
+
+```bash
+export KFP_VERSION=2.0.0-alpha.3   # specify KFP version
+kpt pkg update ./contrib/metacontroller@${KFP_VERSION}
+```
+
+Alternatively, you can copy the content from [Kubeflow Pipelines third-party folder](https://github.com/kubeflow/pipelines/tree/master/manifests/kustomize/third-party/metacontroller) by choosing the appropriate `TAG` in that repository.
 
--- a/contrib/metacontroller/README.md
+++ b/contrib/metacontroller/README.md
@@ -1,8 +1,9 @@
 # Metacontroller
 
-- [Official documentation] (https://metacontroller.github.io/metacontroller/)
+- [Official documentation](https://metacontroller.github.io/metacontroller/)
 - [Official repoitory](https://github.com/metacontroller/metacontroller)
 
 ## Upgrade
 
 To update this component pull the latest content from [Kubeflow Pipelines third-party folder](https://github.com/kubeflow/pipelines/tree/master/manifests/kustomize/third-party/metacontroller).
+
--- a/contrib/metacontroller/README.md
+++ b/contrib/metacontroller/README.md
@@ -1,0 +1,8 @@
+# Metacontroller
+
+- [Official documentation] (https://metacontroller.github.io/metacontroller/)
+- [Official repoitory](https://github.com/metacontroller/metacontroller)
+
+## Upgrade
+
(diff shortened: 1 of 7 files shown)


contrib/metacontroller/README.md
```diff
@@ -0,0 +1,9 @@
+# Metacontroller
+
+- [Official documentation](https://metacontroller.github.io/metacontroller/)
+- [Official repoitory](https://github.com/metacontroller/metacontroller)
+
+## Upgrade
+
+To update this component pull the latest content from [Kubeflow Pipelines third-party folder](https://github.com/kubeflow/pipelines/tree/master/manifests/kustomize/third-party/metacontroller).
```
zijianjoy, line 8: Can we use an environment variable `KFP_VERSION` to make sure we fetch the correct version of metacontroller?

For example:
```
export KFP_VERSION=2.0.0-alpha.3
kpt pkg get "https://github.com/kubeflow/pipelines/tree/${KFP_VERSION}/manifests/kustomize/third-party/metacontroller" ./contrib/metacontroller
```

This is just an example script which I haven't validated, we can also just ask people to pull from a tag manually.

gkcalat: This is a great idea. Thank you!

zijianjoy: Thank you for the update!

zijianjoy:
/lgtm
/approve

Looks great, thank you so much Ablai!

kimwnasptd:
Thanks for keeping this up to date @gkcalat @zijianjoy!

One small questions that come to mind:
1. why do we need to have the metacontroller under `contrib`?
2. what's the difference with the metacontroller in https://github.com/kubeflow/manifests/tree/master/apps/pipeline/upstream/third-party/metacontroller/base?
3. Right now in our [example kustomization](https://github.com/kubeflow/manifests/blob/master/example/kustomization.yaml) we are not installing that package, since KFP is using the metacontroller bundled with them (link in point 2). Should we change this?
4. Do we need to have both the metacontroller under `contrib` and inside the KFP manifests?

zijianjoy:
@kimwnasptd 

Good questions: My understanding is that Metacontroller makes it easy to write custom controller in any kubernetes cluster. If you confirm that metacontroller is currently not used in anywhere else except KFP, I think it doesn't matter we have metacontroller by itself in `contrib/` folder or within KFP manifests at the moment. However, it might cause implicit conflict if any other Kubeflow component or users decide to adopt Metacontroller. We will need to reverse the process: Extract metacontroller back to `contrib/`. So I feel that the current structure is more future-proof.

kimwnasptd:
Thanks for the explanation! 

I'm OK with merging this update for now. I made these questions because I want to create a proposal around expectations regarding the `/contrib` folder. Mostly to ensure each component should have a clear description of:
1. How could someone use it with Kubeflow, and examples
3. Why someone would use it [problems it tries to solve]
4. Status of the KF versions that the component is compatible with

But, again, this will be a proposal of its own. Just wanted to have a rough idea for the state of metacontroller for now.

/lgtm
/approve

zijianjoy:
@kimwnasptd Nice idea! I think the three points you made regarding clear description is a good start. Once we reach consent in community, feel free to let me know and I can provide content for the three points for metacontroller.

annajung:
Hi @zijianjoy @gkcalat 

As Kimonas pointed out in his [comment](https://github.com/kubeflow/manifests/pull/2255#issuecomment-1194391058), this PR does not change the default behavior of which metacontroller gets installed since the default installation points to the [`/third-party`](https://github.com/kubeflow/manifests/blob/746c523e9e994154df48d4441d01a30e4f9c8de4/apps/pipeline/upstream/env/platform-agnostic-multi-user/kustomization.yaml#L13). 

I also noticed that `/third-party/metacontroller/` already contains the changes you have made in the `/contrib` as part of this PR. Unless you want to change the default installation to use the `/contrib`, it doesn't seem like this has any behavior changes for 1.6. 

That said, I don't think we need to cut another 1.6 RC for this change only. 

Since we are planning to cut a new RC in the future for the notebook WG ([release blocking issue](https://github.com/kubeflow/kubeflow/issues/6572#issuecomment-1194513442)), we can include this change at that point if you like to continue your work to change the default installation to use the metacontroller from the `/contrib` 

Let me know what you think! thanks!

PR #2256, opened 2022-07-20
Sync kserve web app manifests v0.8.1

kimwnasptd:
Did a couple of changes to wrap up the work around the web app in this PR and testing. Specifically:
1. I created a new patch release for the MWA https://github.com/kserve/models-web-app/tree/v0.8.1/config
2. Created a distinct script for the MWA, for syncing the manifests, and decoupled it from a single one for doing both the app and kserve
3. Doubled the timeout in the e2e test script. I noticed that sometimes even when everything would run as expected 5mins wasn't enough for some steps
4. Update the cleanup script, in the e2e test script, to use `kserve.io` and not `kubeflow.org` (we had missed this place)

cc @annajung @surajkota

### DIFF
--- a/contrib/kserve/models-web-app/base/istio.yaml
+++ b/contrib/kserve/models-web-app/base/istio.yaml
@@ -11,7 +11,7 @@ spec:
   http:
   - match:
     - uri:
-        prefix: /models/
+        prefix: /kserve-endpoints/
     rewrite:
       uri: /
     route:
--- a/contrib/kserve/models-web-app/base/kustomization.yaml
+++ b/contrib/kserve/models-web-app/base/kustomization.yaml
@@ -13,7 +13,7 @@ commonLabels:
 images:
 - name: kserve/models-web-app
   newName: kserve/models-web-app
-  newTag: v0.7.0
+  newTag: v0.8.0
 configMapGenerator:
   - name: kserve-models-web-app-config
     literals:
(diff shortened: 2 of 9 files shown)


kimwnasptd:
I also run the e2e script and it completed, plus the web apps worked as expected.

contrib/kserve/models-web-app/base/istio.yaml
```diff
@@ -11,7 +11,7 @@ spec:
   http:
   - match:
     - uri:
-        prefix: /models/
+        prefix: /kserve-endpoints/
```
annajung, line 14: Does this mean that the [notebook configmap](https://github.com/kubeflow/manifests/blob/master/apps/centraldashboard/upstream/base/configmap.yaml#L30) should change as well? I was able to successfully test out standalone model web app, but with the full kubeflow installation, I am just getting a blank page for models page which points to `http://localhost:8080/_/kserve-endpoints/?ns=kubeflow-user-example-com`

contrib/kserve/models-web-app/base/kustomization.yaml
```diff
@@ -13,7 +13,7 @@ commonLabels:
 images:
 - name: kserve/models-web-app
   newName: kserve/models-web-app
-  newTag: v0.7.0
+  newTag: v0.8.0
```
annajung, line 16: There is a mismatch between the tag here (0.8.0) and the one listed in the readme (0.8.1)

contrib/kserve/models-web-app/base/istio.yaml
kimwnasptd: Can you confirm you are also using the `kserve` overlay when deploying the CentralDashboard? We had introduced this in the previous release to support both KServe and KFServing.

https://github.com/kubeflow/manifests/blob/v1.6-branch/example/kustomization.yaml#L33

contrib/kserve/models-web-app/base/kustomization.yaml
kimwnasptd: Indeed, but there wasn't any change in the image of the models web app between `v0.8.0` and `v0.8.1` so I think it's OK

contrib/kserve/models-web-app/base/istio.yaml
annajung: Thanks for the clarification Kimonas! I retested with your branch and got it up and running successfully with full installation! Sorry for the confusion!

PR #2259, opened 2022-07-26
Solve the problem relating to Kubeflow authservice-0 permission denied

thkim0022:
Resolves #

Add the `initContainers` in `spec.template.spec` at `common/oidc-authservice/base/statefulset.yaml`

**Checklist:**
- [ ] Unit tests pass:
    2. `make test`


juliusvonkohout:
This looks very bad. 
Never ever use root/superuser commands during runtime. In my instance (1.5.1) it runs without problems `uid=1000(1000) gid=0(root) groups=111(authservice)` so far.

PR #2260, opened 2022-07-27
Update kubeflow/kubeflow manifests from v1.6.0-rc.1

kimwnasptd:
Last work item for cutting the RC1 in manifests

I've also verified that the `test/e2e` script is working as expected. Profiles are being created and the web apps work as expected.

cc @annajung


majidhali:
After cloning the manifest with the latest fixes I got the below error:


2022/07/28 08:11:53 well-defined vars that were never replaced: kfp-app-version,kfp-app-name
Error: map[string]interface {}(nil): yaml: unmarshal errors:
  line 66: mapping key "livenessProbe" already defined at line 56
  line 78: mapping key "ports" already defined at line 62
error: no objects passed to apply
Retrying to apply resources

YGuyomar:
@majidhali 
apps/profiles/upstream/manager/manager.yaml contains a duplicate definition of liveness probe & ports

majidhali:
Thanks @YGuyomar I used the old profiles with the updated Jupyter directory and problem finally solved!

jceb:
> @majidhali apps/profiles/upstream/manager/manager.yaml contains a duplicate definition of liveness probe & ports

The issue still seems to persist.

Issue #2225, opened 2022-06-16
Could not find CSRF cookie XSRF-TOKEN in the request


wangzheyuan:
Following #https://github.com/kubeflow/manifests/pull/2155, I can create notebooks, tensorbroads and volume over HTTP successfully. But when creating a model server over HTTP, I encountered the error.
[403] Could not find CSRF cookie XSRF-TOKEN in the request. http://172.16.40.233:30828/kserve-endpoints/api/namespaces/kubeflow-user-example-com/inferenceservices
![16553695433980](https://user-images.githubusercontent.com/23622441/174042792-06441f42-9102-4cfa-b247-b853f5cb8e23.png)

PR #2262, opened 2022-08-02
add APP_SECURE_COOKIES in environment variable for KServe Models App.

wangzheyuan:
Resolves https://github.com/kubeflow/manifests/issues/2225

Following  https://github.com/kubeflow/kubeflow/pull/6395, I can create notebooks, tensorbroads and volume over HTTP successfully. But when creating a model server over HTTP, I encountered the error `Could not find CSRF cookie XSRF-TOKEN in the request`. 
I added  APP_SECURE_COOKIES=false in environment variable for KServe Models App, then I can create model servers over HTTP.

### DIFF
--- a/.github/workflows/katib_kind_test.yaml
+++ b/.github/workflows/katib_kind_test.yaml
@@ -3,6 +3,8 @@ on:
   pull_request:
     paths:
       - apps/katib/upstream/**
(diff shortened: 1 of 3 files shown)


wangzheyuan:
@googlebot rescan

PR #2269, opened 2022-08-22
Extend Training Operator GH action to also apply a TFJob and assert it runs

NickLoukas:

### DIFF
--- a/.github/workflows/train_operator_kind_test.yaml
+++ b/.github/workflows/train_operator_kind_test.yaml
@@ -29,3 +29,8 @@ jobs:
         kubectl create ns kubeflow
         kustomize build overlays/kubeflow | kubectl apply -f -
         kubectl wait --for=condition=Ready pods --all --all-namespaces --timeout 180s
+
+    - name: Create a TFJob
+      run: |
+        kubectl create -f tests/gh-actions/kf-objects/tfjob.yaml
+        kubectl wait --for=condition=Succeeded tfjobs tfjob-simple -n kubeflow --timeout 180s
--- a/tests/gh-actions/kf-objects/tfjob.yaml
+++ b/tests/gh-actions/kf-objects/tfjob.yaml
@@ -1,0 +1,21 @@
+apiVersion: "kubeflow.org/v1"
+kind: TFJob
+metadata:
+  name: tfjob-simple
+  namespace: kubeflow
+spec:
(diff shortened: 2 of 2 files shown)


.github/workflows/train_operator_kind_test.yaml
```diff
@@ -29,3 +29,8 @@ jobs:
         kubectl create ns kubeflow
         kustomize build overlays/kubeflow | kubectl apply -f -
         kubectl wait --for=condition=Ready pods --all --all-namespaces --timeout 180s
+
+    - name: Create a TFJob
+      run: |
+        kubectl create -f tests/gh-actions/kf-objects/tfjob.yaml
+        kubectl wait --for=condition=Running tfjobs tfjob-simple -n kubeflow --timeout 300s
```
kimwnasptd, line 36: Doesn't `kubectl wait --for=condition=Running` mean that the script will wait until the TFJob starts running?

We should instead wait until it completes

NickLoukas: I changed the condition check to `Succeeded`.

NickLoukas:
Hey @johnugeorge & @andreyvelich! We are trying to use a TFJob in a GH action to test the Training Operator and we're currently using this sample (https://github.com/kubeflow/training-operator/blob/master/examples/tensorflow/simple.yaml). 
The problem is that it take too much time for the TFJob to succeed and we would like to ask if you could provide us with an example that completes as fast as possible.

johnugeorge:
We use this https://github.com/kubeflow/training-operator/blob/master/sdk/python/test/e2e/test_e2e_tfjob.py#L33 in our GHA

NickLoukas:
I've sent a fixup for the commit and changed the TFJob yaml. Now the Job completes successfully.

kimwnasptd:
Thanks @NickLoukas! Can you add a dummy commit here and create a new file to trigger the GH Action?

Just to verify that everything works as expected in the repo as well

NickLoukas:
I added a dummy commit with a test file and everything works as expected here as well.

kimwnasptd:
Looks good! Let's remove this commit and we can merge

NickLoukas:
Just rebased and dropped the test commit.

kimwnasptd:
Thanks for the work!

/lgtm
/approve

PR #2270, opened 2022-08-22
Extend Pipelines GH action to also deploy a pipeline and assert it runs

NickLoukas:
In the pipelines GH actions when we try to deploy a pipeline with the `kfp` command we get the error bellow:
```
Failed to automatically set namespace.
(400)
Reason: Bad Request
HTTP response headers: HTTPHeaderDict({'Audit-Id': 'd67e99dd-88d7-4892-96c2-68d4001764dd', 'Cache-Control': 'no-cache, private', 'Content-Length': '553', 'Content-Type': 'application/json', 'Date': 'Tue, 02 Aug 2022 16:23:59 GMT'})
HTTP response body: {"error":"Invalid input error: Invalid resource references for experiment. ListExperiment requires filtering by namespace.","code":3,"message":"Invalid input error: Invalid resource references for experiment. ListExperiment requires filtering by namespace.","details":[{"@type":"type.googleapis.com/api.Error","error_message":"Invalid resource references for experiment. ListExperiment requires filtering by namespace.","error_details":"Invalid input error: Invalid resource references for experiment. ListExperiment requires filtering by namespace."}]}
```

The error is produced in this line `kfp run submit -e test_pipeline -f ./tests/gh-actions/test_pipeline.py.yaml`.


NickLoukas:
I added an extra commit and fixed the error above. Now we use the Client and create the pipeline run directly inside the python script. Also we install the `profiles` and create a kubeflow-user to use for the pipeline deployment.

kimwnasptd:
Thanks @NickLoukas! Can you add a dummy commit here and create a new file to trigger the GH Action?

Just to verify that everything works as expected in the repo as well

PR #2271, opened 2022-08-22
Fix typo in Katib GH action

NickLoukas:
This is a follow up PR to this (https://github.com/kubeflow/manifests/pull/2249) to fix the path of the katib test.

### DIFF
--- a/.github/workflows/katib_kind_test.yaml
+++ b/.github/workflows/katib_kind_test.yaml
@@ -35,6 +35,6 @@ jobs:
 
     - name: Create katib experiment
       run: |
-        kubectl apply -f tests/gh-actions/katib_test.yaml
+        kubectl apply -f tests/gh-actions/kf-objects/katib_test.yaml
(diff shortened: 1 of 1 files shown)


.github/workflows/katib_kind_test.yaml
```diff
@@ -35,6 +35,6 @@ jobs:
 
     - name: Create katib experiment
       run: |
-        kubectl apply -f tests/gh-actions/katib_test.yaml
+        kubectl apply -f tests/gh-actions/kf-objects/katib_test.yaml
```
annajung, line 38: Looks like the resources are created under `kubeflow-user` namespace which is not created before. In addition, it looks for a label to be enabled to create the experiment.

Running the follow commands before applying this yaml file made it work for me. Otherwise, it will throw an error for missing namespace

```
kubectl create namespace kubeflow-user
kubectl label namespace kubeflow-user katib.kubeflow.org/metrics-collector-injection=enabled
```

kimwnasptd: Good catch! I'll send a follow up PR asap to fix this

PR #2287, opened 2022-09-27
Update kubeflow/pipelines manifests from 2.0.0-alpha.5

annajung:
Signed-off-by: Anna Jung (VMware) <antheaj@vmware.com>

- Sync pipelines to 2.0.0-alpha.5 which includes a fix for https://github.com/kubeflow/pipelines/issues/8256

cc @kubeflow/release-team @chensun @kimwnasptd

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ This repo periodically syncs all official Kubeflow components from their respect
 | Katib | apps/katib/upstream | [v0.14.0-rc.0](https://github.com/kubeflow/katib/tree/v0.14.0-rc.0/manifests/v1beta1) |
 | KServe | contrib/kserve/kserve | [release-0.8](https://github.com/kserve/kserve/tree/8079f375cbcedc4d45a1b4aade2e2308ea6f9ae8/install/v0.8.0) |
 | KServe Models Web App | contrib/kserve/models-web-app | [v0.8.1](https://github.com/kserve/models-web-app/tree/v0.8.1/config) |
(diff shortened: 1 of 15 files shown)


tests/gh-actions/install_pipelines.sh
```diff
@@ -7,4 +7,4 @@ kubectl apply -f third-party/metacontroller/base/crd.yaml
 echo "Waiting for crd/compositecontrollers.metacontroller.k8s.io to be available ..."
 kubectl wait --for condition=established --timeout=30s crd/compositecontrollers.metacontroller.k8s.io
 kustomize build env/cert-manager/platform-agnostic-multi-user | kubectl apply -f -
-kubectl wait --for=condition=Ready pods --all --all-namespaces --timeout 180s
\ No newline at end of file
+kubectl wait --for=condition=Ready pods --all --all-namespaces --timeout 300s
```
annajung, line 10: @kimwnasptd heads up that I had to increase the time to allow pods to be ready and pass this condition. When testing locally, it took about ~4-5 mins for all pods to come up for me

kimwnasptd: @annajung let's actually double that, to have some leeway. Let's do 600s

annajung: Done - I agree that increasing the timeout doesn't hurt since when the condition is all met, it completes.

kimwnasptd:
Thanks!

/lgtm

PR #2296, opened 2022-10-11
Harden the istio configuration

### LINKED ISSUE #2285: Insecure Istio configuration in Kubeflow 1.6
istio-ingressgateway service is set to [type LoadBalancer in 1.6 by default](https://github.com/kubeflow/manifests/blob/master/common/istio-1-14/istio-install/base/install.yaml#L3603) which will open a external endpoint to the service without a TLS/SSL certificate if a user has any load balancer controller installed in their cluster. This is unexpected and undocumented change in behaviour for users and is also insecure depending on the defaults the load balancer controller will use. (e.g. on AWS, if you have aws-load-balancer-controller installed, it will open HTTP port as well and hence we would not recommend anyone to install 1.6 from this repository without modifications)

There are a few more customisations documented here which are good security practices and bug fixes but have not been applied in 1.6 - https://github.com/kubeflow/manifests/tree/master/common/istio-1-14#changes-to-the-upstream-manifests-using-kustomize. It is worth considering applying these for 1.6 as well.

[1.6](https://github.com/kubeflow/manifests/blob/master/common/istio-1-14/istio-install/base/kustomization.yaml) vs [1.5](https://github.com/kubeflow/manifests/blob/v1.5-branch/common/istio-1-11/istio-install/base/kustomization.yaml#L5-L15)

juliusvonkohout:
The istio nodeport (loadbalancer)instead of a proper ClusterIP seems dangerous to me https://oteemo.com/think-nodeport-kubernetes/
Fixes #2285
@kimwnasptd 

Resolves #

**Checklist:**
- [ ] Unit tests pass:
    2. `make test`

### DIFF
--- a/common/istio-1-14/istio-install/base/patches/disable-debugging.yaml
+++ b/common/istio-1-14/istio-install/base/patches/disable-debugging.yaml
@@ -1,0 +1,18 @@
+# Pentest enahncement: check port 15010 & 8080 in istiod: According to https://istio.io/latest/docs/ops/best-practices/security/#control-plane port 15010
+# is not that problematic (only resource discovery). Other parts of the documentation also say| 15010 | GRPC | XDS and CA services (Plaintext, only for secure networks) | 
+# We have a secure network layer and only XDS is served. 
+# Port 8080 is not listed in the service and even if it would be somehow reachable by IP it only "offers read access". 
+# Nevertheless we set ENABLE_DEBUG_ON_HTTP=false do disable it entirely.
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: istiod
+  namespace: istio-system
+spec:
+  template:
+    spec:
+      containers:
+      - name: discovery
+        env:
+        - name: ENABLE_DEBUG_ON_HTTP
+          value: 'false'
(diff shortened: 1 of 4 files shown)


common/istio-1-14/istio-install/base/patches/disable-debugging.yaml
```diff
@@ -0,0 +1,18 @@
...
+metadata:
+  name: istiod
+  namespace: istio-system
+spec:
+  template:
+    spec:
+      containers:
+      - name: discovery
+        env:
+        - name: ENABLE_DEBUG_ON_HTTP
```
kimwnasptd, line 17: @juliusvonkohout does the `ENABLE_DEBUG_ON_HTTP` only affect the (deprecated) 8080 port, that has the `debugging interface` right?

juliusvonkohout: It does definitely not affect port 80 which serves the istio-ingressgateway and therefore centraldashboard. It was tested for several months on istio-cni. https://istio.io/latest/docs/ops/best-practices/security/#control-plane says "Port 8080 exposes the debug interface, which offers read access to a variety of details about the clusters state. This can be disabled by set the environment variable ENABLE_DEBUG_ON_HTTP=false on Istiod. Warning: many istioctl commands depend on this interface and will not function if it is disabled."

juliusvonkohout:
Closed in favor of https://github.com/kubeflow/manifests/pull/2357 since the master branch is now at 1.16

PR #2324, opened 2022-11-22
Use K8s 1.25 for the tests

apo-ger:
Updating the KinD cluster version to K8s 1.25


kimwnasptd:
Thanks @apo-ger!

/lgtm
/approve

PR #2322, opened 2022-11-21
Add unmaintained notice to contrib seldon

annajung:
Signed-off-by: Anna Jung (VMware) <antheaj@vmware.com>

**Which issue is resolved by this Pull Request:**
Part of https://github.com/kubeflow/manifests/issues/2311

**Description of your changes:**
Add README to `/contrib/seldon` to mark the directory as unmaintained and out of date

**Note:**
Tagging seldon component [owners](https://github.com/kubeflow/manifests/blob/master/contrib/seldon/OWNERS) @axsaucedo @adriangonz @cliveseldon @ryandawsonuk to ensure awareness of the out-of-date label on the component.

To remove the unmaintained label, check out the missing requirements https://github.com/kubeflow/manifests/issues/2311 and leave a comment on the issue to indicate that you'll take action to provide the missing requirements and are willing to keep the component up to date.

cc @kimwnasptd @kubeflow/wg-manifests-leads

Issue #2311, opened 2022-11-14
Tracking for `/contrib` components evaluation and clean up


annajung:
To enable new components to integrate with Kubeflow, a [contrib component guidelines](https://github.com/kubeflow/manifests/blob/master/proposals/20220926-contrib-component-guidelines.md) have been added with a list of [requirements](https://github.com/kubeflow/manifests/blob/master/proposals/20220926-contrib-component-guidelines.md#component-requirements).

To enforce the requirements for existing components, each component will be evaluated to determine its status. In this process, any components that do not meet the requirements will be marked as `UNMAINTAINED/OUT-OF-DATE` and eventually removed from the [manifest](https://github.com/kubeflow/manifests) repo as stated in the [deprecation plan](https://github.com/kubeflow/manifests/blob/master/proposals/20220926-contrib-component-guidelines.md#deprecation-plan).

In discussion with Manifest WG @kimwnasptd, this issue will be used to track progress for each component. 

| Component | Evaluation Status | Meets Requirement | Missing Requirements | PR to label as unmaintained | PR to remove (After 1.7 release) |
| -- | -- | -- | -- | -- | -- |
| [application](https://github.com/kubeflow/manifests/tree/master/contrib/application) | Completed | No | All | https://github.com/kubeflow/manifests/pull/2313 (merged) | https://github.com/kubeflow/manifests/pull/2430 |
| [basic-auth](https://github.com/kubeflow/manifests/tree/master/contrib/basic-auth) | Completed | No | All | https://github.com/kubeflow/manifests/pull/2313/ (merged) | https://github.com/kubeflow/manifests/pull/2430 |
| [dex-auth](https://github.com/kubeflow/manifests/tree/master/contrib/dex-auth) | Completed | No | - `README.md` is outdated <br> - Component does not work with the latest version of Kubeflow <br> - Missing `UPGRADE.md` <br> - No test <br> - Outdated (2 years) | https://github.com/kubeflow/manifests/pull/2316 (merged) | https://github.com/kubeflow/manifests/pull/2430 |
| [experimental](https://github.com/kubeflow/manifests/tree/master/contrib/experimental) | Completed | No | All | https://github.com/kubeflow/manifests/pull/2313/ (merged) | https://github.com/kubeflow/manifests/pull/2430 |
[...]

PR #2326, opened 2022-11-22
Updated and tested Seldon Core 1.15.0

axsaucedo:
Signed-off-by: Alejandro Saucedo <axsauze@gmail.com>

Resolves part of #2311 addressing #2322.

* Adding requirements as per [Guidelines for /contrib Components](https://github.com/kubeflow/manifests/blob/master/proposals/20220926-contrib-component-guidelines.md)
* Fully tested kubeflow installation of Seldon Core's latest version 1.15.0
* Adding UPGRADE.md
* Adding automated tests (via `make -C contrib/seldon/ test`)
* Adding example testing for manual interaction

## Testing

A section was added under `contrib/seldon/README.md` containing the instructions to test, which are below.

It is now possible to run the automated tests with `make -C contrib/seldon/ test`.

## Explicit overview of manual install & testing 

Installation can be done with the kustomize command:

```
kustomize build seldon-core-operator/base | kubectl apply -n kubeflow -f -
```

We can create a test model once the "Install Seldon Operator" is configured

```
# Create namespace for model
kubectl create namespace seldon
```

We can create an echo model with the following command:

```
kubectl apply -f - << ENDapiVersion: machinelearning.seldon.io/v1
kind: SeldonDeployment
metadata:
  name: echo
  namespace: seldon
spec:
  predictors:
  - name: default
    replicas: 1
    graph:
      name: classifier
      type: MODEL
    componentSpecs:
    - spec:
        containers:
        - image: seldonio/echo-model:1.15.0-dev
          name: classifier
END
```

We can verify that model is running:

```
kubectl get pods -n seldon

NAME                                         READY   STATUS    RESTARTS   AGE
echo-default-0-classifier-679cb5fb68-qd4nm   2/2     Running   0          25m
```

Also we can verify that the correct virtualservice was created:

```
kubectl get virtualservice -n seldon

NAME   GATEWAYS                        HOSTS   AGE
echo   ["kubeflow/kubeflow-gateway"]   ["*"]   42m
```

Finally we can send a request (you will need to fetch the Dex Auth Token / Cookie):

```
export CLUSTER_IP=# Your cluster IP
export SESSION=# Your dex session

curl -H "Content-Type: application/json" -H "Cookie: authservice_session=${SESSION}" \
   -d '{"data": {"ndarray":[[1.0, 2.0, 5.0]]}}' \
   http://{CLUSTER_IP}/seldon/seldon/echo/api/v1.0/predictions

{"data":{"names":["t:0","t:1","t:2"],"ndarray":[[1.0,2.0,5.0]]},"meta":{"metrics":[{"key":"mycounter","type":"COUNTER","value":1},{"key":"mygauge","type":"GAUGE","value":100},{"key":"mytimer","type":"TIMER","value":20.2}],"requestPath":{"classifier":"seldonio/echo-model:1.15.0-dev"}}}
```


axsaucedo:
Thank you @juliusvonkohout - I have signed the CLI a while back (the https://cla.developers.google.com/clas confirms I have valid CLA), it seems the job is still "waiting for status", and I believe I saw it pass in a previous commit, is there a way to re-trigger the check?

annajung (commented):
Thanks for the prompt response in getting the component updated to work with the latest kubeflow release!

contrib/seldon/Makefile
```diff
@@ -10,17 +10,35 @@
...
+	sed -i "s#'seldon-webhook-service.kubeflow.svc'#"'seldon-webhook-service.$$(SERVICE_NAMESPACE).svc#' seldon-core-operator/base/resources.yaml
 	sed -i 's#kubeflow/kubeflow-gateway#$$(SERVICE_NAMESPACE)/kubeflow-gateway#' seldon-core-operator/base/resources.yaml
 	cp kustomization.tpl seldon-core-operator/base/kustomization.yaml
 	cp kustomizeconfig.tpl seldon-core-operator/base/kustomizeconfig.yaml
 
 .PHONY:clean-kustomize
 clean-kustomize:
 	rm -rf seldon-core-operator/base
+
+test:
```
annajung, line 28: @kimwnasptd I think this should be sufficient to meet the test requirement for contrib, do we want to encourage to add the test into github actions workflow as well?

juliusvonkohout:
I think Seldon also has proper support for PodSecurityStandards restricted. At least that was the case for the previous version. Nevertheless that is something that should also be tested.

contrib/seldon/UPGRADE.md
```diff
@@ -0,0 +1,5 @@
+# Upgrading Documentation
```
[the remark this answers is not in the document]
axsaucedo: Ok good point, I have added this in the upgrading docs including an env var so it can be done intuitively

contrib/seldon/README.md
```diff
@@ -14,5 +14,75 @@ kustomize build seldon-core-operator/base | kubectl apply -n kubeflow -f -
 This kustomize spec was created from the seldon-core-operator helm chart with:
```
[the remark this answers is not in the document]
axsaucedo: Ok I've now added a one-liner explaining what it does, let me know your thoughts

contrib/seldon/README.md
```diff
@@ -14,5 +14,75 @@ kustomize build seldon-core-operator/base | kubectl apply -n kubeflow -f -
...
 
 ```
-make clean seldon-core-operator/base
+make clean-kustomize seldon-core-operator/base
 ```
+
+## Testing
+
+```
+make test
```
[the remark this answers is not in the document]
axsaucedo: My apologies, it seems the latest changes weren't updated, it should now work correctly

contrib/seldon/Makefile
```diff
@@ -10,17 +10,35 @@
...
 	#kfctl strangeness means these variables don't always get resolved see https://github.com/kubeflow/manifests/issues/1660
 	#sed -i 's#cert-manager.io/inject-ca-from:.*#cert-manager.io/inject-ca-from: $$(CERTIFICATE_NAMESPACE)/$$(CERTIFICATE_NAME)#g' seldon-core-operator/base/resources.yaml
 	sed -i 's#cert-manager.io/inject-ca-from:.*#cert-manager.io/inject-ca-from: kubeflow/seldon-serving-cert#g' seldon-core-operator/base/resources.yaml
-	sed -i "s#'seldon-webhook-service.kubeflow.svc.cluster.local'#"'$$(SERVICE_NAME).$$(SERVICE_NAMESPACE).svc.cluster.local#' seldon-core-operator/base/resources.yaml
-	sed -i "s#commonName: 'seldon-webhook-service.kubeflow.svc'#"'commonName: $$(SERVICE_NAME).$$(CERTIFICATE_NAMESPACE).svc#' seldon-core-operator/base/resources.yaml
-	sed -i "s#'seldon-webhook-service.kubeflow.svc'#"'$$(SERVICE_NAME).$$(SERVICE_NAMESPACE).svc#' seldon-core-operator/base/resources.yaml
+	sed -i "s#'seldon-webhook-service.kubeflow.svc.cluster.local'#"'seldon-webhook-service.$$(SERVICE_NAMESPACE).svc.cluster.local#' seldon-core-operator/base/resources.yaml
```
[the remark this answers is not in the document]
axsaucedo: Ok thanks for the heads up, I've now added explicit dependency requirements in the UPGRADING.md, and referenced it in the README.md.

axsaucedo:
Ok it seems the CLA is passing now @juliusvonkohout.

image


@annajung thank you for the comments, these should now be resolved - let me know if anything else required.

contrib/seldon/Makefile
kimwnasptd: I'd propose that yes, let's have a very minimal test in place for the GH Actions. This will allow us to run these tests periodically to easily smoke test these components

kimwnasptd: @axsaucedo could you add a simple test workflow `.github/workflows/contrib_seldon_core.yaml` and define a minimal test case?

It can be something as simple as just verifying that the Pods of seldon core become ready. You can use as reference the rest of the apps that we are testing https://github.com/kubeflow/manifests/blob/master/.github/workflows/jwa_kind_test.yaml

If you want to test slightly more advanced cases, like creating a CR and testing if it also becomes Ready then that would also be more that welcome! But as a firs step testing that the core Pods are up should be enough

kimwnasptd: Actually, just running this `make test` inside of the GH Action should suffice

axsaucedo: Ok sounds good, let me add that

axsaucedo:
@kimwnasptd ok I have now added the test - for some strange reason I can't seem to get the "Create KinD Cluster" step to pass in the CI, it seems it doesn't find the file `tests/gh-actions/kind-cluster-1-22.yaml` - have you seen this before?

kimwnasptd:
@axsaucedo we recently changed slightly our files for installing KinD https://github.com/kubeflow/manifests/pull/2331.

I'd suggest to do a rebase to the latest code and use the [`tests/gh-actions/kind-cluster.yaml`](https://github.com/kubeflow/manifests/blob/master/tests/gh-actions/kind-cluster.yaml) file for setting up the KinD cluster.

axsaucedo:
Thanks @kimwnasptd I've now rebased and updated, but now I'm getting `./tests/gh-actions/kind-cluster.yaml: Permission denied` - is there any permission updates that need to be set up in order for this to run, or is it an incorrect path?

.github/workflows/seldon_kind_test.yaml
```diff
@@ -0,0 +1,29 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+
+    - name: Install KinD
+      run: ./tests/gh-actions/install_kind.sh
+
+    - name: Create KinD Cluster
+      run: ./tests/gh-actions/kind-cluster.yaml
```
kimwnasptd, line 18: @axsaucedo I believe our issue is in this line, we'll need to run kind and use that file for creating the cluster.

The command should be
```bash
kind create cluster --config tests/gh-actions/kind-cluster.yaml
```
Similarly to https://github.com/kubeflow/manifests/blob/master/.github/workflows/jwa_kind_test.yaml#L20

contrib/seldon/README.md
```diff
@@ -11,8 +13,74 @@ kustomize build seldon-core-operator/base | kubectl apply -n kubeflow -f -
 
 ## Updating
 
-This kustomize spec was created from the seldon-core-operator helm chart with:
+See [UPGRADING.md](UPGRADING.md)
```
kimwnasptd, line 16: We'll need to rename this here to `UPGRADE.md`, since now it will redirect to a 404 page

.github/workflows/seldon_kind_test.yaml
axsaucedo: Oh my mistake, not sure how but I assumed this was an `.sh` script, but definitely need to grab a coffee

contrib/seldon/README.md
axsaucedo: Good catch! I've updated now

.github/workflows/seldon_kind_test.yaml
kimwnasptd: Can definitely relate to that

axsaucedo:
@kimwnasptd it seems I've managed to get everything working locally, however I still get an error with a missing CRD, it seems that HPA CRD for v2beta1 is not set up - is this expected? Should I be looking to set it up as part of the Seldon tests?

The error in the operator is:

```
no matches for kind \"HorizontalPodAutoscaler\" in version \"autoscaling/v2beta1\
```

kimwnasptd:
I think the problem is in this line in the logs https://github.com/kubeflow/manifests/actions/runs/3549119873/jobs/5961121555
```
{"level":"error","ts":1669390311.322579,"logger":"controller.seldon-controller-manager","msg":"Reconciler error","reconciler group":"machinelearning.seldon.io","reconciler kind":"SeldonDeployment","name":"echo","namespace":"seldon","error":"no matches for kind \"HorizontalPodAutoscaler\" in version \"autoscaling/v2beta1\"","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.2/pkg/internal/controller/controller.go:266\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.2/pkg/internal/controller/controller.go:227"}
```

I see that in K8s 1.22 The `HorizontalPodAutoscaler` had also the version `v2beta1`, but in K8s 1.25 they only have `v2beta2`
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#horizontalpodautoscaler-v1-autoscaling
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.25/#horizontalpodautoscaler-v1-autoscaling

axsaucedo:
@kimwnasptd it seems that we may need to update some of the KIND tests as currently Seldon supports Kubernetes clusters up to 1.24 - currently we are in the process of upgrading the components, particularly HPA as per https://github.com/SeldonIO/seldon-core/pull/4172, however  have been careful as it would introduce breaking changes. Is there a way to set up the KIND cluster with kubernetes 1.24? I could add an extra config file to provide a pre-1-25-k8s kind test. Thank you  - let me know if you need more details.

axsaucedo:
@kimwnasptd thanks again for having a look, let me know your thoughts on the suggestion above, it does seem like the only way from our current state is to run the integration tests on a pre-1.25 k8s cluster, I can have a look at adding a `pre-1-24-...` kind setup config file if you think this can work (and add a note that emphasises the work + timeframe for post-1.25 support.

kimwnasptd:
@axsaucedo we discussed it with the release team as well in today's meeting (link to [notes](https://docs.google.com/document/d/1Cs9XYl_SHfGP9vs1eHyqb6JAcpE3ilM-5iRuBnV9AB4/edit#heading=h.qbxr0jfrv0f)).

In the KF 1.7 release we are targeting for K8s 1.25, but since we expect some vendors to not yet support this version we are also supporting K8s 1.24.

I'm OK with having this component working with K8s 1.24, but let's have a clear text somewhere at the top of the README that exposes the K8s version. So that it will be clear to users that will use KF 1.7 that this component will not work with 1.25 (to avoid confusion, since we will be promoting that the target K8s version for KF 1.7 will be K8s 1.25)

cc @annajung @DomFleischmann @jbottum

kimwnasptd:
So @axsaucedo to merge this PR could you:
1. Create a `tests/gh-actions/kind-cluster-1-24.yaml` configuration file for KinD and K8s 1.24
2. Update the GH Action for Seldon to use that yaml instead
3. Update the README to note that this works for K8s until 1.24

axsaucedo:
Ok perfect @kimwnasptd - very much appreciated. I will look at actioning these three points. Also absolutely agree on making a clear point in the README, as we also have a clear action plan towards ensuring seldon as a whole supports k8s 1.25+, we already have that PR suggested so it's a matter on how we can make sure we provide users with migration paths before we add it. 

In that case we'll add these changes as suggested and re-trigger the tests - I'll let you know if there are any issues / further questions. Thanks.

.github/workflows/seldon_kind_test.yaml
```diff
@@ -0,0 +1,32 @@
+name: Build & Apply Seldon manifests in KinD
+on:
+  pull_request:
+    paths:
+      - contrib/seldon/**
```
annajung, line 5: I think this should be okay for now, but @kimwnasptd we should think about other ways to trigger `/contrib` tests in the future and not just trigger based on changes to the `contrib/seldon` directory.

contrib/seldon/README.md
```diff
@@ -1,6 +1,14 @@
...
+# Seldon Core
 
-## Install Seldon Operator
+[Seldon Core](https://github.com/SeldonIO/seldon-core/) is a framework to deploy your machine learning models on Kubernetes at scale.
+
+# Requirements
+
+* Kubernetes 1.18 - 1.24
+
+Support for Kubernetes 1.25 is currently part of (SeldonIO/seldon-core#4172)[https://github.com/SeldonIO/seldon-core/pull/4172]
```
annajung, line 9:
```suggestion
Support for Kubernetes 1.25 is currently part of [SeldonIO/seldon-core#4172](https://github.com/SeldonIO/seldon-core/pull/4172)
```

contrib/seldon/README.md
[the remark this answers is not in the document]
annajung: verify that it works - thanks!

contrib/seldon/README.md
axsaucedo: Good catch - thank you for the commit suggest

README.md
```diff
@@ -274,7 +274,15 @@ Install the Katib official Kubeflow component:
 kustomize build apps/katib/upstream/installs/katib-with-kubeflow | kubectl apply -f -
 ```
 
-#### Central Dashboard
+#### Seldon
```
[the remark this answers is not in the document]
axsaucedo: Ok that makes sense, I have now reverted it to what it was

annajung:
Thanks for working with us to update Seldon to meet the contrib guidelines
/lgtm

axsaucedo:
Fantastic @annajung - thanks for the update, let me know if anything else needed fromy my side to merge 👍

.github/workflows/seldon_kind_test.yaml
kimwnasptd: Was thinking the same thing! We could probably use the `workflow_dispatch` event as well in these workflows, to be able to manually trigger them https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch

Let's play around with this more once we have the components update their manifests and have some basic tests in place.

kimwnasptd:
Was about to comment regarding the README.md, but @annajung had covered it already https://github.com/kubeflow/manifests/pull/2326#discussion_r1035086822, nice!

@axsaucedo thank you for your time and great work on updating these! The tests will also help us ensuring the quality of the manifests we bring along and even raise the standard.

/lgtm
/approve

example/kustomization.yaml
```diff
@@ -38,7 +38,7 @@ resources:
 # Notebook Controller
 - ../apps/jupyter/notebook-controller/upstream/overlays/kubeflow
 # Profiles + KFAM
-- ../apps/profiles/upstream/overlays/kubeflow
+# - ../apps/profiles/upstream/overlays/kubeflow
```
umka1332, line 41: Are there any reason to remove Profile controller from Kubeflow example installation?

juliusvonkohout:
@axsaucedo  why did you delete kubeflow-edit-seldon.yaml?

Issue #2325, opened 2022-11-22
K8s 1.25 and Manifests WG


kimwnasptd:
Tracking issue for all updates that we have to do in the common components of manifests to work with K8s 1.25. With a quick look we took with @apo-ger the biggest item is the deprecation of the `policy/v1beta1` API version, and the object that affects most of the components is `PodDisruptionBudget`

This is a draft plan of the common components and the target versions:
| Status | Component | Proposed Version | Context/PRs |
| --- | --- | ---| --- |
| ✅ | Istio | v1.16 | https://github.com/kubeflow/manifests/pull/2327 |
| ✅  | Knative | v1.8 | https://github.com/kubeflow/manifests/issues/2325#issuecomment-1325718698 https://github.com/knative/community/blob/main/mechanics/RELEASE-SCHEDULE.md#upcoming-releases|
| ✅ | Cert Manager | v1.10 | https://github.com/kubeflow/manifests/pull/2330 |

I'll also update the issue further with other changes we'll need to do for K8s 1.25

cc @kubeflow/wg-manifests-leads

apo-ger:
We noticed that the generated manifests of `Istio v1.6.0` include some `policy/v1` `PodDisruptionBudget` resources that we need to remove. (see https://github.com/istio/istio/issues/12602 and https://github.com/istio/istio/issues/24000)

Until now we have used two patches to address this:
- [common/istio-1-16/istio-install/base/patches/remove-pdb.yaml](https://github.com/kubeflow/manifests/blob/master/common/istio-1-14/istio-install/base/patches/remove-pdb.yaml)
- [common/istio-1-16/cluster-local-gateway/base/patches/remove-pdb.yaml](https://github.com/kubeflow/manifests/blob/master/common/istio-1-14/cluster-local-gateway/base/patches/remove-pdb.yaml)
    
However these patches do not work with kustomize `v3.2.0`. The problem is that `v3.2.0` doesn't have the appropriate openapi schema for the `policy/v1` API version resources:
- https://github.com/kubernetes-sigs/kustomize/issues/4495

This is fixed in `kustomize v4+`:
- https://github.com/kubernetes-sigs/kustomize/issues/3694#issuecomment-799700607
    
I opened a PR with the following proposed changes: 
- Disable the delete patches until we upgrade to `kustomize v4+`.
  - tracked in: https://github.com/kubeflow/manifests/issues/1797
 - As a temporary workaraound we remove PodDisruptionBudget resources
[...]

PR #2327, opened 2022-11-22
common: Add Istio v1.6.0 manifests

apo-ger:
Changes:
- Upgrade Istio to 1.16.0 to work with K8s 1.25
- Update kustomization file in example to deploy istio-1-16
- Update Istio GH action script to install Istio 1.16 for testing
- Introduce a workaround to remove `PodDisruptionBudget` resources manually with yq before deploying Istio manifests
  - See: https://github.com/kubeflow/manifests/issues/2325#issuecomment-1323909056
- Update README file with instructions

Refs: kubeflow/manifests#2325


juliusvonkohout:
@apo-ger  we would like to switch to istio-cni as discussed with @kimwnasptd to get rootless istio and apply restricted podsecuritystandards in the future. Here is the simple guide https://istio.io/latest/docs/setup/additional-setup/cni/

kimwnasptd:
@juliusvonkohout let's have a separate issue for integrating `istio-cni` in manifests. This way we can give a heads up to distros as well and make sure everyone is in sync with such a change. 

Also, I'd like to not block this PR with the `istio-cni` discussion since it's needed for running other K8s 1.25 tests https://github.com/kubeflow/manifests/pull/2330#issuecomment-1326287471 https://github.com/kubeflow/manifests/pull/2331#issuecomment-1326349823

kimwnasptd:
@apo-ger in the same spirit of https://github.com/kubeflow/manifests/pull/2330#issuecomment-1326271985 https://github.com/kubeflow/manifests/pull/2331#issuecomment-1326349823 let's also run the workflows that depend on Istio when the Istio version changes

kimwnasptd:
I think the reason that the workflows weren't triggered was because of the path:
```
      - apps/centraldashboard/upstream/**
      - /tests/gh-actions/install_istio.sh
```
Let's try to use relative paths for the install_istio.sh file as well to see if this fixes the issue

apo-ger:
Re-testing the GH action workflows since we merged https://github.com/kubeflow/manifests/pull/2331

PR #2286, opened 2022-09-27
proposal: Guidelines for /contrib components

kimwnasptd:
Create an initial proposal regarding the expectations we should have around the components in `/contrib` as well as a deprecation plan for unmaintained components.

I'd like to wait 2 weeks before merging this PR, to gather feedback on the proposal. Then once it's merged I'll do a pass on the components and create issues for the ones that don't meet the requirements. If there won't be any feedback in those issues, then the components will enter a deprecation phase.

@holdenk @Jeffwan @krishnadurai @yanniszark @juliusvonkohout @zijianjoy @gkcalat @zijianjoy @gkcalat @Jeffwan @woop @tedhtchang @Tomcli @animeshsingh @pvaneck @axsaucedo @adriangonz @cliveseldon @ryandawsonuk @ellistarn @yuzisun @cliveseldon @animeshsingh @deadeyegoodwin

cc-ing the release team as well @annajung @DomFleischmann @yubozhao @surajkota @DnPlas @jbottum

PR #2329, opened 2022-11-23
Update 20220926-contrib-component-guidelines.md

juliusvonkohout:
@kimwnasptd  @annajung this was missing from https://github.com/kubeflow/manifests/pull/2286#issuecomment-1273554536 and exspecially the kuberay integration shows that proper projects can easily implement it.


annajung:
I think this is a fair ask and definitely a good practice! If we like to enforce this during the [current clean up effort](https://github.com/kubeflow/manifests/issues/2311), we should try to get this merged soon & communicate this to /contrib component owners

cc @kimwnasptd

juliusvonkohout:
@kimwnasptd 
The most important ones (Seldon and kserve) already work with the predecessor of PSS. I am running it with similar restrictions (and istio-cni) for over one year.  Kuberay does so as well and they will go even one step further as detailed in https://github.com/ray-project/kuberay/pull/750. So the restricted PSS still allows to run as a specific non-root user but the kuberay developers want to become even better than that.

kimwnasptd:
After some discussion with @juliusvonkohout we decided to use this PR towards the overarching goal of better defining the implementation details (or technical requirements) that we would expect from the contrib components.

We've had context around this in other places, like https://github.com/kubeflow/manifests/pull/2286#issuecomment-1273538637.

The first step will be to break down the requirements to 2 categories:
* Structural requirements
* Technical requirements

The first category is regarding what we currently have. Requirements regarding the file structure of the addons. Things like having a README, exposing upgrade instructions and examples. 

Then the next one will be a list of more technical requirements that we will expect these addons to follow

kimwnasptd:
Some aspects that I believe we'll need to re-evaluate are:
* Defining the least subset of **required** criteria for accepting new components
* Re-think the deprecation process, since we are adding more requirements

Also we could think about a model where we explicitly "rank"/document how much an addon aligns with the requirements. So that users looking at different addons can know the state and be able to decide which ones they want to add in their platform

cc @annajung

kimwnasptd:
Adding some items we discussed with @juliusvonkohout and @kromanow94 live as functional requirements:
* Running with baseline PSS
* [stretch] Namespace isolation
* [stretch] Running with strict PSS
* [stretch] Running with Istio sidecars

juliusvonkohout:
> Adding some items we discussed with @juliusvonkohout and @kromanow94 live as functional requirements:
> 
>     * Running with baseline PSS
> 
>     * [stretch] Namespace isolation
> 
>     * [stretch] Running with strict PSS
> 
>     * [stretch] Running with Istio sidecars

i have added the necessary changes.

kimwnasptd:
/lgtm

Thanks @juliusvonkohout!

PR #2330, opened 2022-11-24
common: Upgate cert-manager manifests to v1.10.1 

apo-ger:
Changes:
- Upgrade cert-manager manifests to latest version (v1.10.1) to support K8s-v1.25
- Remove `crd-preserve-unknown-fields` patch as `spec.preserveUnknownFields` field is deprecated in [v1 CRR API](https://kubernetes.io/docs/reference/kubernetes-api/extend-resources/custom-resource-definition-v1/#CustomResourceDefinitionSpec)
- Add Cert Manager version reference in main `README` file
- Trigger the test workflow of each component, that uses Cert Manager, when Cert Manager manifests are changed


kimwnasptd:
Thanks for the PR @apo-ger! We have a couple of components that depend on Cert Manager, with Katib and Pipelines coming at the top of my head. https://github.com/kubeflow/manifests/tree/master/.github/workflows

Let's update their GH Action workflows and trigger the test workflow of each component, that uses Cert Manager, when Cert Manager manifests are changed.

kimwnasptd:
We have a helper script for installing Cert Manager, so you can check which workflows are using that script to know which of them to update https://github.com/kubeflow/manifests/blob/master/tests/gh-actions/install_cert_manager.sh

apo-ger:
@kimwnasptd 
>Let's update their GH Action workflows and trigger the test workflow of each component, that uses Cert Manager, when Cert Manager manifests are changed.

On it! Although I expect the workflows to fail, as they are currently installing `Istio v1.14`: https://github.com/kubeflow/manifests/blob/master/tests/gh-actions/install_istio.sh

The Istio helper script is updated in https://github.com/kubeflow/manifests/pull/2327

apo-ger:
Retesting GH Action workflows now that we merged https://github.com/kubeflow/manifests/pull/2327 and https://github.com/kubeflow/manifests/pull/2331

I excluded the `Katib` and `Kserve` workflows, similar to what we did in https://github.com/kubeflow/manifests/pull/2331#issuecomment-1326464742

PR #2273, opened 2022-08-22
Update kubeflow/katib manifests from v0.14.0

annajung:
Signed-off-by: Anna Jung (VMware) <antheaj@vmware.com>

**Description of your changes:**
- Sync Katib 0.14.0 ([ref](https://github.com/kubeflow/manifests/issues/2200#issuecomment-1220490112))

cc @kubeflow/release-team

PR #2331, opened 2022-11-24
Refactor kind configuration file for testing with K8s-v1.25

apo-ger:
Follow-up from (https://github.com/kubeflow/manifests/pull/2324)

cc. @kimwnasptd 
Signed-off-by: Apostolos Gerakaris <apoger@arrikto.com>


kimwnasptd:
@apo-ger with the same spirit of running workflows that depend on a common component when the manifests of the common component changes https://github.com/kubeflow/manifests/pull/2330#issuecomment-1326271985; lets slightly change the workflows to:
1. Have one file for the KinD configuration. So we'll rename https://github.com/kubeflow/manifests/blob/master/tests/gh-actions/kind-cluster-1.25.yaml to just `kind-cluster.yaml`
2. Update the GH Action workflows to use that file always, and not a version specific one
3. Trigger the workflows when that `kind-cluster.yaml` file changes

This way we have only one file for the configuration and run everything when we change the K8s version

apo-ger:
@kimwnasptd Updated the PR with the proprosed changes

kimwnasptd:
@apo-ger let's exclude the update on Katib and KServe here.

For Katib though we had bumped into this in the past with @annajung but hadn't managed to get to the bottom of it https://github.com/kubeflow/manifests/pull/2273#issuecomment-1233092205. We'll need to take a detailed look again.

Let's unblock this PR for now and look into these afterwards

apo-ger:
@kimwnasptd Done!

PR #2357, opened 2023-01-18
Harden the istio 1.16 configuration

juliusvonkohout:
The istio nodeport (loadbalancer)instead of a proper ClusterIP seems dangerous to me https://oteemo.com/think-nodeport-kubernetes/
Fixes https://github.com/kubeflow/manifests/issues/2285
@kimwnasptd

i had to create this from https://github.com/kubeflow/manifests/pull/2296 since the master branch switched to 1.16.

**Checklist:**
- [ ] Unit tests pass:
    2. `make test`

### DIFF
--- a/common/dex/base/service.yaml
+++ b/common/dex/base/service.yaml
@@ -3,12 +3,11 @@ kind: Service
 metadata:
   name: dex
 spec:
-  type: NodePort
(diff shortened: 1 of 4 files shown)


kimwnasptd:
Thanks!

/lgtm
/approve

PR #2355, opened 2023-01-12
Kserve documents and Tests - issue# 2311

andyi2it:
**Which issue is resolved by this Pull Request:**
Resolves #2311 for kserve

**Description of your changes:**
* Adds Readme and Upgrade docs for kserve.
* Adds e2e for kserve
* Updates kserve manifests to v0.10

**Checklist:**
- [ ] Unit tests pass:
  **Make sure you have installed kustomize == 3.2.1**
    1. `make generate-changed-only`
    2. `make test`

PR #2369, opened 2023-01-30
Update kserve manifests from v0.10.0

DomFleischmann:
Update kserve/kserve manifests from v0.10.0

@yuzisun @kimwnasptd


kimwnasptd:
Looks like the ISVC never becomes Ready. I tried to re-run the steps of the action locally and found that the ISVC has the following event:
```
Events:
  Type     Reason         Age                  From                Message
  ----     ------         ----                 ----                -------
  Warning  InternalError  21s (x15 over 103s)  v1beta1Controllers  no runtime found to support predictor with model type: {sklearn <nil>}
```

So most probably this is something to do with the ISVC we apply and how we defined the runtime. The current ISVC we apply is  https://github.com/kubeflow/manifests/blob/f4810ed177d342afb258b9f983ebb1ffaac4f3cb/tests/gh-actions/kf-objects/kserve_test.yaml

And was a slightly modified copy from https://github.com/kserve/kserve/blob/release-0.10/docs/samples/v1beta1/sklearn/v1/sklearn.yaml

kimwnasptd:
@yuzisun do you have any pointers? I'm looking around at the KServe docs to verify that most probably we need to update the ISVC yaml we apply

kimwnasptd:
Looking around at the KServe issues I found an identical one, but it's solution is not good news https://github.com/kserve/kserve/issues/2643

Specifically they mention that the problem was with Istio version `1.15.x` is not compatible with the latest KServe. In Kubeflow we targeted for 1.16 https://github.com/kubeflow/manifests#kubeflow-components-versions

@yuzisun could you give us some input? Is KServe not compatible with the Istio 1.16 version?

DomFleischmann:
I'm closing this PR as it is a duplicate of #2355

PR #2370, opened 2023-01-31
Update kubeflow/kubeflow manifests from a74094c3a

kimwnasptd:
Update the manifests for Notebooks WG that live in `kubeflow/kubeflow` for the first RC.

Due to some permission issues I've made the PR based on the commit https://github.com/kubeflow/kubeflow/commit/a74094c3ae6c63308bc908d0e816100214977462, which is what we'll use for cutting the `v1.7.0-rc.0` tag in kubeflow/kubeflow

cc @DomFleischmann @chensun


juliusvonkohout:
@kimwnasptd does it include the renaming to workbenches?

kimwnasptd:
@juliusvonkohout we didn't make any progress on that front in the 1.7 cycle. We'll be prioritizing this for 1.8 and moving it forward in 2 gears, a UX part as you described (UI changes and docs) and a parallel discussion for technical changes. We'll open a dedicated issue within the next weeks, once we get a breather from 1.7

kimwnasptd:
@elikatsis I propose that we merge this PR now, and we cut the RC from that commit in kubeflow/kubeflow afterwards.

The only downside is that the README shows a commit hash, instead of a nice `v1.7.0-rc.0`, but it's better than delaying the whole release

PR #2380, opened 2023-02-07
Uncomment profiles manifests in example dir

DomFleischmann:
It seems like the profiles manifests where commented out accidentally when updating the seldon manifests. This was causing the one-line installation to fail. By uncommenting this line the installation should work again.


kimwnasptd:
Nice catch! 

/lgtm
/approve

PR #2371, opened 2023-01-31
Add Julius von Kohout as reviewer for /contrib

kimwnasptd:
@juliusvonkohout has been helping on the Manifests WG for a very long time and has done an amazing job on reviews on the `/contrib` folder. For this reason I'd like to promote Julius to be a reviewer for the `/contrib` dir.

I'm also creating this PR after looking at https://github.com/kubeflow/manifests/pull/2337. I believe that we should have a more granular process of adding someone as an approver in the whole repo and taking iterative steps. The first one is to acknowledge Julius' help on reviews and ensure the OWNERS file represent that. Then move progressively based on the (great!) contributions that will be getting merged.

At the same time I want to ensure that we have a common process for next steps for everyone, and I think https://github.com/kubeflow/community/pull/591 is the best place for this.

Contributions:
* KServe (review): https://github.com/kubeflow/manifests/pull/2355
* BentoML (review): https://github.com/kubeflow/manifests/pull/2350
* KubeRay: https://docs.google.com/document/d/1SP-COL6O-ETMQny7zXbrbEL4zZBG3REErWjfcvFZVEw/edit?resourcekey=0-IaUGVemd2XX6MGTgifDbpg

cc @kubeflow/wg-manifests-leads

PR #2383, opened 2023-02-07
WIP: Kuberay integration

juliusvonkohout:
Resolves https://github.com/kubeflow/kubeflow/issues/6680

@kevin85421

**Checklist:**
- [ ] Unit tests pass:
    2. `make test`


kevin85421:
I will update the progress in this comment. 

* `Makefile`: Transform the latest release KubeRay Helm chart (i.e. 0.4.0 for now) to Kustomize. **[Status: ready to review]**
  * By running the command `make`, the Helm chart will be transformed and stored as a YAML file `kuberay-operator/base/resources.yaml`.

* `README.md`: **[Status: TODO]**
  *  [x] Install the manifest in one command.
  *  [ ] Introduction of Ray
  *  [ ] Architecture
  *  [ ] Example
* Test: **[Status: TODO]**
* `UPGRADE.md`: **[Status: TODO]**
* RayCluster example YAML: **[Status: TODO]**

juliusvonkohout:
Does 0.4.0 have the securitycontext set everywhere?

kevin85421:
> Does 0.4.0 have the securitycontext set everywhere?

Yes, the PR https://github.com/ray-project/kuberay/pull/752 also includes in the release 0.4.0. See [here](https://github.com/ray-project/kuberay/commits/release-0.4) for more details.

juliusvonkohout:
> > Does 0.4.0 have the securitycontext set everywhere?
> 
> Yes, the PR [ray-project/kuberay#752](https://github.com/ray-project/kuberay/pull/752) also includes in the release 0.4.0. See [here](https://github.com/ray-project/kuberay/commits/release-0.4) for more details.

Are they also enforced everywhere?

juliusvonkohout:
i do not see the securitycontext in raycluster_example.yaml or resources.yaml. Shouldnt there be

```
          securityContext:
            allowPrivilegeEscalation: false
            capabilities:
              drop: ["ALL"]
            runAsNonRoot: true
            seccompProfile:
              type: RuntimeDefault
```

everywhere as shown in https://github.com/ray-project/kuberay/pull/750/files ?

kevin85421:
> i do not see the securitycontext in raycluster_example.yaml or resources.yaml. Shouldnt there be
> 
> ```
>           securityContext:
>             allowPrivilegeEscalation: false
>             capabilities:
>               drop: ["ALL"]
>             runAsNonRoot: true
>             seccompProfile:
>               type: RuntimeDefault
> ```
> 
> everywhere as shown in https://github.com/ray-project/kuberay/pull/750/files ?

Good catch. Will update it today.

contrib/kuberay/raycluster_example.yaml
```diff
@@ -0,0 +1,98 @@
...
+    rayStartParams:
+      dashboard-host: '0.0.0.0'
+      block: 'true'
+    # pod template
+    template:
+      metadata:
+        # Custom labels. NOTE: To avoid conflicts with KubeRay operator, do not define custom labels start with `raycluster`.
+        # Refer to https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/
+        labels: {}
+      spec:
```
kevin85421, line 21: TODO: SecurityContext

kevin85421:
This PR is ready for review cc @juliusvonkohout @richardsliu. Could you approve the CI workflow? Thanks!

cc some KubeRay folks @architkulkarni @DmitriGekhtman @gvspraveen @Yicheng-Lu-llll @mjconnor @Jeffwan

juliusvonkohout:
> > i do not see the securitycontext in raycluster_example.yaml or resources.yaml. Shouldnt there be
> > ```
> >           securityContext:
> >             allowPrivilegeEscalation: false
> >             capabilities:
> >               drop: ["ALL"]
> >             runAsNonRoot: true
> >             seccompProfile:
> >               type: RuntimeDefault
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > everywhere as shown in https://github.com/ray-project/kuberay/pull/750/files ?
> 
> Good catch. Will update it today.

What about the initcontainers?

kevin85421:
> What about the initcontainer securitycontext?

Added. Btw, what's the reason to add `securityContext` here? The namespaces created by Kubeflow (`example/`) do not apply PSS. Why do we set the `securityContext` by default rather than making users configure that?

> Is there a reason for the logfile in the repository?

Removed.

> We have to check https://github.com/kubeflow/manifests/blob/master/proposals/20220926-contrib-component-guidelines.md#component-requirements for conformance

Maybe you can do that part. You are more familiar with Kubeflow.

> the test failed

Does KubeRay use free plan GitHub Actions? If so, the CPU may be insufficient.


# Others

* I fetched your master branch and rebase it. Everything goes wrong, so I need to force push to make the branch back. Would you mind adding your comments again?

* I prefer to use Fig. 1 compared with Fig. 2. The example in README primarily focuses on Jupyter Notebook / Kubeflow Central Dashboard / KubeRay operator. Hence, Fig. 2 includes a lot of information that is unrelated to the example. This will make users confused.
  <details>
  <summary>Fig. 1</summary>
  architecture
  </details>
  <details>
  <summary>Fig. 2</summary>
  architecture
  </details>

Any thoughts? Thanks!

juliusvonkohout:
"Added. Btw, what's the reason to add securityContext here? The namespaces created by Kubeflow (example/) do not apply PSS. Why do we set the securityContext by default rather than making users configure that?" Setting and enforcing are two different things. We want to at least set the securitycontext

juliusvonkohout:
Figure 1 will be wrong quite soon. We rename everything to workbenches. Jupyter notebooks is also wrong it is a Notebook inside a Jupyterlab like a car and gasoline, file and server. Furthermore we also want to call it from KFP directly. We can include both images, but then please use the SVG version and rename to workbench/Jupyterlab, get rid of the notebook term and add KFP.

juliusvonkohout:
And thanks for your amazing effort so far

kevin85421:
> Setting and enforcing are two different things. We want to at least set the securitycontext

Got it. Although the configuration has no relationship with this integration, it is nice to have.

> Figure 1 will be wrong quite soon. We rename everything to workbenches. 

I will add some sentences to explain it, but I prefer not to change the name of "Kubeflow Central Dashboard" in Figure 1. The example is based on release 1.6, and if we introduce new terms from future releases, users may feel confused. You can open a new PR to update it if you want the example to support the future release.

> Jupyter notebooks is also wrong it is a Notebook inside a Jupyterlab like a car and gasoline, file and server. 

Fixed.

> Furthermore we also want to call it from KFP directly. We can include both images, but then please use the SVG version and rename to workbench/Jupyterlab, get rid of the notebook term and add KFP.

KFP (Kubeflow pipeline) seems not to be a necessary component for users to understand this example. It may make users feel confused. Again, you can open a new PR later to add a section to describe how to launch a JupyterLab instance via KFP.  

> please use the SVG version

Will do it.

juliusvonkohout:
KFP or Kubeflow pipelines is the main and most used part of Kubeflow. Way more important than the workbenches and Jupyterlab. You can just add sommething parallel to Jupyterlab that is called KFP in the diagram.

So 
1. From Jupyterlab to ray cluster "Create Ray cluster and connect to it via Ray client"
2. From KFP to ray cluster "Create Ray cluster and connect to it via Ray client"
3. KFP can just be below the Jupyterlab rectangle
4. From centraldashboard to KFP "start pipeline"

By the way the to be renamed notebook controller also does not matter in any way for normal Kubeflow users only KFP and Jupyterlab.

"Got it. Although the configuration has no relationship with this integration, it is nice to have." We will only allow secure by default stuff. Security is not an add-on, but mandatory functionality. 

 "There must be an OWNERS file with at least 2 users" @juliusvonkohout  and @kimwnasptd  so that is fine

The component must work with the latest version of Kubeflow, and its dependencies. That is also fine, because we do not have the UI integration yet. But this is something for the next PR.

But i am wondering what the istio sidecars are doing to ray. Maybe we need to have a call to check that on your installation. Do you see Istio initcontainers and sidecars (proxy) for ray head, worker etc?

juliusvonkohout:
By the way who approved the CICD workflow last time?

juliusvonkohout:
@kimwnasptd @DomFleischmann  can you check and approve the workflows?

kevin85421:
> Add KFP to the diagram

Done

> Do you see Istio initcontainers and sidecars (proxy) for ray head, worker etc?

No, maybe I can try to put the Istio integration as an item for next quarter.

juliusvonkohout:
> No, maybe I can try to put the Istio integration as an item for next quarter.

This might be problematic, because in a full Kubeflow they are injected by default.
So why are they not injected in your installation?

https://istio.io/latest/docs/setup/additional-setup/sidecar-injection/ is the documentation

since profile namespaces have the istio-injection=enabled label they will be added by default.

Or did you annotate the ray pods with `sidecar.istio.io/inject="false"` ?

Or are you not in a Kubeflow profile namespace?

juliusvonkohout:
`NAMESPACE=default` defintely looks wrong. We need to test in a namespace where also your jupyterlabs from Kubeflow are running

juliusvonkohout:
@kimwnasptd  please approve

juliusvonkohout:
Istio support for ray is tracked here https://github.com/ray-project/kuberay/issues/1005

juliusvonkohout:
@kevin85421 after https://github.com/kubeflow/manifests/pull/2371 i can approve

kimwnasptd:
Thanks for the great work @juliusvonkohout! Awesome to see Ray start being part of Kubeflow

/lgtm
/approve

PR #2395, opened 2023-02-16
Update kubeflow/kubeflow manifests from v1.7.0-rc.1

DomFleischmann:
Update kubeflow/kubeflow manifests from v1.7.0-rc.1


kimwnasptd:
Looks like the action that build the TWA has not yet completed, to push the image with the expected tag
https://github.com/kubeflow/kubeflow/actions/runs/4194199839/jobs/7272051193

Will re-trigger the job once it succeeds and the image is pushed

Issue #2388, opened 2023-02-10
Kubeflow manifests and kustomize 5


kimwnasptd:
Kustomize 5 has just been released, with the fix we wanted for ordering https://github.com/kubernetes-sigs/kustomize/issues/3794 https://github.com/kubeflow/manifests/issues/1797.

Will use this issue to expose the current status of installing the Kubeflow manifests with Kustomize 5.

/cc @DomFleischmann @annajung @juliusvonkohout

kimwnasptd:
The situation with Kustomize 5 is that:
1. The manifests can be built, and we can avoid the ordering issue with the fix from https://github.com/kubernetes-sigs/kustomize/issues/3794
2. We can't use the same `kustomization.yaml` for both kustomize 5 and earlier versions, since the [`sortOptions`](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/sortoptions/) field is supported only by Kustomize 5
3. Because of a regression in Kustomize 5 we'll need to slightly adjust some of our commands to work around it https://github.com/kubernetes-sigs/kustomize/issues/5039
4. While the warnings we see are not blockers, for now, we'll need to some manifests work to mitigate them

I'll expose more technical details in the following messages, but I propose that we do the switch and just go with Kustomize 5

kimwnasptd:
Specifically, if we try to build the [`example`](https://github.com/kubeflow/manifests/blob/master/example/kustomization.yaml) we'll get the following warnings (multiple times):
```
# Warning: 'patchesStrategicMerge' is deprecated. Please use 'patches' instead. Run 'kustomize edit fix' to update your Kustomization automatically.
# Warning: 'vars' is deprecated. Please use 'replacements' instead. [EXPERIMENTAL] Run 'kustomize edit fix' to update your Kustomization automatically.
# Warning: 'bases' is deprecated. Please use 'resources' instead. Run 'kustomize edit fix' to update your Kustomization automatically.
# Warning: 'patchesJson6902' is deprecated. Please use 'patches' instead. Run 'kustomize edit fix' to update your Kustomization automatically.
```

We could try to do `kustomize edit fix`, but there's a good chance we'll need to do manual changes as well

kimwnasptd:
Next up, if we'd try to apply these manifests we'll see that the generated manifests look like this:
```
2023/02/22 15:47:37 well-defined vars that were never replaced: kfp-app-name,kfp-app-version
apiVersion: v1
kind: Namespace
metadata:
[...]

PR #2399, opened 2023-02-23
Update instructions for Kustomize 5

kimwnasptd:
Refs https://github.com/kubeflow/manifests/issues/2388

This PR updates the instructions and the example kustomization to work with Kustomize 5. Note that after this PR is merged users won't be able to use earlier versions of kustomize, since we need the `sortOptions` feature of Kustomize 5 https://github.com/kubeflow/manifests/issues/2388#issuecomment-1440073723. This holds for the one-liner install. The manual instructions are not affected and people can use earlier versions of kustomize there.

Next work items, although not critical for the KF 1.7 release, are to update our manifests so that we don't see warnings from Kustomize https://github.com/kubeflow/manifests/issues/2388#issuecomment-1440044713

cc @DomFleischmann @annajung @juliusvonkohout @jbottum 

### DIFF
--- a/README.md
+++ b/README.md
@@ -440,3 +440,5 @@ For security reasons, we don't want to use the default password for the default 
 
 - **Q:** What versions of Istio, Knative, Cert-Manager, Argo, ... are compatible with Kubeflow? \
   **A:** Please refer to each individual component's documentation for a dependency compatibility range. For Istio, Knative, Dex, Cert-Manager and OIDC-AuthService, the versions in `common` are the ones we have validated.
+- **Q:** Can I use earlier version of Kustomize with Kubeflow manifests?
+  **A:** The manual installation instructions work with Kustomize 3.2. To use the one-liner installation you'll need to comment out the `sortOptions` section in the `example/kustomization.yaml`.
--- a/README.md
+++ b/README.md
@@ -84,14 +84,16 @@ The `example` directory contains an example kustomization for the single command
 ### Prerequisites
 
 - `Kubernetes` (up to `1.25`) with a default [StorageClass](https://kubernetes.io/docs/concepts/storage/storage-classes/)
-- `kustomize` (version [`3.2.0`](https://github.com/kubernetes-sigs/kustomize/releases/tag/v3.2.0) or [`4.5.7`](https://github.com/kubernetes-sigs/kustomize/releases/tag/kustomize%2Fv4.5.7))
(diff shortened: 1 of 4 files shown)


annajung:
Thanks Kimonas! I couldn't recreate the race condition described in the https://github.com/kubeflow/manifests/issues/1797 with kustomize 5, but can confirm that with sortOptions added, it does create `MutatingWebhookConfiguration` before the necessary resources.

Testing with `kustomize build example`, only difference between kustomize5 with sortOptions enabled and kustomize3 was the order of when PVCs were created. It looks like kustomize3 places PVC at the end right before the ValidatingWebhookConfiguration, so kustomize5 order is preferrable IMO

lgtm (not adding the actual label to allow other folks to review)

there is still ongoing discussion on if it should be included in 1.7 release in https://github.com/kubeflow/manifests/issues/2388 but should not be a blocker for this PR.

README.md
```diff
@@ -438,7 +440,3 @@ For security reasons, we don't want to use the default password for the default
 
 - **Q:** What versions of Istio, Knative, Cert-Manager, Argo, ... are compatible with Kubeflow? \
   **A:** Please refer to each individual component's documentation for a dependency compatibility range. For Istio, Knative, Dex, Cert-Manager and OIDC-AuthService, the versions in `common` are the ones we have validated.
```
DomFleischmann, line 442: Here is a suggestion for the Kustomize 3.2 QA section.

```suggestion
  **A:** Please refer to each individual component's documentation for a dependency compatibility range. For Istio, Knative, Dex, Cert-Manager and OIDC-AuthService, the versions in `common` are the ones we have validated.
  
- **Q: ** How can I still use Kustomize version 3.2?
  **A: ** The only thing incompatible with Kustomize 3.2 is the `sortOptions` section in the `example/kustomization.yaml` file. I you want to use Kustomize 3.2 just comment out the entire section (`#` character in front of each line) and execute the one-liner. 
```

kimwnasptd: @DomFleischmann I pushed a commit with a slightly refactored section for the earlier version of Kustomize. PTAL

kimwnasptd:
Adding a summary here as well from the last community (Tue 28 Feb) and release meeting (Mon 6 March). We decided to move forward with using Kustomize 5 as the default installation tool.

The rationale is that:
1. With feedback until now we have heard from distros that they managed to use Kustomize 5 without any issues
2. There are no (currently known) deprecations that affect our use of kustomize
3. We'd like to move to a newer version of Kustomize, which hasn't been updated for a long time
4. We have documented a clear path on how people can use earlier versions of Kustomize, if they need to

So at this point we have confidence from both Manifests WG and feedback from distros that Kustomize 5 does not have any issues we haven't yet documented. Plus if adopters are not ready for the version switch they can still use an earlier version of kustomize. So let's move forward with this PR and also cherry-pick it to the KF 1.7 release branch

PR #2403, opened 2023-03-03
Adding kubeflow-groups headers for kubeflow components

axel7083:
Following those PR:
- https://github.com/kubeflow/kubeflow/pull/7022
- https://github.com/kubeflow/pipelines/pull/8918

The [oidc-authservice](https://github.com/arrikto/oidc-authservice) support **GROUPS_HEADER**, the istio configuration need to be adjusted to allow kubeflow components to consume it. Adding `kubeflow-groups` allows the components to access the header.


juliusvonkohout:
We are moving away from oidc-atuhservice and this PR is stale

PR #2428, opened 2023-03-28
Update kubeflow/kubeflow manifests from v1.7.0

DomFleischmann:
Update kubeflow/kubeflow manifests for v1.7.0


difince:
@difince - add myself to get notified when the PR is merged

PR #2435, opened 2023-04-05
Replacing autoscaling/v2beta2 by autoscaling/v2

axel7083:
## Description

apiVersion `autoscaling/v2beta2` for HorizontalPodAutoscaler is deprecated and need to be replaced by apiVersion `autoscaling/v2`.

The change has already been made in multiple files but was still v2beta2 in `common/knative/knative-serving/base/upstream/serving-core.yaml` .


annajung:
Thanks for catching that!
/lgtm

ruckc:
Ran into this issue this morning, glad its an easy solution.

annajung:
cc @kimwnasptd for review

PR #2466, opened 2023-06-01
Add end-to-end testing proposal

annajung:
Move the content of [end-to-end manifest testing proposal doc](https://docs.google.com/document/d/1ipY1RxnAuJCFHJAd9xc7KfFaiMZQTQtk92u_HY5eBl4) to github

During May 18th meeting, it was agreed to move the content of the proposal into the Manifest repo and collect the proposal in one place.

The date in the file name (20230323) represents the day the proposal was approved.

cc @DomFleischmann @kimwnasptd

### DIFF
--- a/proposals/20230323-end-to-end-testing.md
+++ b/proposals/20230323-end-to-end-testing.md
@@ -1,8 +1,7 @@
 # End-to-end Testing
 
-**Authors**: Dominik Fleischmann ([@domFleischmann](https://github.com/domFleischmann)),
-Kimonas Sotirchos ([@kimwnasptd](https://github.com/kimwnasptd)),
-and Anna Jung ([@annajung](https://github.com/annajung))
+**Authors**: Dominik Fleischmann ([@domFleischmann](https://github.com/domFleischmann)), Kimonas
+Sotirchos ([@kimwnasptd](https://github.com/kimwnasptd)), and Anna Jung ([@annajung](https://github.com/annajung))
 
 ## Background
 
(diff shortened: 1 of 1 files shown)


proposals/20230323-end-to-end-testing.md
```diff
@@ -0,0 +1,372 @@
...
+Enable the e2e testing for the Manifest repo and leverage it to shorten the manifest testing phase of the Kubeflow
+release cycle and to increase quality of the Kubeflow release by ensuring Kubeflow components and dependencies work
+correctly together.
+
+## Proposal
+
+After some initial conversations, it has been agreed to create integration tests based on GitHub Actions, which will
+spawn an EC2 instance with enough resources to deploy the complete Kubeflow solution and run some end-to-end testing.
+
+### Alternative solutions considered
```
kimwnasptd, line 32: Can we move this section to the end of the doc? IMO it's more clear if the proposal is focused only on what we propose, and then at the end expose some info on the alternatives. So the reader's focus is always on the selected approach while going through the doc

proposals/20230323-end-to-end-testing.md
```diff
@@ -0,0 +1,372 @@
...
+- [Create an EC2 instance](#create-an-ec2-instance)
+- [Install a Kubernetes on the instance](#install-a-kubernetes-on-the-instance)
+- [Deploy Kubeflow](#deploy-kubeflow)
+- [Run tests](#run-tests)
+- [Log and report errors](#log-and-report-errors)
+- [Clean up](#clean-up)
+
+### Create an EC2 instance
+
+Access the AWS credentials (stored as GH Secrets) and create an EC2 instance
```
kimwnasptd, line 59: Could expose some more info here regarding what are the type of info that we'll need to be storing as GH Secrets? Like:
1. The AWS account name
2. The key used for authenticating 
3. etc...

proposals/20230323-end-to-end-testing.md
```diff
@@ -0,0 +1,372 @@
...
+While juju provides more capability, the proposal is to use the tool as config management and a medium to deploy and
+connect with EC2 instances.
+
+**Note**: Using GitHub Secrets to store AWS credentials will not allow any forked repositories to access the secrets.
+
+### Install a Kubernetes on the Instance
+
+Install Kubernetes on the EC2 instance where Kubeflow will be deployed and tested
+
+To install Kubernetes, we explored two options and propose to use **KinD**
```
kimwnasptd, line 83: Following my rationale abovee, let's only keep this section for how to install KinD and put MicroK8s in the Alternatives Considered section at the end

kimwnasptd (commented):
Thanks @annajung! Left some small comments, mostly regarding the structure. Rest LGTM

proposals/20230323-end-to-end-testing.md
annajung: Moved to the last section

proposals/20230323-end-to-end-testing.md
annajung: Moved to the last section

proposals/20230323-end-to-end-testing.md
annajung: I created another step in the beginning called "Create credentials required by the AWS" to address what variables need to be created. PTAL

kimwnasptd: Looks great, thanks!

kimwnasptd:
Also the more I think of defining the test as an ipynb the more I start to like it. This way we can also test KFP credentials from inside the cluster and will reduce the setup logic, since we won't need to be port-forwarding in-cluster services.

Will also work in parallel with injecting the required AWS credentials in the repo.

/lgtm
/approve

PR #2473, opened 2023-07-03
To enable Katib to work on ARM64

hwang9u:
The existing docker images with `v0.15.0-rc.0` tag do not support linux/arm64. So I encountered errors `Back-off restarting failed container` and was able to resolve them by making the change to `latest`, resulting in proper functioning.


annajung:
Hi @hwang9u, to update the Katib manifest, the PR needs to be made against the [Katib repo](https://github.com/kubeflow/katib) as direct changes to any component not managed by the manifest WG would be rewritten when the manifest groups trigger a sync

Please move the PR to the correct repo

/close

Issue #2469, opened 2023-06-13
Move away from AuthService


kimwnasptd:
My current understanding is that the `gcr.io/arrikto` oci registry will be taken down at some point*.

I can think of the following moving pieces for this process:
1. Moving away from this component for the KF 1.8 release
2. Understand whether we need to patch previous versions of Kubeflow
3. Existing installations that are using AuthService

From my current understanding there won't be any migration of the image from the side of the [AuthService project](https://github.com/arrikto/oidc-authservice).

The first information that we need is for how long will he `gcr.io/arrikto` registry be up. After we have this we can have more concrete next steps. Hopefully I'll have some more news on this relatively soon*.

The way that I think this play out though is that we'll need to mirror images in the Kubeflow project and do patches on some* KF installations so that the mirrored image will be used instead. And of course moving away in KF 1.8 either way.

cc @kubeflow/wg-manifests-leads @DnPlas @annajung @jbottum

jbottum:
Is there a manifest WG meeting today ?

DnPlas:
Thanks @kimwnasptd, I have some questions.
1. Moving away from the current Authservice implementation in the upcoming 1.8 release would mean that:
* We may need to provide steps for upgrading Kubeflow 1.7 to 1.8.
* We already have a solution for replacing Authservice, do you mind sharing details about it? What would be the impact on Kubeflow components in terms of changes in manifests and/or code?
* Would this change impact the current release timeline?

2. Mirroring the existing image and patching some KF installations - For this I'd add that only the image will be kept, and no source code. Though I agree on this one, I think we have to do it regardless of how long the image registry is up.

kimwnasptd:
@DnPlas 

> Moving away from the current Authservice implementation in the upcoming 1.8 release would mean that:

1. Yes we'll provide upgrade instructions on how to migrate to `oauth2-proxy` for users that come from a 1.6, and have AuthService in their cluster
    * I'd expect the `oauth2-proxy` and AuthService manifests to be completely separated, so it'll be a matter of removing AuthService manifests and applying `oauth2-proxy` manifests
2. The 1.8 manifests will be using a completely different component `oauth2-proxy`
    * There shouldn't be any differences to other components
3. I wouldn't expect it to affect the release timeline

[...]

PR #2480, opened 2023-07-13
Update AuthService image for KF 1.7

kimwnasptd:
Part of #2469 

Mirrored the `gcr.io/arrikto/kubeflow/oidc-authservice:e236439` over to `docker.io/kubeflowmanifestswg/oidc-authservice:e236439`

### DIFF
--- a/common/oidc-authservice/base/statefulset.yaml
+++ b/common/oidc-authservice/base/statefulset.yaml
@@ -18,7 +18,7 @@ spec:
       serviceAccountName: authservice
       containers:
       - name: authservice
-        image: gcr.io/arrikto/kubeflow/oidc-authservice:e236439
+        image: docker.io/kubeflowmanifestswg/oidc-authservice:e236439
         imagePullPolicy: Always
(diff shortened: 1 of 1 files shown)


common/oidc-authservice/base/statefulset.yaml
```diff
@@ -18,7 +18,7 @@ spec:
       serviceAccountName: authservice
       containers:
       - name: authservice
-        image: gcr.io/arrikto/kubeflow/oidc-authservice:e236439
+        image: docker.io/kubeflowmanifestswg/oidc-authservice:e236439
```
annajung, line 21: does it need to be changed in /base/kustomization.yaml file?

annajung:
@kimwnasptd we should also merge this in before the 1.8 RC

common/oidc-authservice/base/statefulset.yaml
kimwnasptd: Good catch! Updated

Issue #2450, opened 2023-04-25
[kubeflow 1.8] Dependencies versions


DnPlas:
This issue will be used for tracking, discussing and sharing the dependencies versions.

> Please note these versions are subject to agreement of all WG and the release team, and may suffer changes in the following days before the release cycle begins on May 8th, 2023.

| Component    | Version used in 1.7 | Version used in 1.8 |
|--------------|------------------|-------------|
| [Knative](https://knative.dev/docs/reference/relnotes/)      |     1.8.1             |     1.10        |
| [Istio](https://istio.io/latest/news/releases/)        |         1.16         |     1.17       |
| [Kubernetes](https://kubernetes.io/releases/)   |         1.25/1.24         |      1.25/1.26 ~1.27 (stretch goal)~      |
| [Cert-Manager](https://cert-manager.io/docs/installation/supported-releases/) |        1.10.1          |   1.12*          |
| [Kustomize](https://github.com/kubernetes-sigs/kustomize/releases)    |         3.2 or 5.0         |       5.0.1      |
| [Dex](https://github.com/dexidp/dex/releases)          |        2.31          |     2.36        |
| [Argo](https://github.com/argoproj/argo-workflows/releases)         |           3.3.8       |      3.3.10       |
| Tekton       |      x            |    0.47 or 0.48      |
| Oidc         |          e236439       |       e236439      |

* This version EOL is end of August 2023 and it's tied to k8s 1.26. For more info, [here](https://cert-manager.io/docs/installation/supported-releases/#upcoming-releases)

annajung:
Based on the Kubeflow [1.8 release timeline](https://github.com/kubeflow/community/pull/621/files), 1.8 is planned to be released in **early Oct**.

Kubernetes
- Kubernetes 1.25 is [EOL 2023-10-28](https://kubernetes.io/releases/#release-v1-25)
- Kubernetes 1.26 is [EOL 2024-02-28](https://kubernetes.io/releases/#release-v1-26)
- Kubernetes 1.27 is [EOL 2024-06-28](https://kubernetes.io/releases/#release-v1-27)
- Latest version of Kubernetes will be 1.28, planned to be [released on August 15th, 2023](https://www.kubernetes.dev/resources/release/#summary), which is during Kubeflow 1.8 manifest testing phase

GKE - pipeline CI/CD
- [GKE release schedule](https://cloud.google.com/kubernetes-engine/docs/release-schedule#schedule-for-release-channels)
  - Stable version
[...]

PR #2482, opened 2023-07-13
Add Kubernetes version 1.26 and set it to default for testing

annajung:
Part of https://github.com/kubeflow/manifests/issues/2450

- Move default KinD 1.25 to `kind-cluster-1-25.yaml`
- Add support for Kubernetes 1.26 and set it to default for testing
- Update the requirements.txt to run e2e with katib 0.15 and kfp 1.8.22

### DIFF
--- a/.github/workflows/pipeline_kind_test.yaml
+++ b/.github/workflows/pipeline_kind_test.yaml
@@ -37,8 +37,8 @@ jobs:
         pip3 install kfp==1.8.22
         kustomize build apps/profiles/upstream/overlays/kubeflow | kubectl apply -f -
         kubectl wait --for=condition=Ready pods --all --all-namespaces --timeout 180s
-        kustomize build common/user-namespace/base | awk '!/well-defined/' | kubectl apply -f -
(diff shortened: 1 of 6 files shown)


annajung:
cc @kimwnasptd

kimwnasptd:
@annajung I remember that we had to use the `awk '!/well-defined/'` command because of this bug in kustomize https://github.com/kubernetes-sigs/kustomize/issues/5039. 

Can we avoid these awk commands by instead using [kustomize 5.0.1](https://github.com/kubernetes-sigs/kustomize/releases/tag/kustomize%2Fv5.0.1)?

annajung:
I think we can upgrade to [kustomize 5.0.3](https://github.com/kubernetes-sigs/kustomize/releases/tag/kustomize%2Fv5.0.3)

I ran a few tests to compare results between kustomize 5 with hack vs. [kustomize 5.0.3](https://github.com/kubernetes-sigs/kustomize/releases/tag/kustomize%2Fv5.0.3) without the hack and looks like output is the same. I'll make a PR to upgrade Kustomize and update this PR.

PR #2486, opened 2023-07-17
Upgrade Knative serving to 1.10.2 and eventing to 1.10.1

annajung:
Part of https://github.com/kubeflow/manifests/issues/2450

- Upgrade Knative serving to 1.10.2
- Upgrade Knative eventing to 1.101

cc @kimwnasptd

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ used from the different projects of Kubeflow:
 | Component | Local Manifests Path | Upstream Revision |
 | - | - | - |
 | Istio | common/istio-1-17 | [1.17.3](https://github.com/istio/istio/releases/tag/1.17.3) |
-| Knative | common/knative/knative-serving <br /> common/knative/knative-eventing | [1.8.1](https://github.com/knative/serving/releases/tag/knative-v1.8.1) <br /> [1.8.1](https://github.com/knative/eventing/releases/tag/knative-v1.8.1) |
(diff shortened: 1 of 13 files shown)


annajung:
I used upterm to see what's happening inside the cluster, and it looks like a few of the pods are not coming up because of insufficient resources, so they're in a pending state until the timeout hits, which results in a failed test.

I will take a deeper look into it this week

annajung:
Due to insufficient resources and the default setting that does not allow scheduling pods on the control plane nodes, at some point in the test, pods cannot run and stay in the pending state and fails to meet the condition check for all pods in "ready" state.

It looks like there are two approaches to solving the issue, both have been tested and were successful in running the KServe test
1. Allow scheduling on the control plane
2. Add another worker for kind installation

Not sure which one is better approach @kimwnasptd thoughts?

kimwnasptd:
@annajung I'd also vote for adding another worker node in our config file, just to unblock us for now.

annajung:
@yuzisun  Looks like KServe is using serving 1.10.1 and eventing 1.10.0 according to the [quick install script](https://github.com/kserve/kserve/blob/9b917d8349e4885abe4621db47def0a385ecd76d/hack/quick_install.sh#L34C1-L35C45) and the [github actions](https://github.com/kserve/kserve/blob/9b917d8349e4885abe4621db47def0a385ecd76d/test/scripts/gh-actions/setup-kourier.sh#L27), do you have any concerns with 1.10.2 and eventing 1.10.1?

we like to get this in before the feature freeze Aug 2nd

PR #2504, opened 2023-08-11
[kubeflow 1.8] Update kserve manifests from v0.11.0

DnPlas:
Sync kserve/kserve tag v0.11.0 manifests

WG leads @yuzisun
WG liaison @helloericsf 
cc: @annajung @jbottum

### DIFF
--- a/contrib/kserve/kserve/kserve_kubeflow.yaml
+++ b/contrib/kserve/kserve/kserve_kubeflow.yaml
@@ -2,8 +2,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.4.0
-  creationTimestamp: null
+    controller-gen.kubebuilder.io/version: v0.12.0
   labels:
     app: kserve
     app.kubernetes.io/name: kserve
@@ -84,6 +83,7 @@ spec:
                                       type: object
                                     type: array
                                 type: object
+                                x-kubernetes-map-type: atomic
                               weight:
                                 format: int32
                                 type: integer
@@ -130,10 +130,12 @@ spec:
                                       type: object
                                     type: array
                                 type: object
+                                x-kubernetes-map-type: atomic
                               type: array
                           required:
                             - nodeSelectorTerms
                           type: object
+                          x-kubernetes-map-type: atomic
                       type: object
                     podAffinity:
                       properties:
@@ -165,6 +167,7 @@ spec:
                                           type: string
                                         type: object
                                     type: object
+                                    x-kubernetes-map-type: atomic
                                   namespaceSelector:
                                     properties:
                                       matchExpressions:
@@ -188,6 +191,7 @@ spec:
                                           type: string
                                         type: object
                                     type: object
+                                    x-kubernetes-map-type: atomic
                                   namespaces:
                                     items:
                                       type: string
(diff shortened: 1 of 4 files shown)


contrib/kserve/kserve/kserve_kubeflow.yaml
```diff
@@ -20110,7 +20630,8 @@ apiVersion: admissionregistration.k8s.io/v1
 kind: MutatingWebhookConfiguration
 metadata:
   annotations:
-    cert-manager.io/inject-ca-from: kubeflow/serving-cert
+    cert-manager.io/inject-ca-from: kserve/serving-cert
```
annajung, line 20633: There are several places where annotation is `cert-manager.io/inject-ca-from: kserve/serving-cert`, but should be `cert-manager.io/inject-ca-from: kubeflow/serving-cert`

annajung:
Verified that changing back to kubeflow successfully creates inference service

Also, the Kserve version should be updated in the `./contrib/kserve/tests/requirements.txt` to [0.11](https://pypi.org/project/kserve/0.11.0/)

DnPlas:
> Verified that changing back to kubeflow successfully creates inference service
> 
> Also, the Kserve version should be updated in the `./contrib/kserve/tests/requirements.txt` to [0.11](https://pypi.org/project/kserve/0.11.0/)

ACK, I used the script as is. I can definitely make those changes manually, but I wonder if kserve wants to modify this in their own repo to prevent kubeflow/manifests diverging from kserve/kserve. 
In these scenarios what's the recommended way? @annajung

annajung:
Maybe start with an issue to kserve/kserve repo, to notify them of this and to understand when they'll be able to provide new release with the fix? Not sure if this will require patch release, but I would assume it would? cc @yuzisun 

I would say, base on their timeline and kubeflow release timeline, we should decide how to move forward. If the timeline doesn't align, then I would suggest manual fix for now so that it doesn't block the release schedule and can be updated when we cut future RCs.

DnPlas:
> 

Yeah, I agree. If the dates don't align with kserve by the time all the other components are synced, we can proceed and make the change manually. I'll open an issue and ping the team. Thanks @annajung.

yuzisun:
@DnPlas @annajung Can you help update the kserve_kubeflow.yaml ? we have fixed the 0.11 [kubeflow release artifact](https://github.com/kserve/kserve/releases/download/v0.11.0/kserve_kubeflow.yaml).

contrib/kserve/kserve/kserve_kubeflow.yaml
```diff
@@ -20087,9 +20607,9 @@ metadata:
   name: serving-cert
   namespace: kubeflow
 spec:
-  commonName: kserve-webhook-server-service.kubeflow.svc
+  commonName: kserve-webhook-server-service.kserve.svc
   dnsNames:
-    - kserve-webhook-server-service.kubeflow.svc
+    - kserve-webhook-server-service.kserve.svc
```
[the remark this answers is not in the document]
DnPlas: Resolving, see my [other comment](https://github.com/kubeflow/manifests/pull/2504#discussion_r1299977451)

contrib/kserve/kserve/kserve_kubeflow.yaml
DnPlas: Resolving, see my [other comment](https://github.com/kubeflow/manifests/pull/2504#discussion_r1299977451)

DnPlas:
Updated and ready for approval @kimwnasptd

Issue #2507, opened 2023-08-16
MySQL Pod fails to become ready on KinD cluster


kimwnasptd:
We've seen in the past that the Pod with MySQL never becomes ready in a KinD cluster https://github.com/kubeflow/manifests/pull/2273#issuecomment-1233092205

It keeps on biting us, so let's try to fix it once and for all. We also saw it when preparing for the 1.8 release in https://github.com/kubeflow/manifests/pull/2506

Doing a `kubectl describe` showed the following interesting event:
```
Warning  Unhealthy         15s (x5 over 75s)  kubelet            Startup probe failed: mysqladmin: [Warning] Using a password on the command line interface can be insecure.
mysqladmin: connect to server at 'localhost' failed
error: 'Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2)'
Check that mysqld is running and that the socket: '/var/run/mysqld/mysqld.sock' exists!
```

And I see the following logs 
<details><summary>Logs</summary>

```
2022-08-31 15:25:46+00:00 [Note] [Entrypoint]: Entrypoint script for MySQL Server 8.0.29-1.el8 started.
2022-08-31 15:25:46+00:00 [Note] [Entrypoint]: Switching to dedicated user 'mysql'
2022-08-31 15:25:46+00:00 [Note] [Entrypoint]: Entrypoint script for MySQL Server 8.0.29-1.el8 started.
2022-08-31 15:25:46+00:00 [Note] [Entrypoint]: Initializing database files
2022-08-31T15:25:46.194799Z 0 [System] [MY-013169] [Server] /usr/sbin/mysqld (mysqld 8.0.29) initializing of server in progress as process 48
2022-08-31T15:25:46.194840Z 0 [ERROR] [MY-010338] [Server] Can't find error-message file '/usr/share/mysql-8.0/errmsg.sys'. Check error-message file location and 'lc-messages-dir' configuration directive.
2022-08-31T15:25:46.199453Z 1 [System] [MY-013576] [InnoDB] InnoDB initialization has started.
2022-08-31T15:25:46.537790Z 1 [System] [MY-013577] [InnoDB] InnoDB initialization has ended.
2022-08-31T15:25:47.566600Z 6 [Warning] [MY-010453] [Server] root@localhost is created with an empty password ! Please consider switching off the --initialize-insecure option.
2022-08-31 15:25:49+00:00 [Note] [Entrypoint]: Database files initialized
[...]

PR #2508, opened 2023-08-16
Update Katib tests for MySQL issue with GH Runner

kimwnasptd:
Resolves https://github.com/kubeflow/manifests/issues/2507

Playing around with instructions found in https://kind.sigs.k8s.io/docs/user/known-issues/#apparmor and https://github.com/actions/runner-images/issues/181#issuecomment-610874237

### DIFF
--- a/.github/workflows/katib_kind_test.yaml
+++ b/.github/workflows/katib_kind_test.yaml
@@ -26,6 +26,13 @@ jobs:
     - name: Install cert-manager
       run: ./tests/gh-actions/install_cert_manager.sh
 
+    # https://kind.sigs.k8s.io/docs/user/known-issues/#apparmor
+    - name: AppArmor
+      run: |
+        set -x
+        sudo apt-get install apparmor-profiles
+        sudo apparmor_parser -R /etc/apparmor.d/usr.sbin.mysqld
+
     - name: Build & Apply manifests
       run: |
         cd apps/katib/upstream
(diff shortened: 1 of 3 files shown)


kimwnasptd:
@DnPlas @annajung great news, the apparmor change indeed solved the issue!

Now I'm getting a follow-up error, but this is K8s land issue that should be simple to fix
```
pod/katib-db-manager-5ddbffd67-2g75p condition met
pod/katib-mysql-66c8cdff4f-5z6jz condition met
pod/katib-ui-58b54d465f-ndcls condition met
Error from server (NotFound): pods "helper-pod-create-pvc-d67f7a04-01a7-4c2d-920d-cd076006fad6" not found
```

kimwnasptd:
To resolve the above I only made the test to explicitly wait for the pods in the `istio-system`, `cert-manager` and `kubeflow` namespaces.

Moving along I'm getting an error that no matching objects were found when running the command https://github.com/kubeflow/manifests/blob/9424fc7ba9a1a3a21f3d869a0b37ad3188723298/.github/workflows/katib_kind_test.yaml#L41

The issue seems to be that the Trial object is not yet created when this command is run. To mitigate I'll try to have an extra wait that ensure the Experiment becomes `Running` before checking for the underlying Trial object.

annajung:
Thanks @kimwnasptd for figuring out the issue!

/lgtm

Issue #2518, opened 2023-08-31
Change KF 1.8.0 manifestes , istio-ingressgateway service to use NodePort by default


yingding:
Since the Kubeflow manifests 1.7.0 the default manifests behaviour of `istio-ingressgateway` service has been changed from `NodePort` to `ClusterIP` (https://github.com/kubeflow/manifests/blob/master/common/istio-1-16/istio-install/base/patches/service.yaml)

Should there be no local loadbalancer e.g. Metallb configured, the External-Cluster IP shows `<None>`, and there will also be no NodePorts.

I suggest to change it back for KF 1.8.0 manifests

yingding:
> https://oteemo.com/think-nodeport-kubernetes/ is one of the reasons we do not want to have nodeports.

Thanks Julius, I see. I understand the intension to use `ClusterIP` by default to prevent NodePorts exposure.
I am closing this issue.

yingding:
> https://oteemo.com/think-nodeport-kubernetes/ is one of the reasons we do not want to have nodeports.

Thanks Julius, I see. I understand the intension to use `ClusterIP` by default to prevent NodePorts exposure.
I am closing this issue.

PR #2519, opened 2023-08-31
change the default istio-ingressgateway service to use NodePort

yingding:
Patch for issue #2518 to allow istio-ingressgateway service to have NodePort be default


juliusvonkohout:
Hello, 

@yingding nodeport is a security nightmare. We switched away from it on purpose. Please do not use nodeport on your cluster. It is really bad. If I were you, I would not like to take the legal liability for something like that.

But we really welcome your effort. Do you want to join the manifest WG meeting next time? We can discuss together and find something to contribute :-) Also tomorrow is the notebooks WG meeting where you can join.

PR #2481, opened 2023-07-13
Update AuthService image for KF 1.6

kimwnasptd:
**Which issue is resolved by this Pull Request:**
Part of #2469 

**Description of your changes:**
Mirrored the `gcr.io/arrikto/kubeflow/oidc-authservice:e236439` over to `docker.io/kubeflowmanifestswg/oidc-authservice:e236439`

PR #2521, opened 2023-09-04
chore: replace arrikto/kubeflow container registry with kubeflowmanifestswg

DnPlas:
Use the kubeflow manifests WG mirrored container images for pulling oidc-authservice.

Part of #2469

Mirrored the `gcr.io/arrikto/kubeflow/oidc-authservice:e236439` over to `docker.io/kubeflowmanifestswg/oidc-authservice:e236439`

cc: @kimwnasptd @annajung

### DIFF
--- a/common/oidc-client/oidc-authservice/base/kustomization.yaml
+++ b/common/oidc-client/oidc-authservice/base/kustomization.yaml
@@ -43,5 +43,5 @@ configurations:
   - params.yaml
 images:
   - name: gcr.io/arrikto/kubeflow/oidc-authservice
-    newName: gcr.io/arrikto/kubeflow/oidc-authservice
+    newName: docker.io/kubeflowmanifestswg/oidc-authservice
     newTag: e236439


common/oidc-client/oidc-authservice/base/kustomization.yaml
```diff
@@ -42,6 +42,6 @@ vars:
 configurations:
   - params.yaml
 images:
-  - name: gcr.io/arrikto/kubeflow/oidc-authservice
```
kimwnasptd, line 45: We'll only need to update the `newName` key with the dockerhub registry, and make sure the `name` matches what is initially in the StatefulSet

common/oidc-client/oidc-authservice/base/statefulset.yaml
```diff
@@ -18,7 +18,7 @@ spec:
       serviceAccountName: authservice
       containers:
       - name: authservice
-        image: gcr.io/arrikto/kubeflow/oidc-authservice:e236439
```
kimwnasptd, line 21: we'll need to not touch this at all, in order for the image transformer to use `name` and detect the old arrikto image and then `newName` to convert to the new name

kimwnasptd (commented):
@DnPlas left some small comments. High overview is that we only need to update the `newName`, and let `name` match the previous value in the image transformer.

Take also a look on how this is done in https://github.com/kubeflow/manifests/pull/2481 or https://github.com/kubeflow/manifests/pull/2480

common/oidc-client/oidc-authservice/base/kustomization.yaml
DnPlas: Changed, thanks!

common/oidc-client/oidc-authservice/base/statefulset.yaml
DnPlas: Changed, thanks!

kimwnasptd:
thanks!

/approve

PR #2529, opened 2023-09-14
feat: add pvcviewer to sync script

DnPlas:
Update the `hack/sync-kubeflow-manifests.sh` to be able to sync the pvcviewer-controller.

This change has to be backported to `v1.8-branch`

cc: @kimwnasptd


kimwnasptd:
Thanks @DnPlas!

/lgtm
/approve

Issue #2524, opened 2023-09-11
PVCViewer manifests weren't copied to v1.8.0-rc.0


TobiasGoerke:
The [manifests](https://github.com/kubeflow/kubeflow/tree/master/components/pvcviewer-controller/config) of the newly added PVCViewer weren't copied to the current release, causing the volumes web app to throw errors when opening a viewer.

Also, we should merge the following PRs before doing so:
* https://github.com/kubeflow/kubeflow/pull/7237
* https://github.com/kubeflow/kubeflow/pull/7247

annajung:
cc @DnPlas @kubeflow/release-team to track for 1.8

DnPlas:
hey @TobiasGoerke , thanks for raising this issue. Regarding the PVCViewer manifests, I recently synced them in `master`, you can find them [here](https://github.com/kubeflow/manifests/tree/master/apps/pvcviewer-controller/upstream) and the change should land soon in `v1.8-branch` https://github.com/kubeflow/manifests/pull/2535.

TobiasGoerke:
Great, thanks for taking care of this, @DnPlas!

Shall we also add the pvcviewer-controller to the `example/kustomization.yaml`? Like:
`- ../apps/pvcviewer-controller/upstream/base`

kimwnasptd:
@TobiasGoerke yes, we'll need to add it. I can make a PR for this

PR #2539, opened 2023-09-19
chore: add pvcviewer-controller to `example/kustomization.yaml`

DnPlas:
Add pvcviewer-controller to `example/kustomization.yaml`
Part of #2524 

This change should be forward ported to `master`

cc: @kimwnasptd


kimwnasptd:
I confirm with the above we no longer see errors in the volumes web app, due to errors trying to fetch PVCViewer CRs when the CRD is not applied.

Issue #2517, opened 2023-08-30
oauth2-proxy and programmatic access clients


kromanow94:
## Description

This is a follow up for https://github.com/kubeflow/manifests/pull/2409.

It was discussed that the oauth2-proxy integration with Istio should also cover m2m setup with authorization bearer tokens issued by oidc provider with emphasis on also supporting the default kubernetes oidc served behind `https://kubernetes.default.svc.cluster.local`.

The exact setup and reasoning was described it this and following comments: https://github.com/kubeflow/manifests/pull/2409#issuecomment-1694495858.

`oauth2-proxy` should be configured to skip jwt validation when a specific `issuer=audience` pair is provided with `extra_jwt_issuers` configuration option. This pair will be `https://kubernetes.default.svc.cluster.local=https://kubernetes.default.svc.cluster.local` for the default k8s oidc. This will forward the traffic directly to Istio and then we can have a `RequestAuthentication` resource that will verify the JWT and authorize access to the services based on JWT Payload.

As mentioned in the https://github.com/kubeflow/manifests/pull/2409, I'll provide a PR for this feature.

### Istio
When using the default k8s oidc (which k8s configures by default if not provided with external oidc client details), the issuer is somewhat problematic. This is because the issuer is served behind self-signed certificates and accessible under endpoint that requires authentication.

Configuring traffic with `RequestAuthentication` for the default k8s oidc like below will not work and `istiod` will output following log:
```yaml
apiVersion: security.istio.io/v1beta1
kind: RequestAuthentication
metadata:
  name: kubernetes-jwt
  namespace: istio-system
spec:
  jwtRules:
    - forwardOriginalToken: true
      issuer: https://kubernetes.default.svc.cluster.local
```

```
2023-08-29T14:46:13.084271Z     info    model   The JWKS key is not yet fetched for issuer https://kubernetes.default.svc.cluster.local/ (), using a fake JWKS for now
```

Following the log about JWKS and knowing the detail about cert and auth of oidc issuer, I was able to verify that if we were to take the JWKS and provide it directly as string in `RequestAuthentication`, this issue is not happening anymore.

#### Example
```
$ curl https://kubernetes.default.svc.cluster.local/.well-known/openid-configuration -H "Authorization: Bearer $(cat /run/secrets/kubernetes.io/serviceaccount/token)" | jq
{
  "issuer": "https://kubernetes.default.svc.cluster.local/",
[...]

PR #2544, opened 2023-10-03
oauth2-proxy with istio mesh config and m2m bearer tokens

### LINKED ISSUE #2516: oauth2-proxy and Istio mesh support
## Description

This is a follow up for https://github.com/kubeflow/manifests/pull/2409.

It was discussed that the oauth2-proxy integration with Istio should be managed with Istio Mesh Config instead of EnvoyFilter. The exact setup and reasoning was described it this and following comments: https://github.com/kubeflow/manifests/pull/2409#issuecomment-1694495858.

Summary: it's better to use mesh config with `envoyExtAuthzHttp` extension provider instead of `EnvoyFilter` because it's better supported, it's described by Istio to be used in this way, provides the integration with oauth2-proxy for the whole istio cluster that configurable with `AuthorizationPolicy` and mitigates issues described in https://github.com/istio/istio/issues/27790.

As mentioned in the https://github.com/kubeflow/manifests/pull/2409, I'll provide a PR for this feature.

@kimwnasptd @juliusvonkohout @axel7083

kromanow94:
Resolves #2516, #2517

Changes are based on discussion and description available here:
* https://github.com/kubeflow/manifests/pull/2409#issuecomment-1694495858
* https://github.com/kubeflow/manifests/issues/2516
* https://github.com/kubeflow/manifests/issues/2517

**TL;DR:**
* Drop EnvoyFilter in oauth2-proxy kustomization
* Use Istio `envoyExtAuthzHttp` instead of EnvoyFilter
* Enable m2m bearer tokens
* Use kustomize components for configuration tailoring
* Enable scenario with in-cluster Kubernetes OIDC served behind self-signed certificates
* Explain new auth routine
* Use oauth2-proxy in logout routine

**Testing setup:**
In `common/oidc-client/oauth2-proxy/README.md` are instructions on how to enable usage of `oauth2-proxy`. This covers the example scenario with Kubeflow deployed on kind, which enables configuration for in-cluster OIDC behind self-signed certificates.

To verify if oauth2-proxy is working for standard user login and m2m tokens:
* try to login to Kubeflow with example user/password. After that, you should be able to see:
    * logged in user is `user@example.com`
    * there is a cookie named `oauth2_proxy_kubeflow` in your browser
* create a Pod with `curl` image and try calling `istio-ingressgateway` with Authorization Bearer Token from Service Account:
    ```bash
    [user@example ~]$ kubectl run -ti --rm --image curlimages/curl curl --command -- sh
    ~ $ curl istio-ingressgateway.istio-system/api/workgroup/exists -H "Authorization: Bearer $(cat /run/secrets/kubernetes.io/serviceaccount/token)"
    {"hasAuth":true,"user":"system:serviceaccount:default:default","hasWorkgroup":false,"registrationFlowAllowed":false}
    ```

**Comments:**
@axel7083 , I'd appreciate your feedback and it would be great if you could test this setup

@juliusvonkohout , I know the file structure looks a little bit different but I thought it makes a little bit more sense. That said, I'm open for discussion and we can change the file structure.

From implementation perspective everything is in place. The only missing parts are adding some more README, probably doing some clean up and addressing feedback.

I'd appreciate review.

---

## Discovered missing points

- [x] enable oauth2-proxy as default
- [x] enable m2m by default
- [x] enable self-signed certs by default
- [x] write e2e tests for m2m - https://github.com/kubeflow/manifests/pull/2417
  - [x] list pipeline runs
  - [x] list experiments in a namespace
  - [x] starting a pipeline - good to have if time allows, might require considerate amount of work
  - [x] test kserve (create new e2e github workflow based on the current kserve test)
    - use only the scenario with standard istio installation (not cni)
- [x] HA for oauth2-proxy with multiple replicas
- [x] describe oauth2-proxy upgrade instructions


juliusvonkohout:
@kromanow, i can take a look at this next week. Please sign the CLA as well.

kromanow94:
@juliusvonkohout I signed the CLA.

juliusvonkohout:
I think we are missing tests. as for example https://github.com/kubeflow/manifests/blob/master/tests/gh-actions/install_knative-cni.sh and https://github.com/kubeflow/manifests/blob/master/tests/gh-actions/install_istio-cni.sh could be extended with oauth 2

kromanow94:
@juliusvonkohout thanks for the review!

About the tests - good point. I'm not sure about the "could be extended with oauth2" part. You mean not to create new tests for oauth2-proxy but just extend the ones for knative and istio-cni?

I think until there is support for oidc-authservice, I'd suggest having those as a separate tests. But, if you think extending the current ones is good enough - that's fine with me.

common/oidc-client/oauth2-proxy/README.md
```diff
@@ -0,0 +1,213 @@
...
+
+The details of this custom integration can be found here:
+* https://github.com/arrikto/oidc-authservice/blob/0c4ea9a/server.go#L509
+* https://github.com/kubeflow/kubeflow/blob/c6c4492/components/centraldashboard/public/components/logout-button.js#L50
+
+To login again, user have to refresh the page.
+
+## Example installation
+To install Kubeflow configured to use `oauth2-proxy` with `istio`
+`envoyExtAuthzHttp` extension, following changes has to be done to the
```
juliusvonkohout, line 149: You are very often missing "the" in your sentences within the readme. For example "extension, following changes has to be done to the" should be "extension, THE following changes HAVE to be applied to the"

common/oidc-client/oauth2-proxy/README.md
```diff
@@ -0,0 +1,213 @@
...
+
+If this is not configured, user will be presented with following Istio error:
+```
+Jwks doesn't have key to match kid or alg from Jwt
+```
+
+## Issues with this setup
+While not directly an Istio or oauth2-proxy issue, currently Kubeflow is configured
+to always redirect to the URL provided in logout response body object under key
+`afterLogoutURL`. This is custom integration with `oidc-authservice` component.
```
juliusvonkohout, line 136: body object under THE key
`afterLogoutURL`. This is A custom integration with the `oidc-authservice` component.

common/oidc-client/oauth2-proxy/README.md
```diff
@@ -0,0 +1,213 @@
...
+oauth2-proxy has redirect capabilities and can redirect to the base Kubeflow
+page but because of this custom integration, currently logging out will remove
```
juliusvonkohout, line 138: Kubeflow page, but

common/oidc-client/oauth2-proxy/README.md
```diff
@@ -0,0 +1,213 @@
...
+the authentication cookie but not redirect to the Kubeflow Home Page.
+
+The details of this custom integration can be found here:
+* https://github.com/arrikto/oidc-authservice/blob/0c4ea9a/server.go#L509
+* https://github.com/kubeflow/kubeflow/blob/c6c4492/components/centraldashboard/public/components/logout-button.js#L50
+
+To login again, user have to refresh the page.
+
+## Example installation
+To install Kubeflow configured to use `oauth2-proxy` with `istio`
```
juliusvonkohout, line 148: @kimwnasptd shall we make it the default in this PR?

common/oidc-client/oauth2-proxy/base/README.md
```diff
@@ -0,0 +1,6 @@
+# oauth2-proxy
+
+This deployment of `oauth2-proxy` was configured in a way that's as close as
```
juliusvonkohout, line 3: Do we have upgrade instructions for auth2-proxy?

common/oidc-client/oauth2-proxy/base/deployment.yaml
```diff
@@ -1,18 +1,16 @@
...
+    app: oauth2-proxy
 spec:
   replicas: 1
   selector:
     matchLabels:
       app.kubernetes.io/name: oauth2-proxy
   template:
     metadata:
-      annotations:
-        sidecar.istio.io/inject: "false"
```
juliusvonkohout, line 15: Why is this annotation being removed?

common/oidc-client/oauth2-proxy/base/kustomization.yaml
```diff
@@ -1,42 +1,58 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
+namespace: oauth2-proxy
```
juliusvonkohout, line 4: @kimwnasptd before it was in the istio-system namespace and dex is still in the auth namespace if i remeber correctly

juliusvonkohout:
> @juliusvonkohout thanks for the review!
> 
> About the tests - good point. I'm not sure about the "could be extended with oauth2" part. You mean not to create new tests for oauth2-proxy but just extend the ones for knative and istio-cni?
> 
> I think until there is support for oidc-authservice, I'd suggest having those as a separate tests. But, if you think extending the current ones is good enough - that's fine with me.

For the time being extending them might be enough. @kimwnasptd what do you think?
I think we should at least test the human session and the m2m authentication in the github workflow.

common/oidc-client/oauth2-proxy/base/kustomization.yaml
```diff
@@ -1,42 +1,58 @@
...
+  # If Kubernetes is managed by kind, vCluster, minikube or similar tool,
+  # most probably the Kubernetes OIDC Issuer will be server in-cluster
+  # behind self-signed certs. This option will configure
+  # --ssl-insecure-skip-verify which will accept self-signed-certificates.
+  - ALLOW_SELF_SIGNED_ISSUER=false
+
+  # This will configure oauth2-proxy option --skip-jwt-bearer-tokens which
+  # will pass the requests with 'Authorization' header with Bearer Token
+  # matching --extra-jwt-issuers.
+  - ENABLE_M2M_TOKENS=false
```
juliusvonkohout, line 43: Should this not be enabled by default?

common/oidc-client/oauth2-proxy/components/allow-unauthenticated-issuer-discovery/README.md
```diff
@@ -0,0 +1,45 @@
+# Unauthenticated issuer discovery
+
+If you're running Kubernetes with kind, vCluster, minikube or some other
+tool for local development on Kubernetes, there is a high change that the
```
juliusvonkohout, line 4: then there is a high chance that the Kubernetes OIDC Issuer is managed within the cluster behind self-signed
certificates.

common/oidc-client/oauth2-proxy/components/configure-self-signed-kubernetes-oidc-issuer/README.md
juliusvonkohout, line 1: The readme is empty and i am wondering whether running the scrip once is enough in case of istio restarts.

common/oidc-client/oauth2-proxy/components/istio-external-auth/patches/deployment.jwt-refresh-interval.yaml
```diff
@@ -0,0 +1,18 @@
...
+    spec:
+      containers:
+        - name: discovery
+          env:
+            # Istio will refresh the JWT Public Keys provided with
+            # RequestAuthentication by default every 20 minutes. For deployment
+            # from scratch this is not ideal because there is a high chance that
+            # the istiod will be available before dex is available, triggering
+            # Istio mechanism to use a placeholder jwt until refreshed.
+            - name: PILOT_JWT_PUB_KEY_REFRESH_INTERVAL
```
juliusvonkohout, line 17: Will there be any service interruption or high load on the system because of that?

common/oidc-client/oauth2-proxy/components/istio-m2m/README.md
juliusvonkohout, line 1: Is this a TODO ?

juliusvonkohout:
Maybe not all my questions are are as elaborated and thoughtful as deserved by this PR, but i only have limited time to review this at the moment.

kromanow94:
I'm sorry for the delay in responding to the feedback on this merge request. The past few weeks have been quite busy for me, but I'm now back on track and ready to address your comments. Thanks for your patience and understanding!

In this comment I focus on the tests. I'll try to answer all other comments tomorrow and fill up the missing documentation in incoming days.

@juliusvonkohout I tried applying your suggestions around tests. In `tests/gh-actions/install_istio-cni.sh` I used the `common/istio-cni-1-17/istio-install/overlays/oauth2-proxy` which uses `common/oidc-client/oauth2-proxy/components/istio-external-auth` Kustomize Component. I also had to add the `common/oidc-client/oauth2-proxy/overlays/m2m-self-signed` to have the oauth2-proxy itself. The tests fails because now the traffic needs to be authorized. The test output is:
```
FAILED test_sklearn.py::test_sklearn_kserve - requests.exceptions.HTTPError: 403 Client Error: Forbidden for url: http://localhost:8080/v1/models/isvc-sklearn:predict
```

Currently none of the tests defined in github workflows uses any sort of authorization. I'm not sure if the model endpoints are always expected to allow anonymous access or not. I noticed that in the `contrib/kserve/tests` model is deployed without istio sidecar so in-cluster access is always allowed and only `istio-ingressgateway` protects the endpoint.

We can modify what url hosts/paths have to go through oauth2-proxy, for example excluding `/v1/models/*` so the models are always served with anonymous access. I'd apprecieate some feedback here because the side effect at current form is that all models will have to use m2m auth when accessing through `istio-ingressgateway`.

With the above in mind, I think it makes sense to create a separate github workflow testing M2M authorization with Self Signed certificates. I'd appreciate feedback on the tests scope. I think it makes sense to use a modification of `.github/workflows/kserve_kind_cni_test.yaml` and/or `.github/workflows/centraldb_kind_test.yaml`. Additionally, testing the endpoint `api/workgroup/exists` could be useful to see if the returned user information matches expected from the Authorization Bearer token (like a Service Account).

I had a look on the tests in `contrib/kserve/tests` and I was able to verify what exactly would have to be done to enable the test with m2m token. Let me present it in the form of diff:
```diff
diff --git a/.github/workflows/kserve_kind_cni_test.yaml b/.github/workflows/kserve_kind_cni_test.yaml
index b20102e9..6fc655a5 100644
--- a/.github/workflows/kserve_kind_cni_test.yaml
+++ b/.github/workflows/kserve_kind_cni_test.yaml
@@ -54,6 +54,7 @@ jobs:
       - name: Run kserve tests
         run: |
           export KSERVE_INGRESS_HOST_PORT=localhost:8080
+          export KSERVE_M2M_TOKEN="$(kubectl -n default create token default)"
           cd ./contrib/kserve/tests && pytest .
 
       - name: Run kserve models webapp test
diff --git a/contrib/kserve/tests/utils.py b/contrib/kserve/tests/utils.py
index 27de1d18..0f20aae2 100644
--- a/contrib/kserve/tests/utils.py
+++ b/contrib/kserve/tests/utils.py
@@ -70,6 +70,12 @@ def predict_str(service_name, input_json, protocol_version="v1",
     host = urlparse(isvc["status"]["url"]).netloc
     headers = {"Host": host, "Content-Type": "application/json"}
 
+    token = os.environ.get("KSERVE_M2M_TOKEN")
+    if token is not None:
+        headers.update({
+            "Authorization": f"Bearer {token}"
+        })
+
     if model_name is None:
         model_name = service_name
```

kromanow94 (commented):
I addressed all comments and added a few readmes.

juliusvonkohout:
This will close https://github.com/kubeflow/manifests/issues/2469

common/oidc-client/oauth2-proxy/base/deployment.yaml
juliusvonkohout: Yes, that makes sense

common/oidc-client/oauth2-proxy/base/deployment.yaml
```diff
@@ -21,52 +19,67 @@ spec:
...
-        - name: oauth2-proxy
-          secret:
-            secretName: oauth2-proxy
       containers:
         - name: oauth2-proxy
           image: quay.io/oauth2-proxy/oauth2-proxy:v7.4.0
           args:
-            - '--alpha-config=/etc/oauth2_proxy/oauth2_proxy.yml'
-            - '--config=/etc/oauth2_proxy/oauth2_proxy.cfg'
-            - '--silence-ping-logging'
```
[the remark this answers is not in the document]
juliusvonkohout: Yes that sound good

common/oidc-client/oauth2-proxy/base/kustomization.yaml
juliusvonkohout: Yes, we want to integrate with existing oauth deployments in enterprise environments

common/oidc-client/oauth2-proxy/base/kustomization.yaml
juliusvonkohout: Then lets enable ALLOW_SELF_SIGNED_ISSUER and ENABLE_M2M_TOKENS=true by default

juliusvonkohout:
This here is abandoned, but could be picked up https://github.com/kubeflow/manifests/pull/2417

common/oidc-client/oauth2-proxy/base/README.md
juliusvonkohout: Maybe add some documentation about that

juliusvonkohout:
ah and as long as you are not a member of the org you cannot be in the owners file. Google either has to add you or we have to remove you there for the time being.

juliusvonkohout:
@zijianjoy can you approve the workflows or give me and @kimwnasptd write rights to this repository?

.github/workflows/pipeline_m2m_kind_test.yaml
```diff
@@ -0,0 +1,82 @@
...
+          pipeline = client.list_pipelines().pipelines[0]
+          pipeline_name = pipeline.display_name
+          pipeline_id = pipeline.pipeline_id
+          pipeline_version_id = client.list_pipeline_versions(pipeline_id).pipeline_versions[0].pipeline_version_id
+          experiment_id = client.create_experiment("m2m-test", namespace="kubeflow-user-example-com").experiment_id
+
+          print(f"Starting pipeline {pipeline_name}.")
+          run_id = client.run_pipeline(experiment_id=experiment_id, job_name="m2m-test", pipeline_id=pipeline_id, version_id=pipeline_version_id).run_id
+
+          while True:
```
juliusvonkohout, line 74: Should we set a timeout of e.g. 3 minutes?

common/oidc-client/oauth2-proxy/README.md
```diff
@@ -0,0 +1,231 @@
...
+    * A state parameter to validate the authorization response's authenticity and
+      ensure it originates from the initial request.
+
+   Post-authentication, a cookie is set in the user's browser, used by Istio and `oauth2-proxy`
+   for authorizing requests.
+
+   `oauth2-proxy` also checks for an `Authorization` header with a JWT bearer token. It trusts
+   JWTs issued by the configured OIDC provider (default `dex`) and can be configured to trust
+   additional JWTs. If a valid JWT is present, `oauth2-proxy` forwards the request to Istio
+   along with the authorization header.
```
juliusvonkohout, line 118: Where does the header come from? Is it the kubeflow-userid header? How is the kubeflow-userid header derived from the JWT in the bearer token header? This all must be in detail in the big diagram.

juliusvonkohout:
This looks amazing. Now i think we need a big diagram, similar to the one i presented at Kubeflow summit, but way more detailed :-D

kromanow94:
> This looks amazing and means we can close the abandoned end-to-end PR as well then. Now i think we need a big diagram, similar to the one i presented at Kubeflow summit, but way more detailed :-D

Thanks!

I started doing auth diagram for ml-pipeline api-server with links to the lines in code where some auth magic happens. I want to combine it with a bigger-picture networking diagram. It'll be a lot of work but will definitely help around auth and networking.

juliusvonkohout:
@axel7083 do you want to review as well ?

nithin8702:
Hi, Are we good to merge this PR?

juliusvonkohout:
@kimwnasptd i plan to merge this on wednesday as discussed.

.github/workflows/pipeline_m2m_kind_test.yaml
juliusvonkohout: Or even better a for loop with 10 tries and 15 seconds sleep

.github/workflows/kserve_m2m_kind_test.yaml
```diff
@@ -0,0 +1,70 @@
...
+
+      - name: Install test dependencies
+        run: pip install -r ./contrib/kserve/tests/requirements.txt
+
+      - name: Port forward
+        run: |
+          INGRESS_GATEWAY_SERVICE=$(kubectl get svc --namespace istio-system --selector="app=istio-ingressgateway" --output jsonpath='{.items[0].metadata.name}')
+          nohup kubectl port-forward --namespace istio-system svc/${INGRESS_GATEWAY_SERVICE} 8080:80 &
+
+      - name: Run kserve tests with m2m token from SA default/default
```
juliusvonkohout, line 55: Please add some security tests:
1. userA namespace with serviceaccounttoken  from userA
2. userA namespace with serviceaccounttoken  from userA
3. userA namespace with serviceaccounttoken  from userB
4. userB namespace with serviceaccounttoken  from userA

The last two tests should fail.

common/oidc-client/oauth2-proxy/components/istio-m2m/requestauthentication.yaml
```diff
@@ -0,0 +1,14 @@
...
+  name: m2m-token-issuer
+  namespace: istio-system
+spec:
+  jwtRules:
+    - forwardOriginalToken: true
+      issuer: M2M_TOKEN_ISSUER_PLACEHOLDER
+      outputClaimToHeaders:
+        - claim: sub
+          header: x-auth-request-user
+        - claim: sub
```
juliusvonkohout, line 13: i asume that
```
        - header: kubeflow-userid
          claim: sub
```

means use the claim "sub" from the JWT (serviceaccountoken) and use it as kubeflow-userid header. This header is then used in most of our authorizationpolicies https://github.com/kimwnasptd/kubeflow/blob/52d55bef092d9ec56a1b4c6ae8d57fff241ce2fe/components/access-management/kfam/bindings.go#L95 and also in most of our webservices

Sub is typically "system:serviceaccount:kubeflowusernamespace:default-editor"
while in the DEX JWT we use the users email to set the header kubeflow-userid.
Now i am wondering how our authorizationpolicies are going to react to "system:serviceaccount:kubeflowusernamespace:default-editor" as userid header instead of the email.

kromanow94:
@juliusvonkohout the `api/workgroup/exists` endpoint is served from the `centraldashboard` and the access to the `centraldashboard` should be open afaiu.

Also, this endpoint has nothing to do with namespaced access.

See the code:
* https://github.com/kubeflow/kubeflow/blob/v1.8.0/components/centraldashboard/app/api_workgroup.ts#L254
* https://github.com/kubeflow/kubeflow/blob/v1.8.0/components/centraldashboard/public/components/main-page.pug#L1
* https://github.com/kubeflow/kubeflow/blob/v1.8.0/components/centraldashboard/public/components/main-page.js#L206
* https://github.com/kubeflow/kubeflow/blob/v1.8.0/components/centraldashboard/app/attach_user_middleware.ts

common/oidc-client/oauth2-proxy/components/istio-m2m/requestauthentication.yaml
kromanow94: Yes, I totally agree on the headers and bindings nightmare.

> use the claim "sub" from the JWT (serviceaccountoken) and use it as kubeflow-userid header

That's exactly true.

> how our authorizationpolicies are going to react to "system:serviceaccount:kubeflowusernamespace:default-editor" as userid header instead of the email

Well, technically with the m2m sa tokens you're not reaching the kubeflow user namespace directly but the KF API which has means of authorization through the `TokenReviews` and `SubjectAccessReviews`.

Let's analyze the the code in a following scenario:
1. api-server starts endpoints in

     https://github.com/kubeflow/pipelines/blob/2.0.5/backend/src/apiserver/main.go#L95

    Let's focus on the pipelines run service:
    1. Register Run Service:
        * https://github.com/kubeflow/pipelines/blob/2.0.5/backend/src/apiserver/main.go#L116
    2. proto RPC definition of ListRunsV1
        * https://github.com/kubeflow/pipelines/blob/2.0.5/backend/api/v1beta1/run.proto#L80
    3. code definition of ListRunsV1
        * https://github.com/kubeflow/pipelines/blob/2.0.5/backend/src/apiserver/server/run_server.go#L226
    4. ListRunsV1 calls internal method `listRuns`
        * https://github.com/kubeflow/pipelines/blob/2.0.5/backend/src/apiserver/server/run_server.go#L188 
    5. `listRuns` calls internal method `canAccessRun` which itself calls `s.resourceManager.IsAuthorized`
        * https://github.com/kubeflow/pipelines/blob/2.0.5/backend/src/apiserver/server/run_server.go#L637
    6. `ResourceManager.IsAuthorized` first tries to authenticate over every available authenticator, which are the `TokenReviewAuthenticator` and `HTTPHeaderAuthenticator`
        * here the user identity is either the user email provided directly in the `kubeflow-userid` header or the user identity obtained from provided token
        * https://github.com/kubeflow/pipelines/blob/master/backend/src/apiserver/resource/resource_manager.go#L1667
    8. `TokenReviewAuthenticator.GetUserIdentity` gets the token from `Authorization` header and calls the K8s Auth `authv1.TokenReview` with given token which in return provides `userInfo := review.Status.User`. `GetUserIdentity` return `userInfo.Username` which at this point is the `system:serviceaccount:default:default`.
        * https://github.com/kubeflow/pipelines/blob/2.0.5/backend/src/apiserver/auth/authenticator_token_review.go#L53
    9. Next in `ResourceManager.IsAuthorized` a SubjectAccessReview is created with `r.subjectAccessReviewClient.Create` with arguments specifying RBAC verbs provided in code definition of `RunServer.listRuns`. If the user (sa) is not authorized, an error is thrown
        * https://github.com/kubeflow/pipelines/blob/master/backend/src/apiserver/resource/resource_manager.go#L1703
        * if the identity was obtained from token (service account), the `rolebinding.rbac.authorization.k8s.io/default-editor` provides the RBAC permission
        * if the identity was obtained from header (user), the `rolebinding.rbac.authorization.k8s.io/user-example-com` or similar provides the RBAC permission

1. User calls api to list pipeline runs as unauthorized service account.
    * This can be done by running Pod with curl in `default` namespace:
        ```bash
        $ kubectl -n default run -ti --rm curl --image curlimages/curl --command -- sh
        # v1beta1
        ~ $ curl "istio-ingressgateway.istio-system/pipeline/apis/v1beta1/runs?resource_reference_key.type=NAMESPACE&resource_reference_key.id=kubeflow-user-example-com" -H "Authorization: Bearer $(cat /run/secrets/kubernetes.io/serviceaccount/token)"
        {"error":"Failed to list v1beta1 runs: Failed to list runs due to authorization error. Check if you have permission to access namespace kubeflow-user-example-com: Failed to access run . Check if you have access to namespace kubeflow-user-example-com: PermissionDenied: User 'system:serviceaccount:default:default' is not authorized with reason:  (request: \u0026ResourceAttributes{Namespace:kubeflow-user-example-com,Verb:list,Group:pipelines.kubeflow.org,Version:v1beta1,Resource:runs,Subresource:,Name:,}): Unauthorized access","code":7,"message":"Failed to list v1beta1 runs: Failed to list runs due to authorization error. Check if you have permission to access namespace kubeflow-user-example-com: Failed to access run . Check if you have access to namespace kubeflow-user-example-com: PermissionDenied: User 'system:serviceaccount:default:default' is not authorized with reason:  (request: \u0026ResourceAttributes{Namespace:kubeflow-user-example-com,Verb:list,Group:pipelines.kubeflow.org,Version:v1beta1,Resource:runs,Subresource:,Name:,}): Unauthorized access","details":[{"@type":"type.googleapis.com/google.rpc.Status","code":7,"message":"User 'system:serviceaccount:default:default' is not authorized with reason:  (request: \u0026ResourceAttributes{Namespace:kubeflow-user-example-com,Verb:list,Group:pipelines.kubeflow.org,Version:v1beta1,Resource:runs,Subresource:,Name:,})"}]}
        # v2beta1
        ~ $ curl istio-ingressgateway.istio-system/pipeline/apis/v2beta1/runs?namespace=kubeflow-user-example-com -H "Authorization: Bearer $(cat /run/secrets/kubernetes.io/serviceaccount/token)"
        {"error":"Failed to list runs: Failed to list runs due to authorization error. Check if you have permission to access namespace kubeflow-user-example-com: Failed to access run . Check if you have access to namespace kubeflow-user-example-com: PermissionDenied: User 'system:serviceaccount:default:default' is not authorized with reason:  (request: \u0026ResourceAttributes{Namespace:kubeflow-user-example-com,Verb:list,Group:pipelines.kubeflow.org,Version:v1beta1,Resource:runs,Subresource:,Name:,}): Unauthorized access","code":7,"message":"Failed to list runs: Failed to list runs due to authorization error. Check if you have permission to access namespace kubeflow-user-example-com: Failed to access run . Check if you have access to namespace kubeflow-user-example-com: PermissionDenied: User 'system:serviceaccount:default:default' is not authorized with reason:  (request: \u0026ResourceAttributes{Namespace:kubeflow-user-example-com,Verb:list,Group:pipelines.kubeflow.org,Version:v1beta1,Resource:runs,Subresource:,Name:,}): Unauthorized access","details":[{"@type":"type.googleapis.com/google.rpc.Status","code":7,"message":"User 'system:serviceaccount:default:default' is not authorized with reason:  (request: \u0026ResourceAttributes{Namespace:kubeflow-user-example-com,Verb:list,Group:pipelines.kubeflow.org,Version:v1beta1,Resource:runs,Subresource:,Name:,})"}]}
        ```

1. User calls api to list pipeline runs as authorized service account.
    * This can be done by running Pod with curl in `kubeflow-user-example-com` namespace specifying correct service account:
        ```bash
        $ kubectl -n kubeflow-user-example-com run -ti --rm curl --image curlimages/curl --command --overrides='{"spec": {"serviceAccountName": "default-editor"}}' -- sh
        # v1beta1
        ~ $ curl "istio-ingressgateway.istio-system/pipeline/apis/v1beta1/runs?resource_reference_key.type=NAMESPACE&resource_reference_key.id=kubeflow-user-example-com" -H "Authorization: Bearer $(cat /run/secrets/kubernetes.io/serviceaccount/token)"
        {}  # empty response which is fine because no pipeline runs exist
        # v2beta1
        ~ $ curl istio-ingressgateway.istio-system/pipeline/apis/v2beta1/runs?namespace=kubeflow-user-example-com -H "Authorization: Bearer $(cat /run/secrets/kubernetes.io/serviceaccount/token)"
        {}  # empty response which is fine because no pipeline runs exist
        ```

Having that said, I don't think we have to tailor the RoleBindings since the RBAC is already properly configured.

The AuthorizationPolicies might require a little tailoring for kserve... I wonder what are the scenarios when we want to access the Pod/SVC run in a given Namespace directly.

.github/workflows/kserve_m2m_kind_test.yaml
kromanow94: For Notebooks and Pipelines it makes a lot of sense but I'm having doubts for kserve - currently in the test scenario there is just a new namespace create and the model is not served as a part of a kubeflow profile. Because of that there is not authorizationpolicy that controls access to the model.

I admit my experience with kserve is limited so some guidance or at least relevant links would be appreciated, specifically about listing all kserver models. Also, is it standard to deploy kserve model in Kubeflow outside of KF Profile?

common/oidc-client/oauth2-proxy/components/istio-m2m/requestauthentication.yaml
juliusvonkohout: I have to read this tomorrow, but pipelines is a bad example because it has special stuff for handling service account tokens. I am more interested in Notebooks, Kserve. Can you provide this analysis for them as well?

juliusvonkohout: We really need test for kserve, notebooks, pipelines automation via m2m tokens

kromanow94: ## Simplified analysis for Notebooks:

* list notebooks api route definition
  * https://github.com/kubeflow/kubeflow/blob/v1.8.0/components/crud-web-apps/jupyter/backend/apps/common/routes/get.py#L53
  * this calls `crud_backend/api/notebook.py::list_notebooks`
* `crud_backend/api/notebook.py::list_notebooks` calls `authz.ensure_authorized`
  * https://github.com/kubeflow/kubeflow/blob/v1.8.0/components/crud-web-apps/common/backend/kubeflow/kubeflow/crud_backend/api/notebook.py#L24
* `crud_backend/authz.py::ensure_authorized` calls `crud_backend/authn.py::get_username`
  * https://github.com/kubeflow/kubeflow/blob/v1.8.0/components/crud-web-apps/common/backend/kubeflow/kubeflow/crud_backend/authz.py#L101
  * https://github.com/kubeflow/kubeflow/blob/v1.8.0/components/crud-web-apps/common/backend/kubeflow/kubeflow/crud_backend/authn.py#L12
  * `crud_backend/authn.py::get_username` gets the user id from userid header (email or sa in format `system:serviceaccount:kubeflowusernamespace:default-editor`)
* `crud_backend/authz.py::ensure_authorized` calls `crud_backend/authz.py::is_authorized`
  * https://github.com/kubeflow/kubeflow/blob/v1.8.0/components/crud-web-apps/common/backend/kubeflow/kubeflow/crud_backend/authz.py#L46
  * this calls `create_subject_access_review` which uses the same mechanism as pipelines with `r.subjectAccessReviewClient.Create`

### Summary
Notebooks api uses the same mechanism for user authentication, proxying auth decision to K8s RBAC. 
* if the identity was obtained from token (service account), the `rolebinding.rbac.authorization.k8s.io/default-editor` provides the RBAC permission
* if the identity was obtained from header (user), the `rolebinding.rbac.authorization.k8s.io/user-example-com` or similar provides the RBAC permission

kromanow94: On kserve I found this information:
* https://github.com/kserve/kserve/blob/v0.12.0/config/default/manager_auth_proxy_patch.yaml

This suggests the kserve **might** use the same mechanism based on `SubjectAccessReviews`. Having a look at the kubeflow/manifests I see it's not enabled.

Also, looking through the docs and code I was only able to find this:
* https://github.com/kserve/kserve/tree/v0.12.0/docs/samples/istio-dex
* https://github.com/kserve/kserve/tree/v0.12.0/docs/samples/gcp-iap

The docs above mention that while it's possible to enable authentication, authorization is more complicated and probably we need to add `AuthorizationPolicy`...
> create an [Istio AuthorizationPolicy](https://istio.io/latest/docs/reference/config/security/authorization-policy/) to grant access to the pods or disable it

I executed an experiment on example kubeflow installation to verify access from different namespace to kserve inferenceservice deployed in `kubeflow-user-example-com`:
1. Created InferenceService with sklearn
    ```yaml
    apiVersion: serving.kserve.io/v1beta1
    kind: InferenceService
    metadata:
      name: isvc-sklearn
      namespace: kubeflow-user-example-com
    spec:
      predictor:
        minReplicas: 1
        model:
          modelFormat:
            name: sklearn
          name: ''
          resources:
            limits:
              cpu: 100m
              memory: 256Mi
            requests:
              cpu: 50m
              memory: 128Mi
          storageUri: gs://kfserving-examples/models/sklearn/1.0/model
    ```
2. Created AuthorizationPolicy for given InferenceService Workload:
    ```yaml
    apiVersion: security.istio.io/v1
    kind: AuthorizationPolicy
    metadata:
      labels:
        app: isvc-sklearn-predictor-00001
      name: isvc-sklearn-predictor-00001
      namespace: kubeflow-user-example-com
    spec:
      action: DENY
      rules:
      - {}
      selector:
        matchLabels:
          app: isvc-sklearn-predictor-00001
    ```
3. Called inference service from a default namespace using `default:default` SA Token:
    ```bash
    $ kubectl -n default run -ti --rm curl --image curlimages/curl --command -- sh
    ~$ curl istio-ingressgateway.istio-system/v1/models/isvc-sklearn:predict -H "Authorization: Bearer $(cat /run/secrets/kubernetes.io/serviceaccount/token)" -H "Host: isvc-sklearn.kserve-test.example.com" -H "Content-Type: application/json" -d '{"instances": [[6.8, 2.8, 4.8, 1.4], [6.0, 3.4, 4.5, 1.6]]}'
    {"predictions":[1,1]}
    ```

I'm confused by why this AuthorizationPolicy didnt work. But, looking at the documentation I noticed that the [kserve 0.11.0 docs with istio-dex sample](https://github.com/kserve/kserve/tree/v0.11.0/docs/samples/istio-dex) doesn't mention AuthorizationPolicy but [kserve 0.11.0 docs with istio-dex sample](https://github.com/kserve/kserve/tree/v0.12.0/docs/samples/istio-dex) does. Maybe it's only enabled from future version...? I'm not sure and I'd welcome some comments or ideas.

kromanow94:
I'll create tests for pipelines and notebooks access from different namespaces + notebooks api access with m2m tokens tomorrow.

Does it make sense to create tests for kserve given this comment? https://github.com/kubeflow/manifests/pull/2544#discussion_r1505133466

juliusvonkohout:
> I'll create tests for pipelines and notebooks access from different namespaces + notebooks api access with m2m tokens tomorrow.
> 
> Does it make sense to create tests for kserve given this comment? [#2544 (comment)](https://github.com/kubeflow/manifests/pull/2544#discussion_r1505133466)

Well listing notebooks is enough for now. You can skip kserve. You analysis is amazing, please add these comments rgarding pipelines, notebooks, kserve authentication flow into a readme in this PR. After tests for list notebooks lets merge and create follow up PRs.

juliusvonkohout:
In the follow up PRs i would also like to make you approver for dex and oauth2-proxy.

common/oidc-client/oauth2-proxy/components/istio-m2m/requestauthentication.yaml
rawc0der: @kromanow94 For the InferenceService AuthorizationPolicies, you shoudn't match against the predictor revision number `app: isvc-sklearn-predictor-00001`, this will be inconsistent with the actual request flow as there can be multiple revisions for this given InfService.
Better use some other generally available label on the Deployments like `component=predictor` which is added to all the Deployments managed by this InferenceService.

kromanow94: Hey @rawc0der , thanks for the suggestion! I'll experiment with it.

kromanow94:
From my perspective all that we've agreed on having in this PR is ready.

juliusvonkohout:
Amazing improvement
/approve

kromanow94:
Hurray!

Many thanks for all the support! ❤️🍀

deepakdeore2004:
which kubeflow version will have this change included?

kromanow94:
@deepakdeore2004 officially it will be included in the next release.

PR #2546, opened 2023-10-05
Add tests for rootless containers

juliusvonkohout:
We need to list all containers that do not have runasnonroot set or run as the root user. Based on this list i can create per workinggroup patches and at some point we can just enable the test in the e2e pipeline.

Adding tests to warn and later on error on root containers in the cicd.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -15,8 +15,6 @@ included in the tests.
 
 ### Test Suite
 
-**To use KFserving v0.5/0.6 instead of KServe v0.7, comment the 6,7th lines and uncomment the 10,11th lines here [e2e/utils/kserve.py](https://github.com/kubeflow/manifests/compare/master/tests/e2e/utils/kserve.py#L6-L11)**
-
 The e2e tests are completely independent of the underlying K8s cluster, as well
 as the platform of the cluster. These tests should be able to run in real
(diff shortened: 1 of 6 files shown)


juliusvonkohout:
@annajung @kimwnasptd please review.

annajung:
From quick look, the script checks to two specific fields in pod spec and container spec. In the description you mention needing to create a list of containers that do not have runasnonroot set, that doesn't seem to be part of the current script. Is that intended?

juliusvonkohout:
@annajung thanks for the review.

It is a bit more complicated. In the end we want to determine whether containers run as root or have been run as root (initcontainers)

Some containers are already completed (initcontainers) some are missing a shell or the Id command, some are missing a security context and or the runasnonroot within the securitycontext, some on the pod level and some on the container level.

In the end it is a diverse set of problems for which we need heterogeneous solutions. The script provides per container heuristics and negative hypothesis tests that let you easily gauge whether a container runs or ran as root.

For the moment it is useful for the development focus and creating securitycontext/runasnonroot PRs in each working group. In the long term we may fail the cicd if this script finds something potentially rootful.

juliusvonkohout:
@annajung can we move forward with a /lgtm or do you think it should be refined?

PR #2554, opened 2023-10-18
add profile controller proposal

OmriShiv:
Opening a proposal to discuss changes to the Profile Controller to support custom profile resources


juliusvonkohout:
/close

since this is the wrong repository and it is tracked in https://github.com/kubeflow/kubeflow/pull/7371

PR #2229, opened 2022-06-29
Use hashFromEnv in dex component for static user passwords

mvoitko:
**Which issue is resolved by this Pull Request:**
Resolves #

**Description of your changes:**

**Checklist:**
- [ ] Unit tests pass:
  **Make sure you have installed kustomize == 3.2.1**
    1. `make generate-changed-only`
    2. `make test`

PR #2588, opened 2024-01-02
Fix #2229 - dex-passwords in common/dex/base

kromanow94:
The dex-passwords secret has not been defined properly and the secret wasn't added to the dex kustomization file.

The changes are done as discussed in https://github.com/kubeflow/manifests/pull/2229#pullrequestreview-1799825164.

### DIFF
--- a/common/dex/base/dex-passwords.yaml
+++ b/common/dex/base/dex-passwords.yaml
@@ -2,5 +2,5 @@ apiVersion: v1
 kind: Secret
 metadata:
   name: dex-passwords
-data:
+stringData:
   DEX_USER_PASSWORD: $2y$12$4K/VkmDd1q1Orb3xAt82zu8gk7Ad6ReFR4LCP9UeYE90NLiN9Df72
--- a/common/dex/base/kustomization.yaml
+++ b/common/dex/base/kustomization.yaml
@@ -7,6 +7,7 @@ resources:
 - crds.yaml
 - deployment.yaml
 - service.yaml
+- dex-passwords.yaml
 generatorOptions:
   disableNameSuffixHash: true
 


juliusvonkohout:
Are you sure that the string is allowed like this? i am under the impression, that we have to encode it since it contains the dollar sign.

"You can define the Secret object in a manifest first, in JSON or YAML format, and then create that object. The [Secret](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.29/#secret-v1-core) resource contains two maps: data and stringData. The data field is used to store arbitrary data, encoded using base64. The stringData field is provided for convenience, and it allows you to provide the same data as unencoded strings. The keys of data and stringData must consist of alphanumeric characters, -, _ or .."

https://kubernetes.io/docs/tasks/configmap-secret/managing-secret-using-config-file/#create-the-config-file

kromanow94:
I was able to verify that this form works:
```
$ git checkout fix-pull-2229
$ git reset --hard
$ git log -1
commit 8aafa9a3041e2bec2441dd44df4854288ede6a04 (HEAD -> fix-pull-2229, origin/fix-pull-2229)
Author:     Krzysztof Romanowski <krzysztof.romanowski.kr3@roche.com>
AuthorDate: Tue Jan 2 13:50:19 2024 +0000
Commit:     Krzysztof Romanowski <krzysztof.romanowski.kr3@roche.com>
CommitDate: Tue Jan 2 13:50:19 2024 +0000

    fix #2229 - dex-passwords in common/dex/base

$ kustomize build common/dex/base/ | kubectl apply -f - --dry-run=server -oyaml | yq '.items[] | select(.metadata.name == "dex-passwords")| .data.DEX_USER_PASSWORD' -r | base64 -d; echo
$2y$12$4K/VkmDd1q1Orb3xAt82zu8gk7Ad6ReFR4LCP9UeYE90NLiN9Df72
```

But, I agree that even if this works, we should comply with the documentation because in future the K8s system can be more restrictive and more in-sync with the docs.

I've made a change to define the value encoded with base64.

kromanow94:
Ah, I see that this wasn't necessary... The docs talk about the `key` in `data` and `stringData`. It was the `value` that really had the dollar sign.

>The keys of data and stringData must consist of alphanumeric characters, `-`, `_` or `.`.

With that in mind, I'm more in favor of going with `stringData` to be more explicit. I'll make the change and bring it back.

PR #2617, opened 2024-02-11
Fix: Update docker image pull policy

### LINKED ISSUE #2308: Notebook-controller and Profile-and-kfam Docker Image Pull Policy
Hi,

We notice the below 3 docker images' `imagePullPolicy` are set to `Always `
* `docker.io/kubeflownotebookswg/notebook-controller:v1.6.1`
* `docker.io/kubeflownotebookswg/kfam:v1.6.1`
* `docker.io/kubeflownotebookswg/profile-controller:v1.6.1`

This could lead to hitting the [docker image pull rate limit](https://aws.amazon.com/premiumsupport/knowledge-center/ecs-pull-container-error-rate-limit/#:~:text=Docker%20Hub%20uses%20IP%20addresses,pulls%20per%206%2Dhour%20period)  which is set to **100 pull for anonymous users**  in **every 6 hours**.

We are wondering if the pullPolicy for these images is set to `Always` intentionally or it's a miss that can be set as `IfNotPresent` so they can be cached?

imVipul3000:
Resolves #2308

replace "Always" to "IfNotPresent" for imagePullPolicy so that it will only pull if image is not in local

**Checklist:**
- [x] Unit tests pass:
    2. `make test`

### DIFF
--- a/apps/kfp-tekton/upstream/base/cache-deployer/cache-deployer-deployment.yaml
+++ b/apps/kfp-tekton/upstream/base/cache-deployer/cache-deployer-deployment.yaml
@@ -19,7 +19,7 @@ spec:
       containers:
       - name: main
         image: gcr.io/ml-pipeline/cache-deployer:dummy
-        imagePullPolicy: Always
+        imagePullPolicy: IfNotPresent
         env:
         - name: NAMESPACE_TO_WATCH
           valueFrom:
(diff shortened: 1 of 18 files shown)


juliusvonkohout:
@imVipul3000 

Thanks for your PR. But there is a Problem with changing the OCI images here. The OCI images must be changed in the corresponding repositories such as Kubeflow/pipelines. As a side note we do not support Docker, only OCI images.

Feel free to join our manifests WG meeting for questions.

aqeelat (commented):
I suggest removing the imagePullPolicy altogether and relying on k8s defaults.

juliusvonkohout:
> I suggest removing the imagePullPolicy altogether and relying on k8s defaults.  

Also a valid option

juliusvonkohout:
/close

this is not the right repository, please use the upstream ones

PR #2629, opened 2024-02-27
Update OWNERS

juliusvonkohout:
Reduce to active maintainers, since most new contributors are confused if maintainers do not respond, but are listed in their PRs as approvers.

The latest commit from the removed approvers is 3 years old https://github.com/kubeflow/manifests/commit/7f8e6d83c735874c34bb53ec9cbc105f41b8b71e

CC @andreyvelich and KSC

CC @kubeflow/wg-manifests-leads @kimwnasptd @rimolive 

Here is an example of the problem:

grafik

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,10 +1,6 @@
 approvers:
-  - elikatsis
   - juliusvonkohout
   - kimwnasptd
-  - PatrickXYS
-  - StefanoFioravanzo
-  - yanniszark
 reviewers:
   - juliusvonkohout
   - kimwnasptd


juliusvonkohout:
@elikatsis
@PatrickXYS
@StefanoFioravanzo
@yanniszark

Since you did not contribute in the last 3 years AFAIK, are you still active or do you want to become active again in the near future in this repository?

juliusvonkohout:
@jbottum @johnugeorge @terrytangyuan what do you think?

juliusvonkohout:
i am also looking at getting maybe @kromanow94 and other active contributors as subproject maintainers in the long term. I think whoever contributes in some way consistently is a good candidate.

johnugeorge:
LGTM @juliusvonkohout

elikatsis:
Hello, thanks for tackling this.

In kubeflow/kubeflow we've introduced the concept of emeritus approvers (which is a common practice). Would it make sense to do it here as well? See kubeflow/kubeflow#7402

juliusvonkohout:
> Hello, thanks for tackling this.
> 
> In kubeflow/kubeflow we've introduced the concept of emeritus approvers (which is a common practice). Would it make sense to do it here as well? See [kubeflow/kubeflow#7402](https://github.com/kubeflow/kubeflow/pull/7402)

Done in https://github.com/kubeflow/manifests/pull/2630

PR #2627, opened 2024-02-26
Update kserve manifests from v0.11.2

rimolive:
Update kserve/kserve manifests from v0.11.2

**Which issue is resolved by this Pull Request:**
Resolves #

**Description of your changes:**

**Checklist:**
- [ ] Unit tests pass:
  **Make sure you have installed kustomize == 5.0.3**
    1. `make generate-changed-only`
    2. `make test`

PR #2567, opened 2023-11-01
Update kubeflow/kubeflow manifests from v1.8.0

DnPlas:
GA release of kubeflow/kubeflow 1.8.0

cc: @kimwnasptd @thesuperzapper

PR #2633, opened 2024-02-29
Update kserve manifests from v0.11.2 (#2627) and synchronize from the 1.8 branch

juliusvonkohout:
@rimolive 

cherry pick Update kubeflow/kubeflow manifests from v1.8.0 (#2567)


juliusvonkohout:
Maybe the CLA check is messed up. Maybe just transform this into  a patch and create a fresh PR from that.

juliusvonkohout:
We should drop

```
# If you want to install both KFServing and KServe, install the
# standalone kserve manifests, to avoid conflicts with 0.6.1 KFServing.
# - kserve.yaml
# - aggregated-roles.yaml
# - kserve-runtimes.yaml
# configMapGenerator:
#   - name: kserve-config
#     namespace: kserve
#     behavior: merge
#     envs:
#     - params.env
```

from the main kustomization.yaml 

https://github.com/kubeflow/manifests/blob/7c2de6bbfd486919b36d519f08393399758c30c8/tests/gh-actions/install_kserve.sh should be rewritten to clearly separate the CRD installation and wait after that.

rimolive:
Is this required for 1.8.1?

juliusvonkohout:
> Is this required for 1.8.1?

no for 1.9. The merge just started the wrong way. we should always update master first and backport to older branches as for example 1.8. So this time i have to make a forward-port.

PR #2668, opened 2024-04-05
Reorder oauth2-proxy manifests so that namespace is first

codablock:
This ensures that the namespace is always deployed before anything else, even if fifo ordering is used with Kustomize.

**Checklist:**
- [ ] Unit tests pass:
    2. `make test`

### DIFF
--- a/common/oidc-client/oauth2-proxy/base/kustomization.yaml
+++ b/common/oidc-client/oauth2-proxy/base/kustomization.yaml
@@ -4,8 +4,8 @@ kind: Kustomization
 namespace: oauth2-proxy
 
 resources:
-- deployment.yaml
 - namespace.yaml
+- deployment.yaml
 - oauth2-proxy-config.yaml
 - serviceaccount.yaml
 - service.yaml


juliusvonkohout:
Sorry, our CICD is currently too slow for some tests.

codablock:
@juliusvonkohout Are you sure that slowliness is the cause of the failure? I see Python import errors in the logs...

If I remember correct, the individual GH workflows are only run on source changes in specific folders, and also only on PRs. This can easily lead to non-working workflows that get unnoticed for quite some time. If you want, I can try to investigate when I find time?

juliusvonkohout:
> @juliusvonkohout Are you sure that slowliness is the cause of the failure? I see Python import errors in the logs...
> 
> If I remember correct, the individual GH workflows are only run on source changes in specific folders, and also only on PRs. This can easily lead to non-working workflows that get unnoticed for quite some time. If you want, I can try to investigate when I find time?

Yes, @kromanov is investigating as well. In other PRs it is failing for different reasons. But rebasing your PR to master and checking test_sklearn.py is a good idea.

juliusvonkohout:
@codablock please rebase to master since https://github.com/kubeflow/manifests/actions/runs/8897758758/job/24433376331?pr=2698 was succesful

PR #2672, opened 2024-04-08
Synchronize katib and kubeflow/kubeflow

juliusvonkohout:
@rimolive @diegolovison 

maybe there are even more components that have been merged in the wrong order and left the master outdated.

### DIFF
--- a/apps/katib/upstream/components/controller/trial-templates.yaml
+++ b/apps/katib/upstream/components/controller/trial-templates.yaml
@@ -15,7 +15,7 @@ data:
         spec:
           containers:
             - name: training-container
-              image: docker.io/kubeflowkatib/mxnet-mnist:v0.16.0-rc.1
+              image: docker.io/kubeflowkatib/mxnet-mnist:v0.16.0
               command:
                 - "python3"
                 - "/opt/mxnet-mnist/mnist.py"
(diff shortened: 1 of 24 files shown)


apps/katib/upstream/components/controller/trial-templates.yaml
```diff
@@ -33,7 +33,7 @@ data:
         spec:
           containers:
             - name: training-container
-              image: docker.io/kubeflowkatib/enas-cnn-cifar10-cpu:v0.16.0-rc.1
+              image: docker.io/kubeflowkatib/enas-cnn-cifar10-cpu:v0.16.0
```
rimolive, line 36: Looks like this image tag does not exist.

cc @andreyvelich

juliusvonkohout: I am just merging what we have in 1.8.

andreyvelich: Why this tag doesn't exist ?
Here it is: https://hub.docker.com/layers/kubeflowkatib/enas-cnn-cifar10-cpu/v0.16.0/images/sha256-5c27e2c36638d0764619d99c6de442ee54a8991c6bc5e0c8e5dacb9a75ddd9ed?context=repo

rimolive: Hmm, I tried search for this tag in Docker Hub but can't find it. Thanks for the confirmation!

PR #2675, opened 2024-04-08
Add kuberay aggregated-roles

### LINKED ISSUE #2671: Kubeflow roles do not have the ability to manage ray.io resources
As Kuberay is included in the contrib directory, it may make sense to given users permissions when installing it with resources like this:

```yaml
kind: ClusterRole
metadata:
  name: kubeflow-ray-admin
  labels:
    rbac.authorization.kubeflow.org/aggregate-to-kubeflow-admin: 'true'
rules:
- apiGroups:
  - ray.io
  verbs: ["*"]
  resources: ["*"]
---
kind: ClusterRole
metadata:
  name: kubeflow-ray-editor
  labels:
    rbac.authorization.kubeflow.org/aggregate-to-kubeflow-edit: 'true'
rules:
- apiGroups:
  - ray.io
  verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
  resources: ["*"]
---
kind: ClusterRole
metadata:
  name: kubeflow-ray-viewer
  labels:
    rbac.authorization.kubeflow.org/aggregate-to-kubeflow-view: 'true'
rules:
- apiGroups:
  - ray.io
  verbs: ["get", "list", "watch"]
  resources: ["*"]
---
```

Happy to make a PR if this seems right.

alex-treebeard:
Resolves #2671

Adds aggregated roles as discussed in https://github.com/kubeflow/manifests/issues/2671

**Checklist:**
- [ ] Unit tests pass:
    2. `make test`

@juliusvonkohout 

Thanks for the guidance on the issue.

Ray + Istio warrants an offline chat IMO. I've also tried it myself and am unclear about the UX/security implications of those three levels of istio integration. I'm also conscious that if we switch to ambient mesh that will change this also (as Ray cluster's sidecar would fail if it ran before the istio sidecar).

Please let me know if any naming conventions etc need to be followed in the diff - this is ready for review.


juliusvonkohout (changes requested):
Please try to structure it like https://github.com/kubeflow/manifests/blob/master/apps/katib/upstream/installs/katib-with-kubeflow/kubeflow-katib-roles.yaml

alex-treebeard:
sure, i've
- indented lists (and used yaml not json syntax)
- ordered object keys
- added leading '---'

just let me know if you meant any semantic changes

juliusvonkohout:
/lgtm
/approve

thank you for your contribution.

juliusvonkohout (approved):

PR #2693, opened 2024-04-29
Refactor test workflows 

lampajr:

Upgrade `actions/checkout` to version `v4`.
This is going to be required as `v3` does not run on `node20` which will be required in a while.

[edit] As discussed, this PR also contains some refactoring on the gh workflows, e.g., renaming and dependencies

**Checklist:**
- [ ] Unit tests pass:
    2. `make test`


lampajr:
Here the link to the Github statement https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/

juliusvonkohout:
* consistent nameing e.g. vwa_kind_test.yaml -> volumes-web-application.yaml
* consistent and complete paths for triggering the workflow
*

juliusvonkohout:
@lampajr can you fix the bentoml tests?

lampajr:
> @lampajr can you fix the bentoml tests?

I can try to take a look but looking at the previous runs of the test [1], it looks like it never worked :thinking: 

Alternatively, we could create a different issue to fix bentoml test, as I think it is not directly related to this PR, wdyt?

[1] https://github.com/kubeflow/manifests/actions/workflows/bentoml_kind_test.yaml

lampajr:
> Yes, please update the bentoml stuff, because we have to pass all tests for now. If it is a small change even in this PR here.

I created a separate PR as it looks like the is an error while running the `test.sh` script inside `contrib/bentoml`. See https://github.com/kubeflow/manifests/pull/2704

lampajr:
Rebased after https://github.com/kubeflow/manifests/pull/2704 merge

PR #2695, opened 2024-04-30
Sync kubeflow pipelines manifests 2.1.0

juliusvonkohout:
@rimolive


juliusvonkohout:
I think Deploy and test Kubeflow Pipelines manifests with m2m auth in KinD / build (pull_request) supersedes 
Build & Apply Kubeflow Pipelines manifests in KinD / build (pull_request). Or should we fix it?

rimolive:
We must wait for Kubeflow Pipelines 2.2.0 to be released. This is the version it will be available for Kubeflow 1.9

juliusvonkohout:
> We must wait for Kubeflow Pipelines 2.2.0 to be released. This is the version it will be available for Kubeflow 1.9

I still would like to get this release from March 25  (so only last month) in and update the synchronization script in this PR.

If Google adds a new release i can do the update as well. This way we also get earlier testing exposure in RC0

So please /LGTM if there is nothing problematic

PR #2698, opened 2024-04-30
Update kubeflow/katib manifests from V0.17.0 rc.0

juliusvonkohout:
@rimolive


juliusvonkohout:
I am also still improving the scripts to be fully automatic and idempotent.

PR #2699, opened 2024-05-01
Update kubeflow/training-operator manifests from v1.8.0-rc.0

rimolive:
Resolves #

**Checklist:**
- [x] Unit tests pass:
    2. `make test`


juliusvonkohout:
Yes, I think we should streamline the synchronization script here as well. Feel also free to change it as long as you change all scripts accordingly.

rimolive:
@juliusvonkohout I updated the sync script. Would you mind take a look?

juliusvonkohout:
i did not check the added rbac roles, but that is something for the training operator folks

For synchronizing and rc0
/lgtm
/approve

PR #2702, opened 2024-05-03
Fix kserve upgrade script and update kserve diagram

sivanantha321:
Resolves #
Kserve recently renamed the release artifact from kserve-runtimes.yaml to kserve-cluster-resources.yaml. However, the Makefile still references the old one. This PR resolves this issue.

This PR also updates the kserve diagram.

**Checklist:**
- [ ] Unit tests pass:
    2. `make test`


juliusvonkohout:
Hhm on a first glimpse it looks as if there are synchronisation scripts in two places. @rimolive we probably should take a closer look

juliusvonkohout:
It is definitely not perfect, but at least an improvement. Ill remove the makefile in another PR since we have the synchronization script.

/lgtm
/approve

PR #2724, opened 2024-05-20
add Model Registry networkpolicy

### LINKED ISSUE #2701: Add Kubeflow Model Registry Network Policies
As discussed in the last Manifests WG (2nd May, 2024) we need to setup Network Policies for the new Kubeflow Model Registry, this seems required to make the pod accessible from namespaces other than the `kubeflow` one.

The new network policy should be added here: https://github.com/kubeflow/manifests/blob/master/common/networkpolicies/base/

A good starting point, as suggested in the meeting is: https://github.com/kubeflow/manifests/blob/master/common/networkpolicies/base/ml-pipeline.yaml

**Open question:**

Do we need to apply network policy for both ports (`8080`, the REST interface and `9090`, the gRPC interface) ?

> NOTE: This is a non-blocker for Kubeflow 1.9 release as KFMR won't be deployed by default

tarilabs:
Raising as draft following indication on #2701 but not yet sure how to manually verify: I've tested with `1.9.0-rc.0` on a local cluster, both with _and_ without this network policy, I can reach the service both externally the cluster (authenticated call) and from another pod in a different namespace (namespace "default" using wget from an hello-world pod).

Resolves #2701 

**Checklist:**
- [ ] Unit tests pass:
    2. `make test`

### DIFF
--- a/common/networkpolicies/base/model-registry.yaml
+++ b/common/networkpolicies/base/model-registry.yaml
@@ -24,7 +24,6 @@ spec:
                 operator: In
                 values:
                   - istio-system
-        - podSelector: {} # allow all pods from the same namespace
       ports:
         - protocol: TCP
           port: 8080
--- a/common/networkpolicies/base/model-registry.yaml
+++ b/common/networkpolicies/base/model-registry.yaml
@@ -25,5 +25,10 @@ spec:
                 values:
                   - istio-system
         - podSelector: {} # allow all pods from the same namespace
+      ports:
+        - protocol: TCP
+          port: 8080
+        - protocol: TCP
+          port: 9090
   policyTypes:
     - Ingress
--- a/common/networkpolicies/base/model-registry.yaml
+++ b/common/networkpolicies/base/model-registry.yaml
@@ -1,0 +1,29 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: model-registry
+  namespace: kubeflow
+spec:
+  podSelector:
+    matchExpressions:
+      - key: component
+        operator: In
+        values:
+          - model-registry-server
+  ingress:
+    - from:
+        - namespaceSelector:
+            matchExpressions:
+              - key: app.kubernetes.io/part-of
+                operator: In
+                values:
+                  - kubeflow-profile
+        - namespaceSelector:
+            matchExpressions:
+              - key: kubernetes.io/metadata.name
+                operator: In
+                values:
+                  - istio-system
+        - podSelector: {} # allow all pods from the same namespace
+  policyTypes:
+    - Ingress
--- a/common/networkpolicies/base/kustomization.yaml
+++ b/common/networkpolicies/base/kustomization.yaml
@@ -16,6 +16,7 @@ resources:
   - minio.yaml
   - ml-pipeline-ui.yaml
   - ml-pipeline.yaml
+  - model-registry.yaml
   - poddefaults.yaml
   - pvcviewer-webhook.yaml
   - seldon.yaml


dhirajsb:
lgtm, but need to test it manually in a deployment

juliusvonkohout:
@lampajr please add explicit port numbers for the service.

lampajr:
> @lampajr please add explicit port numbers for the service.

Hi @juliusvonkohout, given that we should make accessible all ports that are exposed by the model registry pod (@tarilabs  can you confirm that?) I think that it is acceptable as it is (without any port specification, similarly to what other components are already doing)

tarilabs:
I would rather have this pov: this networkpolicy proposal aligns with other examples as suggested by action item out of previous meeting (notes).

In short we need 2 "endpoints": REST API (8080) and gRPC (9090).

Could someone kindly comment regardings how to A/B test it, per original message, please?
That is not clear to me, and sorry if asking a banal question

juliusvonkohout:
> I would rather have this pov: this networkpolicy proposal aligns with other examples as suggested by action item out of previous meeting (notes).
> 
> In short we need 2 "endpoints": REST API (8080) and gRPC (9090).
> 
> Could someone kindly comment regardings how to A/B test it, per original message, please?
> That is not clear to me, and sorry if asking a banal question

It is rather trivial and best practices to list those two ports in the networkpolicy. A networkpolicy is just a firewall rule. If you install a full Kubeflow and can still reach both ports it is fine. For standalone test I would add a deny all networkpolicy and this policy here on top to test.

tarilabs:
I've added https://github.com/kubeflow/manifests/pull/2724/commits/31345c974f814d0af9c62b51cb46d73c750b50a8
but I need to test it on a vanilla KF cluster, hence keeping this as draft for now.
Would this align better to expectations, please?

common/networkpolicies/base/model-registry.yaml
```diff
@@ -0,0 +1,34 @@
...
+metadata:
+  name: model-registry
+  namespace: kubeflow
+spec:
+  podSelector:
+    matchExpressions:
+      - key: component
+        operator: In
+        values:
+          - model-registry-server
```
juliusvonkohout, line 12: Are you sure that this targets exactly/only the pods which serve on those ports? It should not target any model-registry pods that do not serve on those ports.

common/networkpolicies/base/model-registry.yaml
```diff
@@ -0,0 +1,34 @@
...
+                operator: In
+                values:
+                  - kubeflow-profile
+        - namespaceSelector:
+            matchExpressions:
+              - key: kubernetes.io/metadata.name
+                operator: In
+                values:
+                  - istio-system
+        - podSelector: {} # allow all pods from the same namespace
```
juliusvonkohout, line 27: I think this one here is already the default networkpolicy allow-same-namespace or so. It does not belong into the one for the model registry.

juliusvonkohout:
Yes next to my  2 comments I would like to add that gpt4 can write you such policies. Just input your service and deployments and it will figure it out.

common/networkpolicies/base/model-registry.yaml
tarilabs: Yes, this can be verified by showing that for the MR manifests, only MR is targeted, the other Deployments part of a MR installation do not have this selector, as also shown in this screenshot:

image

tarilabs: I would consider this point resolved accordingly.

tarilabs:
Tested with:

```diff
diff --git a/common/networkpolicies/base/kustomization.yaml b/common/networkpolicies/base/kustomization.yaml
index 3592bc9a..33bf626c 100644
--- a/common/networkpolicies/base/kustomization.yaml
+++ b/common/networkpolicies/base/kustomization.yaml
@@ -16,6 +16,7 @@ resources:
   - minio.yaml
   - ml-pipeline-ui.yaml
   - ml-pipeline.yaml
+  - model-registry.yaml
   - poddefaults.yaml
   - pvcviewer-webhook.yaml
   - seldon.yaml
diff --git a/common/networkpolicies/base/model-registry.yaml b/common/networkpolicies/base/model-registry.yaml
new file mode 100644
index 00000000..801a2145
--- /dev/null
+++ b/common/networkpolicies/base/model-registry.yaml
@@ -0,0 +1,33 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: model-registry
+  namespace: kubeflow
+spec:
+  podSelector:
+    matchExpressions:
+      - key: component
+        operator: In
+        values:
+          - model-registry-server
+  ingress:
+    - from:
+        - namespaceSelector:
+            matchExpressions:
+              - key: app.kubernetes.io/part-of
+                operator: In
+                values:
+                  - kubeflow-profile
+        - namespaceSelector:
+            matchExpressions:
+              - key: kubernetes.io/metadata.name
+                operator: In
+                values:
+                  - istio-system
+      ports:
+        - protocol: TCP
+          port: 8080
+        - protocol: TCP
+          port: 9090
+  policyTypes:
+    - Ingress
```

per latest reviews.

Without network policy:

Screenshot from 2024-06-03 14-58-26

WITH network policy:

Screenshot from 2024-06-03 15-02-20

## Notes on testing environment used

Since I've tested locally on Minikube, I needed to ensure "calico".
For example:

```
minikube start --cni calico --memory 8192 --cpus 6 # ...
```

> [!CAUTION]
> the `--cni calico` need be supplied at the FIRST start, it cannot be applied to an already-existing minikube

outputs as (on my linux box)

```
$ minikube start --listen-address=0.0.0.0 --cni calico --memory 8192 --cpus 6
😄  minikube v1.33.1 on Fedora 40
✨  Automatically selected the docker driver. Other choices: qemu2, ssh
📌  Using Docker driver with root privileges

...

🔗  Configuring Calico (Container Networking Interface) ...
🔎  Verifying Kubernetes components...
    ▪ Using image gcr.io/k8s-minikube/storage-provisioner:v5
🌟  Enabled addons: storage-provisioner, default-storageclass
🏄  Done! kubectl is now configured to use "minikube" cluster and "default" namespace by default

$ kubectl port-forward svc/istio-ingressgateway -n istio-system 8080:80
Forwarding from 127.0.0.1:8080 -> 8080
Forwarding from [::1]:8080 -> 8080
...
```

(notice the calico line explicitly output)

## Additional notes
This might actually be analogous as experienced by end-user with:
- https://github.com/kubeflow/model-registry/issues/90#issuecomment-2143852195

common/networkpolicies/base/model-registry.yaml
tarilabs: since this line is removed as requested with https://github.com/kubeflow/manifests/commit/5871263a7fe2f3c8403f9f2ca914bba874bcb752 resolving this thread.

tarilabs:
Thank you @juliusvonkohout 🙏 🚀

tiansiyuan:
@tarilabs I tested it on Microk8s 1.28. It works.

PR #2752, opened 2024-06-11
fix: update /models/ to /kserve-endpoints/ for kserve models web app …

OussemaLouati:
Fix KServe model endpoints webapp Url.

> Using `/models/` will result in the following screen when deploying Kubeflow

models-invalid

  

> After Updating the **_ConfigMap_** to use  `/kserve-endpoints/`, the Models endpoint will show correctly: 

kserve-fixed

> Note: Instead of updating the ConfigMap [`centraldashboard-config`](https://github.com/kubeflow/manifests/blob/master/apps/centraldashboard/upstream/overlays/kserve/patches/configmap.yaml#L36)  we can update the actual [VirtualService](https://github.com/kubeflow/manifests/blob/master/contrib/kserve/models-web-app/base/istio.yaml#L14) to use `/models/` instead also.

### DIFF
--- a/apps/centraldashboard/upstream/overlays/kserve/patches/configmap.yaml
+++ b/apps/centraldashboard/upstream/overlays/kserve/patches/configmap.yaml
@@ -33,7 +33,7 @@ data:
             },
             {
                 "type": "item",
-                "link": "/models/",
+                "link": "/kserve-endpoints/",
                 "text": "KServe Endpoints",
                 "icon": "kubeflow:models"
             },


juliusvonkohout:
It seems to have been changed in https://github.com/kubeflow/manifests/commit/fb440d82c6996f5e39ea153b76319acfdf1aa805

@diegolovison  can you test whether it works in the current master branch ?

CC @rimolive

diegolovison:
> @diegolovison can you test whether it works in the current master branch ?

yes, I can but I have a question. Are we going to backport to 1.9 as well?

juliusvonkohout:
> > @diegolovison can you test whether it works in the current master branch ?
> 
> yes, I can but I have a question. Are we going to backport to 1.9 as well?

we have to get this into the next 1.9 RC for kubeflow/kubeflow.

rimolive:
/lgtm

I tested and it worked.

juliusvonkohout:
/approve

this is not the proper fix. It has to be fixed in kubeflow/kubeflow and it will hit us again on the next kubeflow/kubeflow synchronization. But for now we need to be able to release the next RC.

juliusvonkohout:
There is a PR to fix it https://github.com/kubeflow/kubeflow/pull/7603

PR #2778, opened 2024-06-29
PSS labels for the profile controller

biswajit-9776:

- Please include a summary of changes and the related issue. 
- List any dependencies that are required for this change. 
- Please delete the options that are not relevant.
- The following checklist will help you to satisfy the requirements.

  
## ✅ Unit Test Checklist
  
  - [ ] 🛠️ Make sure you have installed kustomize == 5.2.1+     
  - [ ] ✍️ Have you written new tests for your core changes, as applicable?      
  - [ ] 🔄 Have you successfully run existing tests with your changes ?    
  - [ ] 🚀 Have you successfully run existing and new tests with your changes ?

## ✅ Contributor checklist
  - [ ] All the commits have been _signed-off_  (To pass the `DCO` check)
  - [ ] Submit the [Contributor License Agreements](https://cla.developers.google.com/clas) (To pass the `cla/google` check)

---     
 

juliusvonkohout:
@biswajit-9776 where is profiles-system coming from? The patch should touch only [the ](https://github.com/kubeflow/manifests/blob/master/apps/profiles/upstream/base/namespace-labels.yaml) in the Kubeflow namespace at build time and you need to keep the existing labels https://github.com/kubeflow/manifests/blob/60348d70f1614d1661d365ae868c844f7bca5e3d/apps/profiles/upstream/base/namespace-labels.yaml#L15C1-L18C46 .

When the cluster is running and new namespaces are gnerated dynamically, the labels from this file will be applied to each.

juliusvonkohout:
No, it works differently. Please check out https://github.com/kubeflow/manifests/blob/60348d70f1614d1661d365ae868c844f7bca5e3d/apps/profiles/upstream/base/kustomization.yaml#L17C1-L20C26 to understand how the configuration map is created for the profile controller. You have to patch the config map for the dynamic stuff. At runtime the profile controller will use this configmap to create namespace of yet unknown names. The static patch for the kubeflow namespace is something completely different.

biswajit-9776:
Okay I will push a commit later to this PR

biswajit-9776:
As per the files I went through, does https://github.com/kubeflow/manifests/blob/60348d70f1614d1661d365ae868c844f7bca5e3d/apps/profiles/upstream/base/namespace-labels.yaml#L15C1-L18C46 apply as labels to the dynamic namespaces created? Or these labels are just present as env variables in the configmap mounted to the /etc/profile-controller path of those containers in dynamic namespaces?

juliusvonkohout:
Always consider that gpt4 or Gemini can guide you a bit on such general ideas ;-)

Yes just try it and the profile controller and therefore the configmap is in the kubeflow namespace.

contrib/security/PSS/dynamic/restricted/namespace-labels.yaml
```diff
@@ -0,0 +1,19 @@
...
+# To remove a namespace label, use `key: ''`. For example:
+# istio-injection: ''
+#
+# Profile controller will not replace a namespace label if its key already
+# exists. If you want to override the value of a previously applied label, you
+# need to:
+# 1. Remove the label by using `key: ''` and deploy.
+# 2. Add the label by using `key: 'value'` and deploy.
+#
+katib.kubeflow.org/metrics-collector-injection: "enabled"
```
juliusvonkohout, line 15: Only mention what you merge into the existing one, so drop 

```
katib.kubeflow.org/metrics-collector-injection: "enabled"
serving.kubeflow.org/inferenceservice: "enabled"
pipelines.kubeflow.org/enabled: "true"
Comment
app.kubernetes.io/part-of: "kubeflow-profile"
```

contrib/security/PSS/dynamic/baseline/kustomization.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: kustomize.config.k8s.io/v1alpha1
+kind: Component
+
+resources:
+- ../../../../../apps/profiles/upstream/base
+
+configMapGenerator:
+- name: namespace-labels-data
+  namespace: kubeflow
+  behavior: replace
```
juliusvonkohout, line 10: We want to merge, not replace

contrib/security/PSS/dynamic/baseline/kustomization.yaml
```diff
@@ -0,0 +1,14 @@
...
+  files:
+  - namespace-labels.yaml
+  # options:
```
juliusvonkohout, line 13: # options:
  #   disableNameSuffixHash: true

this must be set on the profile controller base level

example/kustomization.yaml
```diff
@@ -94,4 +94,8 @@ resources:
 # Uncomment to enable baseline level standards
 # - ../contrib/security/PSS/static/baseline
 # Uncomment to enable restricted level standards
-# - ../contrib/security/PSS/static/restricted
\ No newline at end of file
+# - ../contrib/security/PSS/static/restricted
```
juliusvonkohout, line 97: You need to add this under a component section at the bottom 

```
kind: Kustomization
resources:
  - ...
components:
  - ...
```

biswajit-9776:
So, we have to use the below change in the base kustomization:
```
options:
  disableNameSuffixHash: true
```
I tried without it and as a result the configMap instead of getting overridden, still exists along with a new configMap with our changes.

biswajit-9776:
After making the above change, I tried commenting the unrequired section of _namespacea-labels.yam_ and after applying our component, I discovered this:
```
Name:         namespace-labels-data
Namespace:    kubeflow
Labels:       kustomize.component=profiles
Annotations:  <none>

Data
====
namespace-labels.yaml:
----

# # Below is a list of labels to be set by default.
# #
# # To add a namespace label, use `key: 'value'`, for example:
# # istio.io/rev: 'asm-191-1'
# #
# # To remove a namespace label, use `key: ''`. For example:
# # istio-injection: ''
# #
# # Profile controller will not replace a namespace label if its key already
# # exists. If you want to override the value of a previously applied label, you
# # need to:
# # 1. Remove the label by using `key: ''` and deploy.
# # 2. Add the label by using `key: 'value'` and deploy.

# katib.kubeflow.org/metrics-collector-injection: "enabled"
# serving.kubeflow.org/inferenceservice: "enabled"
# pipelines.kubeflow.org/enabled: "true"
# app.kubernetes.io/part-of: "kubeflow-profile"
pod-security.kubernetes.io/enforce: "baseline"

```
what I believe is that instead of adding the uncommented stuff, it replaced the everything else with the new stuff we tried to add. To deal with this temporarily, I wish to have an exact copy of base _namespace-label.yaml_ in my component's directory along with PSS labels.

apps/profiles/upstream/base/kustomization.yaml
```diff
@@ -16,5 +16,7 @@ images:
 
 configMapGenerator:
 - name: namespace-labels-data
+  options:
```
juliusvonkohout, line 19: You are not allowed to change stuff in the upstream folder. Please try without it and you will find a way.

juliusvonkohout:
> After making the above change, I tried commenting the unrequired section of _namespacea-labels.yam_ and after applying our component, I discovered this:
> ```
> Name:         namespace-labels-data
> Namespace:    kubeflow
> Labels:       kustomize.component=profiles
> Annotations:  <none>
> 
> Data
> ====
> namespace-labels.yaml:
> ----
> 
> # # Below is a list of labels to be set by default.
> # #
> # # To add a namespace label, use `key: 'value'`, for example:
> # # istio.io/rev: 'asm-191-1'
> # #
> # # To remove a namespace label, use `key: ''`. For example:
> # # istio-injection: ''
> # #
> # # Profile controller will not replace a namespace label if its key already
> # # exists. If you want to override the value of a previously applied label, you
> # # need to:
> # # 1. Remove the label by using `key: ''` and deploy.
> # # 2. Add the label by using `key: 'value'` and deploy.
> 
> # katib.kubeflow.org/metrics-collector-injection: "enabled"
> # serving.kubeflow.org/inferenceservice: "enabled"
> # pipelines.kubeflow.org/enabled: "true"
> # app.kubernetes.io/part-of: "kubeflow-profile"
> pod-security.kubernetes.io/enforce: "baseline"
> 
> ```
> what I believe is that instead of adding the uncommented stuff, it replaced the everything else with the new stuff we tried to add. To deal with this temporarily, I wish to have an exact copy of base _namespace-label.yaml_ in my component's directory along with PSS labels.

Please investigate the officially documented behavior of merge in the configmapgenerator. We need to be sure.

juliusvonkohout (changes requested):

apps/profiles/upstream/base/kustomization.yaml
biswajit-9776: The documentation does day that _merge_ behavior should add or update values to the configMap.
`https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/configmapgenerator/#overriding-base-configmap-values:~:text=merge%3A%20add%20or%20update%20the%20values%20in%20an%20existing%20ConfigMap%20from%20the%20base.`

I also found an issue here at `https://github.com/kubernetes-sigs/kustomize/issues/370#:~:text=%40mr%2Dkaran%20%40anshul,here%20as%20well.` which also like us, wishes the configMap to get configured with merge behavior. But, I guess that's why I see the same in cluster while trying to merge the configMap with new labels. There's a new configMap that gets created with a new suffix to which the workloads point to (as they should) but it doesn't include the values of previous configMap due to the hash suffix.

contrib/security/PSS/dynamic/baseline/namespace-labels.yaml
```diff
@@ -0,0 +1,19 @@
+# Below is a list of labels to be set by default.
```
juliusvonkohout, line 1: write a compent where theis comes from and why we have to duplicate it.

juliusvonkohout (approved):

PR #2780, opened 2024-06-30
Update kubeflow/kubeflow manifests from v1.9.0-rc.1

juliusvonkohout:
@rimolive for /lgtm


juliusvonkohout:
/hold

  newTag: latest must be reveiwed. Mayeb it is fine because it is overwritten at a higher level.

PR #2781, opened 2024-06-30
Update kubeflow/kubeflow manifests from v1.9.0-rc.1

juliusvonkohout:
@rimolive for /lgtm


juliusvonkohout:
I think we can adjust the test to just create an experiment etc. But leave out the pipeline execution itself. That should be enough as an authentication test for now.

PR #2785, opened 2024-07-04
Document the CVE scanning process

StefanoFioravanzo:
This PR adds a new MD file to document the CVE scanning process. 

@juliusvonkohout this is how I think we can provide better clarity to users. Please fact check what I wrote and propose improvements.

As for the data and disclose policy, I kind of made that up. I don't know what you are following. I do know that @akgraner had a doc, but are you guys following any process right now, or is it just dumping the CVE results in the GH action logs and that's it?


juliusvonkohout:
"or is it just dumping the CVE results in the GH action logs and that's it?" yes

Please just add a comment directly in the workflow file or trivy python script  or a section in the readme. We want to get rid of /documents soon.

StefanoFioravanzo:
@juliusvonkohout The main README is fairly long already. Why do you want to get rid of documents? Are you ok with having a separate file (CVE_SCANNING.md) in the root of the repo, instead?

juliusvonkohout:
This is probably dead documentation soon then. What we can do is in the section you created in the main Readme write. "We do CVE image scanning with trivy on each merge to the master branch. Here is the workflow file and here is the python script for further information. The GitHub action run outputs detailed information per image and a summary table.You can also run this locally."

documents/CVE_Scanning.md
```diff
@@ -38,6 +38,10 @@ The CVE scanning process runs on each merge to the master branch of the Kubeflow
 ### Results Storage
 The results of each scan are stored as part of the GitHub Actions job. The results can be viewed at the bottom of the job run output, such as in this example: [GitHub Actions Run](https://github.com/kubeflow/manifests/actions/runs/9744982451/job/26891834585).
 
+To view all past security scans, head to the [Actions](https://github.com/kubeflow/manifests/actions), then expand the workflows in the left sidebar and click on the [Image Extracting and Security Scanning](https://github.com/kubeflow/manifests/actions/workflows/trivy.yaml) workflow. Select one of the actions in the main list, then click on the main pipeline action to access the logs of the action, and expand the `Run image extracting and security scanning script` step to view the CVE logs. You will find a per-image CVE scan and a JSON dump of per-WorkingGroup aggregate metrics.
```
juliusvonkohout, line 41: You can link directly to https://github.com/kubeflow/manifests/actions/workflows/trivy.yaml

documents/CVE_Scanning.md
```diff
@@ -38,6 +38,10 @@ The CVE scanning process runs on each merge to the master branch of the Kubeflow
...
+
+The script also saves all these results locally for convenience. The folder locations and filenames are exposed at the end of the logs. You can run the script locally in your machine to these files.
```
juliusvonkohout, line 43: You can run the script locally on your machine to obtain the detailed JSON files.

juliusvonkohout:
We can add more stuff in follow up PRs and i think it is a summary table not a json file that is seen in the logs.

chasecadet:
I am trying to better track  this all. What @StefanoFioravanzo what is the concern and what is the outcome? 
I also need to know what @juliusvonkohout and team's vision is here and how that maps to our users success.  I know Mr.Kohout is very diligent for his engineering requirements but want to better understand the demand and how we talk about these workflows with users so the work is mapped to value (even if that means just raising awareness). Otherwise we might be wasting precious cycles.  Maybe worth discussing on a community call.

juliusvonkohout:
I think it is mentioned in the release notes and the Blogpost and now in the main Readme. 

Just saying again what is in the workflow file and the python script does not make much sense, because everyone can run this through GPT4 and it might change in the future as well.  Maybe we do it periodically, maybe not, maybe we switch to another GitHub action tomorrow. 

The important information is that we do CVE scanning with trivy and that the results are in the GitHub action logs.

We should not "document" the implementation details, because they might change at any time.

chasecadet:
@juliusvonkohout  one thing I really appreciate about you is that it seems you value efficiency here and do not want to waste cycles on "dead documentation" or moving targets.  

I am extraction this from

> **No one is going to synchronize this every time** 

and 

> Just saying again what is in the workflow file and the python script does not make much sense, because everyone can run this through GPT4 and it might change in the future as well. Maybe we do it periodically, maybe not, maybe we switch to another GitHub action tomorrow.

I would love it if you could help my novice Kubeflow/solutions engineer brain understand  **No one is going to synchronize this every time** better. 

It sounds like a lot of these processes  are in flux(CNCF pun intended but..not actually in Flux Flux loll) so documenting might just be whack-a-mole as community processes settle.

 I am wondering if it is at least worth communicating to users that we are introducing new processes (we are an incubating project after al) and they themselves could be playing whack-a-mole with their systems/communication if they are under the assumption we are more static with our processes.
 
   The reason I say this is, the less we communicate  the more of the story someone can fill in themselves.  In my experience, over communicating does put us out there, but then leaves less to a user/orgs imagination. This of course can lead to some scrutiny, but in the end scrutiny creates more data and I believe (and I have a feeling you do too) that we benefit from data driven decisions unless you have a ground truth you are following we can reference as a "this is where we are heading" kind of deal.  I.E. we are working on mirroring another communities processes. 



This of course aligns to my assumption that our whole goal here is for users to understand their level of effort and risk when implementing our tooling into their environments. We all know that there is no such thing as perfectly secure code(https://github.com/kelseyhightower/nocode), but our stance/capabilities can help users make informed decisions as they consider adoption.  I could be completely wrong. Feel free to hit me with a newspaper and correct my course. I know my curiosity can be exhausting, but if you help me understand I will make it my mission to help others understand as well and  prevent similar PR spam :-). 

Thanks for your time and effort on this!

juliusvonkohout:
I will try to rephrase it. 
If for example someone changes the implementation detail from running on each merge to the master branch to every 24 hours, then in practice the documentation is outdated/wrong. This happened all the time, we even had kfctl documentation a few weeks ago. 

I finally managed to get rid of most outdated documentation over the last year and I am very hesitant to document technical implementation  details that everyone really interested can read faster from the workflow and python file. It also has no benefit for the maintainers/developers, because the single source of truth is the workflow and python file.

For other components such as oauth2-proxy we actually documented something because it is a quite complex system. The trivy scanning is really straightforward and the "natural language documentation" would actually be less useful than the workflow file itself. The workflow file is also English, but better structured.

But we can definitely improve the section in the main Readme or add comments in the python and workflow file directly. My stuff is definitely not flawless. Please create follow up PRs.

Another thing you @chasecadet and @StefanoFioravanzo could drive is exposing this high-level information on the website.
We also have to merge e.g. the disclosure  policy from kubeflow/kubeflow/security into the main Readme and the website. Please create PRs and tag me.

If you add "allow edits by maintainers" I take this as an invite to also work in your branches.

Issue #2794, opened 2024-07-10
RBAC Access denied from Pipeline Run pod


kimwnasptd:
### Validation Checklist

- [X] Is this a Kubeflow issue?
- [X] Are you posting in the right repository ?
- [X] Did you follow the installation guide https://github.com/kubeflow/manifests?tab=readme-ov-file ?
- [X] Is the issue report properly structured and detailed with version numbers?
- [ ] Is this for Kubeflow development ?
- [ ] Would you like to work on this issue?
- [ ] You can join the CNCF Slack and access our meetings at the [Kubeflow Community](https://www.kubeflow.org/docs/about/community/) website. Our channel on the CNCF Slack is here [**#kubeflow-platform**](https://app.slack.com/client/T08PSQ7BQ/C073W572LA2).

### Version

1.9

### Describe your issue

Using the RC2 of Kubeflow and trying to create a Pipeline run. The pipeline fails and I see in the logs of the driver pod of the first step:

```
time="2024-07-10T10:58:15.813Z" level=info msg="capturing logs" argo=true
I0710 10:58:15.905160      33 main.go:108] input ComponentSpec:{
  "executorLabel": "exec-preprocess",
  "inputDefinitions": {
...
F0710 10:58:16.006554      33 main.go:79] KFP driver: driver.Container(pipelineName=tutorial-data-passing, runID=486a5fd7-39fe-4ea7-82f0-3162fcd1e421, task="preprocess", component="comp-preprocess", dagExecutionID=2, componentSpec) failed: failure while getting executionCache: failed to list tasks: rpc error: code = PermissionDenied desc = RBAC: access denied
time="2024-07-10T10:58:16.814Z" level=info msg="sub-process exited" argo=true error="<nil>"
time="2024-07-10T10:58:16.814Z" level=error msg="cannot save parameter /tmp/outputs/pod-spec-patch" argo=true error="open /tmp/outputs/pod-spec-patch: no such file or directory"
time="2024-07-10T10:58:16.814Z" level=error msg="cannot save parameter /tmp/outputs/cached-decision" argo=true error="open /tmp/outputs/cached-decision: no such file or directory"
time="2024-07-10T10:58:16.814Z" level=info msg="/tmp/outputs/condition -> /var/run/argo/outputs/parameters//tmp/outputs/condition" argo=true
Error: exit status 1
```

Looks like someone (Istio most probably) is denying a request from the driver to get the `executionCache`

### Steps to reproduce the issue

[...]

PR #2795, opened 2024-07-11
Fix ml pipeline access from kfp step

kromanow94:
# Fix ml pipeline access from kfp step

## ✏️ A brief description of the changes
I changed the:
* GH Action defined in `.github/workflows/pipeline_test.yaml` to fail if the KF Pipeline Run has failed,
* `AuthorizationPolicy/ml-pipeline` to allow access for both scenarios:
  * when request doesn't have `kubeflow-userid` header to allow access from KFP Steps
  * when request have a valid JWT trusted by Istio to allow programmatic access

## 📦 List any dependencies that are required for this change
N/A

## 🐛 If this PR is related to an issue, please put the link to the issue here.
https://github.com/kubeflow/manifests/issues/2794

## ✅ Contributor checklist
  - All the commits have been _signed-off_  (To pass the `DCO` check)
  - Submit the [Contributor License Agreements](https://cla.developers.google.com/clas) (To pass the `cla/google` check)

@kimwnasptd , @juliusvonkohout

### DIFF
--- a/.github/workflows/pipeline_test.yaml
+++ b/.github/workflows/pipeline_test.yaml
@@ -90,11 +90,14 @@ jobs:
 
           while True:
             status = client.get_run(run_id=run_id).state
-            if status not in ["SUCCEEDED", "FAILED", "ERROR"]:
+            if status in ["PENDING", "RUNNING"]:
               print(f"Waiting for run_id: {run_id}, status: {status}.")
               sleep(10)
             else:
               print(f"Run with id {run_id} finished with status: {status}.")
+              if status != "SUCCEEDED":
+                print("Pipeline failed")
+                raise SystemExit(1)
               break
           ' "${TOKEN}" "${KF_PROFILE}"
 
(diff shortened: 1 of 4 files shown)


juliusvonkohout:
Thank you for the PR, I think the changes here should also trigger https://github.com/kubeflow/manifests/blob/master/.github/workflows/pipeline_run_from_notebook.yaml.

.github/workflows/pipeline_test.yaml
```diff
@@ -90,11 +90,14 @@ jobs:
 
           while True:
             status = client.get_run(run_id=run_id).state
-            if status not in ["SUCCEEDED", "FAILED", "ERROR"]:
+            if status in ["PENDING", "RUNNING"]:
```
kimwnasptd, line 93: @kromanow94 a bit late to request this change, but could we not have the python code directly on the GH action but convert to a `.py` file, and use that from the action?

This will allow users to also run the tests locally if they want

kromanow94: I think it makes a lot of sense. Do you mind if I do it next week?

yhwang: just want to say I also vote for moving to a `.py` file.

kromanow94: I think this is a part of a wider topic to reorganize tests. I created an issue for that:

https://github.com/kubeflow/manifests/issues/2797

PR #2847, opened 2024-08-17
Enable istio support for ray

hansinikarunarathne:

   I followed the steps in [https://docs.ray.io/en/latest/cluster/kubernetes/k8s-ecosystem/istio.html#step-4-create-the-raycluster](https://docs.ray.io/en/latest/cluster/kubernetes/k8s-ecosystem/istio.html#step-4-create-the-raycluster ) except the step 3. 
   
 
image
image
image
inside the notebook

## ✅ Contributor checklist
  - All the commits have been _signed-off_  (To pass the `DCO` check)
  - Submit the [Contributor License Agreements](https://cla.developers.google.com/clas) (To pass the `cla/google` check)

---     
 

contrib/ray/raycluster_example.yaml
```diff
@@ -1,49 +1,29 @@
 apiVersion: ray.io/v1
 kind: RayCluster
 metadata:
-  name: kubeflow-raycluster
+  name: raycluster-istio
```
juliusvonkohout, line 4: Renaming this breaks the test. can you keep it as kubeflow-raycluster ?

juliusvonkohout: And we should install to the default user namespace if that is not already the case.

juliusvonkohout:
@alex-treebeard @kevin85421 @wadhah101 can you take a look as well, since you contributed here before?

contrib/ray/raycluster_example.yaml
```diff
@@ -2,48 +2,29 @@ apiVersion: ray.io/v1
 kind: RayCluster
 metadata:
   name: kubeflow-raycluster
+  namespace: kubeflow
```
juliusvonkohout, line 5: This is not the right namespace. It should be the customer/user namespace

hansinikarunarathne: Can you explain bit more

contrib/ray/raycluster_example.yaml
```diff
@@ -57,7 +37,7 @@ spec:
           # entire Kubernetes node on which it is scheduled.
           resources:
             limits:
-              cpu: "1"
+              cpu: "2"
```
juliusvonkohout, line 40: Why did you change this?

contrib/ray/raycluster_example.yaml
```diff
@@ -57,7 +37,7 @@ spec:
           # entire Kubernetes node on which it is scheduled.
```
juliusvonkohout, line 37: I think this comment can be deleted

contrib/ray/raycluster_example.yaml
```diff
@@ -73,68 +53,52 @@ spec:
...
+            volumeMounts:
+             - mountPath: /tmp/ray
+               name: ray-logs
+            # The resource requests and limits in this config are too small for production!
+            # It is better to use a few large Ray pod than many small ones.
+            # For production, it is ideal to size each Ray pod to take up the
+            # entire Kubernetes node on which it is scheduled.
+            resources:
+             limits:
+               cpu: "2"
```
juliusvonkohout, line 90: Only change values if you have a valid reason

contrib/ray/raycluster_istio_auth_policy.yaml
juliusvonkohout, line 1: This looks very dangerous.

.github/workflows/ray_test.yaml
juliusvonkohout, line 1: Where is the example namespace and multi-tenancy e.g. kubeflow-user-example-com from https://github.com/kubeflow/manifests/tree/master/common/user-namespace/base

contrib/ray/raycluster_istio_auth_policy.yaml
juliusvonkohout: In the Kubeflow namespace there should be only kuberay operator, no raycluster worker or head. Raycluster,worker,head belong into the namespace/kubeflow profile kubeflow-user-example-com

contrib/ray/kuberay-operator/overlays/standalone/namespace.yaml
juliusvonkohout, line 1: This file should be removed

contrib/ray/test.sh
```diff
@@ -30,17 +30,27 @@ function trap_handler {
 
   # Delete KubeRay operator
   kustomize build kuberay-operator/base | kubectl -n $NAMESPACE delete -f -
+
 }
 
 trap trap_handler EXIT
 
+kubectl get namespaces --selector=istio-injection=enabled
```
juliusvonkohout, line 38: kubeflow-user-example-com is the namespace. It must be a proper multi-tenancy supported kubeflow profile namespace

contrib/ray/raycluster_example.yaml
juliusvonkohout: We will also get a cleaner difference in git if you only change relevant values.

contrib/ray/raycluster_istio_auth_policy.yaml
juliusvonkohout: The authorization policy should be in the file contrib/ray/raycluster_example.yaml as well as additional resource and it should only allow the istio service mesh traffic between ray workers, not more and not less.

hansinikarunarathne:
did the requested changes. please review @juliusvonkohout

contrib/ray/raycluster_example.yaml
```diff
@@ -1,60 +1,112 @@
...
+kind: AuthorizationPolicy
+metadata:
+  name: allow-ray-workers-head
+spec:
+  action: ALLOW
+  rules:
+  - to:
+    - operation:
+        ports:
+        - "6379"
```
juliusvonkohout, line 11: Please make sure that you only allow traffic from the default-editor serviceaccount (the one that the workers use as well) in the same namespace

contrib/ray/raycluster_example.yaml
```diff
@@ -1,60 +1,112 @@
...
+    appProtocol: grpc
+  - name: p10010
+    port: 10010
+    appProtocol: grpc
+  - name: p10011
+    port: 10011
+    appProtocol: grpc
+  - name: p10012
+    port: 10012
+    appProtocol: grpc
```
juliusvonkohout, line 75: Why do you need so many ports and why is only one in the authorizationpolicy?

contrib/ray/raycluster_example.yaml
```diff
@@ -1,60 +1,112 @@
...
+      max-worker-port: '10012'
+      node-ip-address: $(hostname -I | tr -d ' ' | sed 's/\./-/g').raycluster-istio-headless-svc.kubeflow-user-example-com.svc.cluster.local
     template:
-      metadata:
-        # Custom labels. NOTE: To avoid conflicts with KubeRay operator, do not define custom labels start with `raycluster`.
-        # Refer to https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/
-        # The ray head must not have an Istio sidecar
-        # TODO add an authorizationpolicy in the future for the ray head
-        labels:
-          sidecar.istio.io/inject: "false"
```
juliusvonkohout, line 35: can you keep 

```
        labels:
          sidecar.istio.io/inject: "true"
```
to have it explicit ?

contrib/ray/test.sh
```diff
@@ -2,46 +2,30 @@
...
-    fi
-    sleep $SLEEP_INTERVAL
-  done
-
-  # Delete KubeRay operator
-  kustomize build kuberay-operator/base | kubectl -n $NAMESPACE delete -f -
-}
-
-trap trap_handler EXIT
+while [[ $(kubectl get namespace $NAMESPACE --no-headers 2>/dev/null | wc -l) -eq 0 ]]; do
```
juliusvonkohout, line 10: No while loops, this can run indefinitely. Please use a for loop with a maximum time

contrib/ray/raycluster_example.yaml
hansinikarunarathne: Did it

contrib/ray/raycluster_example.yaml
hansinikarunarathne: --node-manager-port: Raylet port for node manager.
--object-manager-port: Raylet port for object manager.
--runtime-env-agent-port: Raylet port for runtime env agent.
--dashboard-agent-grpc-port: The port to listen for grpc on
--max-worker-port: Maximum port number worker can be bound to.
--metrics-export-port: The port to use to expose Ray metrics.

These ports are needed when using the Ray Cluster. The access is given for the relevant ports through the authorization policy

contrib/ray/raycluster_example.yaml
```diff
@@ -1,60 +1,96 @@
...
+kind: AuthorizationPolicy
+metadata:
+  name: allow-ray-workers-head
+spec:
+  action: ALLOW
+  rules:
+  - from:
+    - source:
+        principals:
+        - "cluster.local/ns/kubeflow-user-example-com/sa/kubeflow-raycluster"
```
juliusvonkohout, line 11: This must be generic. The namespace could also have another name that you do not know in advance. The same manifests should work for different users.

contrib/ray/test.sh
```diff
@@ -2,46 +2,44 @@
...
+  if [ "$elapsed_time" -ge "$TIMEOUT" ]; then
+    echo "Timeout exceeded. Namespace $NAMESPACE not created."
+    exit 1
+  fi
+  
+  echo "Waiting for namespace $NAMESPACE to be created..."
+  sleep 2
+done
+
+echo "Namespace $NAMESPACE is created!"
```
juliusvonkohout, line 29: Has been created

contrib/ray/raycluster_example.yaml
juliusvonkohout, line 1: Please post a screenshot of the dashboard

contrib/ray/raycluster_example.yaml
hansinikarunarathne: Is there a way to templating that namespace value without using helm?

juliusvonkohout: The Serviceaccount is wrong. It should be default-editor everywhere. Especially in the raycluster itself. 

You can take a look at https://istio.io/latest/docs/reference/config/security/authorization-policy/ for ideas and ask gpt4. Maybe you can just whitelist the whole namespace.

contrib/ray/raycluster_example.yaml
hansinikarunarathne: image
image
image
inside the notebook

contrib/ray/raycluster_example.yaml
hansinikarunarathne: Is it only enough to generalize the service account name(with default-editor)? Because since we are deploying it in a namespace, we have to specify the namespace to give the access to only for that specific namespace(kubeflow-user-example in this case)

contrib/ray/raycluster_example.yaml
hansinikarunarathne: I added ss to descriptions as well

contrib/ray/test.sh
hansinikarunarathne: Resolved

contrib/ray/raycluster_example.yaml
juliusvonkohout: No the other way around. We need to generalize the namespace. It would be alright to allow all serviceaccounts or all pods from the namespace. But the deployments should still run as default-editor. This must be deployable without manual changes per namespace.

juliusvonkohout: Are you setting the default-editor in the raycluster resource itself?

contrib/ray/raycluster_example.yaml
```diff
@@ -73,68 +109,51 @@ spec:
...
+            securityContext:
+             allowPrivilegeEscalation: false
+             capabilities:
+               drop: ["ALL"]
+             runAsNonRoot: true
+             seccompProfile:
+               type: RuntimeDefault
+          volumes:
+            - name: ray-logs
+              emptyDir: {}
```
juliusvonkohout, line 159: I am missing the Serviceaccount default-editor

contrib/ray/raycluster_example.yaml
hansinikarunarathne: I added the default-editor service account to the deployment. 
> We need to generalize the namespace.

This means should I add like this,
cluster.local/ns/kubeflow-user-example-com/sa/*

contrib/ray/raycluster_example.yaml
hansinikarunarathne: added

contrib/ray/raycluster_example.yaml
hansinikarunarathne: I did the changes . @juliusvonkohout

juliusvonkohout:
Please remove the part of "## Step 3: Create a namespace" there shall be no manual namespace creation. All of this is handled by kubeflow profiles.

PR #2908, opened 2024-11-04
Synchronize kserve manifests v0.14.0

juliusvonkohout:
@tarekabouzeid @rimolive for a /lgtm


tarekabouzeid:
@juliusvonkohout  can you restart failed pipelines ?

juliusvonkohout:
> @juliusvonkohout can you restart failed pipelines ?

Yes, now also the update script with the version in the readme should be fixed

juliusvonkohout:
Kserve crds are now so big, that they do not work with client-side apply anymore... But this is now the case for many projects...

Issue #2676, opened 2024-04-11
Istio Ambient support


peterj:
Istio Ambient mode is a different deployment model from the “traditional” (sidecar) mode of Istio. The ambient mode (sidecar-less) doesn’t require injecting sidecars into the deployments.

Here are the high-level differences between the two modes:

- The concerns handled by the sidecar proxy in the sidecar Istio are split into two components in Istio ambient: 

  - **Ztunnel** (handles L4 concerns, mTLS, authorization policies without any HTTP)

    - Ztunnel is installed automatically when profile=ambient

  - **Waypoint proxy** (handles L7 concerns, i.e. traffic splitting, matching, header manipulation, etc., more or less everything that gets defined in the VirtualService)

    - Waypoint is **optional** (not installed by default) and it can be deployed per service account (handles all workloads using the same service account) or  per namespace (handles L7 proxying for all workloads in the namespace)  

- Ingress gateway isn’t installed by default anymore when using profile=ambient

  - It might be worth migrating over to Kubernetes Gateway APIs and deploying the Istio ingress gateway like that, as we’ll have to use the Gateway APIs to deploy waypoint proxies anyway

- Kubernetes Gateway API is used for ingress (and waypoint proxy) deployments

- Any L7 VirtualServices or AuthorizationPolicies must have a “targetRef” section that specifies which waypoint proxy handles the L7 configuration

## Waypoint proxies<a id="waypoint-proxies"></a>

Any VirtualService or AuthorizationPolicy that uses HTTP concepts will require a waypoint proxy. Given that there are 3 namespaces (that I identified so far), I’d suggest a per-namespace deployment of a waypoint proxy.

In addition to the waypoint proxy, the resources will have to be updated to use the waypoint proxies: 

| **Component**                | **Namespace** | **Notes** |
| --- | --- | --- |
| dex                          | auth          |           |
| central-dashboard            | kubeflow      |           |
| jupyter-web-app              | kubeflow      |           |
| volumes-web-app              | kubeflow      |           |
| katib-ui                     | kubeflow      |           |
| ml-pipeline-ui               | kubeflow      |           |
| metadata-envoy-service       | kubeflow      |           |
| kfp-tekton                   | kubeflow      |           |
| kubebench-dashboard          | kubeflow      |           |
| profiles-kfam                | kubeflow      |           |
| tensorboards-web-app-service | kubeflow      |           |
| kserve-models-web-app        | kserve        |           |

## Work items

[...]

PR #2950, opened 2025-01-11
Upgrade Istio to v1.24.2

biswajit-9776:

## ✅ Contributor checklist
  - All the commits have been [_signed-off_](https://github.com/kubeflow/community/tree/master/dco-signoff-hook#signing-off-commits)  (To pass the `DCO` check)

---     
 

biswajit-9776:
Hello @juliusvonkohout, I used the same [profile.yaml](https://github.com/kubeflow/manifests/blob/master/common/istio-1-23/profile.yaml) to install istio-1.24.2 and manually changed _spec.tag: 1.24.2_ as they have removed the _istioctl profile_ command in their 1.24.x release.

However, inspecting the manifest it generates, seems they have removed the ingressgateway service. Could you confirm if this is true as I'm unable to check it in their release docs.

juliusvonkohout:
> Hello @juliusvonkohout, I used the same [profile.yaml](https://github.com/kubeflow/manifests/blob/master/common/istio-1-23/profile.yaml) to install istio-1.24.2 and manually changed _spec.tag: 1.24.2_ as they have removed the _istioctl profile_ command in their 1.24.x release.
> 
> However, inspecting the manifest it generates, seems they have removed the ingressgateway service. Could you confirm if this is true as I'm unable to check it in their release docs.

That is interesting are you sure that profiles are gone? Can you check the git diff ?

juliusvonkohout:
> The profiles are there which you could see in [1.24 release](https://github.com/istio/istio/tree/release-1.24/manifests/profiles), they however [removed](https://istio.io/latest/news/releases/1.24.x/announcing-1.24/change-notes/#:~:text=Removed%20istioctl%20profile%20command.%20The%20same%20information%20can%20be%20found%20in%20Istio%20documentation.) the _istioctl profile_ command which we were able to use before. We could still use our existing profile.yaml for that manually. What do you think about the ingress gateway service? Should we add it as a patch in the 1.24 upgrade?

Could it be that it has been replaced by https://gateway-api.sigs.k8s.io/ ? But yes we definitely need some kind of gateway, either the old style or the new API and a service that exposes it within Kubernetes.

biswajit-9776:
Well we did plan to adopt the gateway API as per [issue](https://github.com/kubeflow/manifests/issues/2676#:~:text=Migrate%20to%20Kubernetes%20Gateway%20API).
Do correct me if I am wrong, we could either first switch to Gateway API and then upgrade to Istio 1.24.x or include the service patches like the old style in this PR and later migrate to Gateway API.

juliusvonkohout:
> Well we did plan to adopt the gateway API as per [issue](https://github.com/kubeflow/manifests/issues/2676#:~:text=Migrate%20to%20Kubernetes%20Gateway%20API). Do correct me if I am wrong, we could either first switch to Gateway API and then upgrade to Istio 1.24.x or include the service patches like the old style in this PR and later migrate to Gateway API.

Lets keep the changer per PR small for easier reviews, so try to replicate the status quo first.

juliusvonkohout:
What a PR. Took me an hour to review :-D

@biswajit-9776 can you create the follow up PR to switch to the gateway API ? 

/lgtm
/approve

biswajit-9776:
Hope the changes didn't bother you much :)
I will see if Gateway API works locally and replaces the work done by files I added to this PR.

PR #2954, opened 2025-01-17
Synchronize kubeflow training operator manifests v1.9.0 rc.0

juliusvonkohout:
@tarekabouzeid @varodrig  for /lgtm


juliusvonkohout:
`Error from server (Invalid): error when creating "STDIN": CustomResourceDefinition.apiextensions.k8s.io "pytorchjobs.kubeflow.org" is invalid: metadata.annotations: Too long: may not be more than 262144 bytes` should actually be fixed by a server-side apply. Lets investigate

PR #2960, opened 2025-01-20
Add the port name to the ml-pipeline-ui container

mprahl:
Without the port name, the ml-pipeline-ui Service was not working properly.


juliusvonkohout:
@mprahl can you elaborate a bit more about what is not working?

PR #2953, opened 2025-01-17
Synchronize kubeflow pipelines manifests 2.4.0

juliusvonkohout:
@tarekabouzeid @varodrig this is WIP and for your review.

PR #2964, opened 2025-01-20
Disable the GKE metadata endpoints on Pipelines by default

mprahl:

## ✏️ A brief description of the changes

When not explicitly set, the Pipelines UI queries the /pipeline/system/cluster-name and /pipeline/system/project-id endpoints, which cause the ml-pipeline-ui pods to crash when it's not deployed on GKE.

---     
 

juliusvonkohout:
Heloo @mprahl please never ever touch the upstream folders. They are synchronized from Kubeflow/pipelines. It should be  fixed in https://github.com/kubeflow/pipelines/pull/11403 and will be synchronized in https://github.com/kubeflow/manifests/pull/2953. If you think that is not true, please create a new PR

PR #2977, opened 2025-02-05
Update oauth2-proxy README.md

juliusvonkohout:
@tarekabouzeid  @varodrig for /lgtm


kromanow94:
A better fix would be to use Redis and connect with oauth2-proxy to store the JWT in there and provide only the key to the JWT in the cookie header.

But, AFAIR, Kubeflow still has limited to no group support so this also works.

Issue #2980, opened 2025-02-06
katib-db-manager : Ping to Katib db failed: dial tcp connect: connection refused


pedrovgp:
### Validation Checklist

- [x] Is this a Kubeflow issue?
- [x] Are you posting in the right repository ?
- [x] Did you follow the [Kubeflow installation guideline](https://github.com/kubeflow/manifests/blob/master/README.md) ?
- [x] Is the issue report properly structured and detailed with version numbers?
- [ ] Is this for Kubeflow development ?
- [x] Would you like to work on this issue?
- [ ] You can join the CNCF Slack and access our meetings at the [Kubeflow Community](https://www.kubeflow.org/docs/about/community/) website. Our channel on the CNCF Slack is here [**#kubeflow-platform**](https://app.slack.com/client/T08PSQ7BQ/C073W572LA2).

### Version

master

### Describe your issue

We are installing Kubeflow 1.10 rc0 on AWS EKS with Kubernetes 1.30.

We did not make any customizations to Katib, we are using the install "apps/katib/upstream/installs/katib-with-kubeflow".

The "katib-db-manager" deploy enters CrashLoopBackOff

We found some similar issues, [1](https://github.com/kubeflow/katib/issues/2425, [2](https://github.com/kubeflow/katib/issues/1115) and [3](https://github.com/docker-library/mysql/issues/653).

We have debugged, I'll open a PR with the investigation and solution proposal.

### Steps to reproduce the issue

1. Checkout the v1.10rc0 branch.
2. Create an EKS cluster in Kubernetes 1.30 version.
3. Configure the cluster to be used by kubectl.
4. Run kubectl apply -k example

juliusvonkohout:
We have not yet upgraded Katib . It is still the same as in Kubeflow 1.9.1. Also here please reproduce with Kind to get help.

pedrovgp:
It's all right, I just raised the ticket for reference, before suggesting the solution. Here is the [PR](https://github.com/kubeflow/manifests/pull/2981)

PR #2981, opened 2025-02-06
FIX katib-db-manager secrets and service to connect to mysql

pedrovgp:
This PR was originated by [issue 2980](https://github.com/kubeflow/manifests/issues/2980)

Investigating the issue, we discovered two problems:
1. katib-db-manager was using the secret katib-mysql-secret to connect to the database. But in this install (1) the database used by katib is mysql (not katib-mysql, which is not even deployed) and (2) the secrets (username and password) are different
2. katib-db-manager points to the Service katib-mysql (instead of pointing to the mysql service), which is fine, since the service should point to the same Deployment as the mysql Service. But the selector was not working properly, and it was not pointing to the right Deployment.

This PR fixes both issues by adding patches to the "katib-with-kubeflow" install mode.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ This repo periodically syncs all official Kubeflow components from their respect
 
 | Component | Local Manifests Path | Upstream Revision |
 | - | - | - |
-| Training Operator | apps/training-operator/upstream | [v1.9.0](https://github.com/kubeflow/training-operator/tree/v1.9.0/manifests) |
+| Training Operator | apps/training-operator/upstream | [778cd72](https://github.com/kubeflow/training-operator/tree/778cd72/manifests) |
 | Notebook Controller | apps/jupyter/notebook-controller/upstream | [v1.9.2](https://github.com/kubeflow/kubeflow/tree/v1.9.2/components/notebook-controller/config) |
 | PVC Viewer Controller | apps/pvcviewer-controller/upstream | [v1.9.2](https://github.com/kubeflow/kubeflow/tree/v1.9.2/components/pvcviewer-controller/config) |
 | Tensorboard Controller | apps/tensorboard/tensorboard-controller/upstream | [v1.9.2](https://github.com/kubeflow/kubeflow/tree/v1.9.2/components/tensorboard-controller/config) |
--- a/apps/training-operator/upstream/overlays/kubeflow/kubeflow-training-roles.yaml
+++ b/apps/training-operator/upstream/overlays/kubeflow/kubeflow-training-roles.yaml
@@ -27,6 +27,7 @@ rules:
       - pytorchjobs
       - xgboostjobs
       - paddlejobs
+      - jaxjobs
     verbs:
       - create
       - delete
@@ -43,6 +44,7 @@ rules:
       - pytorchjobs/status
       - xgboostjobs/status
       - paddlejobs/status
+      - jaxjobs/status
     verbs:
       - get
   - apiGroups:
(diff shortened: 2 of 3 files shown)


juliusvonkohout:
@pedrovgp you need to fix it here https://github.com/kubeflow/katib/tree/master/manifests/v1beta1/installs/katib-with-kubeflow and then synchronize with https://github.com/kubeflow/manifests/blob/master/hack/synchronize-katib-manifests.sh.

But i am wondering because our cicd works well https://github.com/kubeflow/manifests/blob/master/.github/workflows/katib_test.yaml as you can see here https://github.com/kubeflow/manifests/actions/runs/13119220501/job/36600993054

juliusvonkohout:
CC @andreyvelich

pedrovgp:
> But i am wondering because our cicd works well https://github.com/kubeflow/manifests/blob/master/.github/workflows/katib_test.yaml as you can see here https://github.com/kubeflow/manifests/actions/runs/13119220501/job/36600993054

Indeed, that's weird. You mentioned it is the same as in Kubeflow 1.9. Katib in 1.9 was also not working for us.

I don't think it is anything EKS related because the secrets and app mappings were wrong within the cluster, it had nothing to do with any external AWS resources.

juliusvonkohout:
@kubeflow/release-team

juliusvonkohout:
@pedrovgp 

also

```
# Warning: 'patchesStrategicMerge' is deprecated. Please use 'patches' instead. Run 'kustomize edit fix' to update your Kustomization automatically.
# Warning: 'vars' is deprecated. Please use 'replacements' instead. [EXPERIMENTAL] Run 'kustomize edit fix' to update your Kustomization automatically.
# Warning: 'patchesStrategicMerge' is deprecated. Please use 'patches' instead. Run 'kustomize edit fix' to update your Kustomization automatically.
# Warning: 'vars' is deprecated. Please use 'replacements' instead. [EXPERIMENTAL] Run 'kustomize edit fix' to update your Kustomization automatically.
```
needs to be fixed upstream

juliusvonkohout:
and the tests must be fixed to fail on such problems. Maybe wee need to extend https://github.com/kubeflow/manifests/blob/588e49e267f070578ad0e7a717d9ba79d652371d/.github/workflows/katib_test.yaml to also check whether stuff is persisted to the database.

pedrovgp:
Hi @juliusvonkohout , I made the [PR in the Katib repo](https://github.com/kubeflow/katib/pull/2510).

Should I now wait for it to merge to master and after that run https://github.com/kubeflow/manifests/blob/master/hack/synchronize-katib-manifests.sh. and make the PR here?

juliusvonkohout:
> Hi @juliusvonkohout , I made the [PR in the Katib repo](https://github.com/kubeflow/katib/pull/2510).
> 
> Should I now wait for it to merge to master and after that run https://github.com/kubeflow/manifests/blob/master/hack/synchronize-katib-manifests.sh. and make the PR here?

Well as a first step you should extend the tests mentioned above to fail with the current setup. Such problems must be caught in our CI/CD. But otherwise yes.

juliusvonkohout:
/ok-to-test

@pedrovgp please make sure to also have the changes int the upstream katib master branch. In the best case we keep them in synchronization for now.

juliusvonkohout:
@pedrovgp why did you close the PR ?

juliusvonkohout:
ah ok, continued in https://github.com/kubeflow/manifests/pull/2983

PR #2983, opened 2025-02-07
Adding tests for katib mysql

pedrovgp:
This is actually from [this PR](https://github.com/kubeflow/manifests/pull/2981), which I accidentally closed when trying to add the tests only.


juliusvonkohout:
You should use 300-600s as timeout instead of 5s and add the changes from the previous PR.

pedrovgp:
@juliusvonkohout, I could not reproduce the problem with Kind. And even in the cluster I was testing, the problem did not happen again. I believe the deployment of katib-mysql failed in the previous attempt, and that led to the investigation I mentioned. I'll close this one for now, and I'll re-open if we find anything new.

Thanks for guidance.

PR #2973, opened 2025-02-03
Update kubeflow/model-registry manifests from 5d8ed91

tarilabs:
# Pull Request Template for Kubeflow manifests Issues

## ✏️ A brief description of the changes
Updated KF manifests for MR following most recent release update

## 📦 List any dependencies that are required for this change
none

## 🐛 If this PR is related to an issue, please put the link to the issue here.
for KF 1.10 RC builds

## ✅ Contributor checklist
  - Make sure you have tested with kustomize. See [Installation Prerequisites](https://github.com/kubeflow/manifests#prerequisites)
  - All the commits have been [_signed-off_](https://github.com/kubeflow/community/tree/master/dco-signoff-hook#signing-off-commits)  (To pass the `DCO` check)

---     
 
> You can join the CNCF Slack and access our meetings at the [Kubeflow Community](https://www.kubeflow.org/docs/about/community/) website. Our channel on the CNCF Slack is here [**#kubeflow-platform**](https://app.slack.com/client/T08PSQ7BQ/C073W572LA2).

PR #2998, opened 2025-02-16
Synchronize kubeflow model registry manifests v0.2.14

tarilabs:

## ✏️ Summary of Changes
1. make script BSD/MacOSX compatible (e1dcefeab598d13a629fb722f838fa654b20b63a)
2. update the manifests
- include [this request](https://github.com/kubeflow/manifests/pull/2973#issuecomment-2645813716) to make sure to update the readme
- include [this request](https://github.com/kubeflow/manifests/pull/2973#issuecomment-2636264568) to add securityContext

## 📦 Dependencies
none related

## 🐛 Related Issues
none related

## ✅ Contributor Checklist

---     
 
### DIFF
--- a/apps/model-registry/upstream/base/model-registry-deployment.yaml
+++ b/apps/model-registry/upstream/base/model-registry-deployment.yaml
@@ -16,6 +16,10 @@ spec:
       labels:
         component: model-registry-server
     spec:
+      securityContext:
+        seccompProfile:
+          type: RuntimeDefault
+        runAsNonRoot: true
       containers:
         - name: rest-container
           args:
@@ -44,6 +48,11 @@ spec:
             tcpSocket:
               port: http-api
             timeoutSeconds: 2
+          securityContext:
+            allowPrivilegeEscalation: false
+            capabilities:
+              drop:
+                - ALL
         - name: grpc-container
           # ! Sync to the same MLMD version:
           # * backend/metadata_writer/requirements.in and requirements.txt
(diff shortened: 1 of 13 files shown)


apps/model-registry/upstream/base/model-registry-deployment.yaml
juliusvonkohout, line 1: I am wondering a bit why you split the securitycontext here, but in general its good.

juliusvonkohout:
/approve

@tarilabs just check the comment and /lgtm and merge as needed.

juliusvonkohout:
@tarilabs just check the comment and maybe follow up

/lgtm

apps/model-registry/upstream/base/model-registry-deployment.yaml
pboyd: Hi @juliusvonkohout, can these be combined? I don't think they can be (except for perhaps a YAML anchor), but I don't especially like the duplication between them, so if you have a way I'd love to use it.

I see that [`PodSecurityContext`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.31/#podsecuritycontext-v1-core) settings are inherited by containers, but it doesn't include `allowPrivledgeEscalation` or `capabilities`. As far as I can tell, the only way to set those is on each container ([`SecurityContext`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.31/#securitycontext-v1-core)), and this pod has two containers.

I could have moved `runAsUser` and `runAsGroup` up to the pod level, but it was only needed on the MLMD container, since the model registry image already used a non-root user.

PR #3050, opened 2025-03-13
Fix PSS restricted warnings for kubeflow components

akagami-harsh:
# Pull Request Template for Kubeflow Manifests

## ✏️ Summary of Changes
- fixed PSS restricted warning for katib
- working on fixing for other components

## 🐛 Related Issues
- #3015

## ✅ Contributor Checklist
  - [ ] I have tested these changes with kustomize. See [Installation Prerequisites](https://github.com/kubeflow/manifests#prerequisites).
  - [ ] All commits are [_signed-off_](https://github.com/kubeflow/community/tree/master/dco-signoff-hook#signing-off-commits) to satisfy the DCO check.

---     
 
> You can join the CNCF Slack and access our meetings at the [Kubeflow Community](https://www.kubeflow.org/docs/about/community/) website. Our channel on the CNCF Slack is here [**#kubeflow-platform**](https://app.slack.com/client/T08PSQ7BQ/C073W572LA2).

PR #2826, opened 2024-07-31
Add seaweedfs to contrib

pschoen-itsc:
# Pull Request Template for Kubeflow manifests Issues

## ✅ Contributor checklist
  - Make sure you have tested with kustomize. See [Installation Prerequisites](https://github.com/kubeflow/manifests#prerequisites)
  - All the commits have been _signed-off_  (To pass the `DCO` check)
  - Submit the [Contributor License Agreements](https://cla.developers.google.com/clas) (To pass the `cla/google` check)

---     
 
> You can join the CNCF Slack and access our meetings at the [Kubeflow Community](https://www.kubeflow.org/docs/about/community/) website. Our channel on the CNCF Slack is here [**#kubeflow-platform**](https://app.slack.com/client/T08PSQ7BQ/C073W572LA2).

PR #3051, opened 2025-03-14
Integrate profiles with seaweedfs

pschoen-itsc:

## ✏️ Summary of Changes
Use ACLs in seaweedfs to isolate access to artifacts by profile. When a new profile is created, credentials for this profile will be created by seaweedfs and put as secret into the new namespace. ACLs will be set to limited to access from this profile to his subdir in the bucket.

## 🐛 Related Issues
Don't know if there is currently an open issue

## ✅ Contributor Checklist

---     
 

experimental/seaweedfs/base/seaweedfs/seaweedfs-admin-secret.yaml
```diff
@@ -0,0 +1,8 @@
+kind: Secret
+apiVersion: v1
+metadata:
+  name: seaweedfs-admin
+stringData:
+  accesskey: "kubeflow-admin"
+  secretkey: "ichah6aiko6Ubae4xahthahm9oeRaed2"
```
pschoen-itsc, line 7: This is something that should be generated for every new kubeflow installation

experimental/seaweedfs/base/pipeline-profile-controller/deployment.yaml
```diff
@@ -0,0 +1,65 @@
...
+          allowPrivilegeEscalation: false
+          seccompProfile:
+            type: RuntimeDefault
+          runAsNonRoot: true
+          runAsUser: 1000
+          runAsGroup: 0
+          capabilities:
+            drop:
+            - ALL
+        image: alpine/k8s:1.29.13
```
pschoen-itsc, line 35: Change of image because we need the botocore libraries for interaction with the IAM service

pschoen-itsc:
@juliusvonkohout I need to do a bit more testing (for example UI components), but the integration between seaweedfs and kf-pipelines works that way.

experimental/seaweedfs/base/pipeline-profile-controller/deployment.yaml
juliusvonkohout: Can you make it a small patch instead of a full resource ?

experimental/seaweedfs/base/pipeline-profile-controller/sync.py
```diff
@@ -0,0 +1,445 @@
...
+                            "workflows.argoproj.io/default-artifact-repository": "default-seaweedfs-namespaced"
+                        }
+                    },
+                    "data": {
+                        "default-seaweedfs-namespaced": json.dumps({
+                            "archiveLogs": True,
+                            "s3": {
+                                "endpoint": "minio-service.kubeflow:9000",
+                                "bucket": "kubeflow",
+                                "keyFormat": f"{namespace}/artifacts/{{{{workflow.name}}}}/{{{{workflow.creationTimestamp.Y}}}}/{{{{workflow.creationTimestamp.m}}}}/{{{{workflow.creationTimestamp.d}}}}/{{{{pod.name}}}}",
```
juliusvonkohout, line 156: {namespace}/artifacts/ should be artifacts/{namespace}/

experimental/seaweedfs/base/pipeline-profile-controller/sync.py
```diff
@@ -0,0 +1,445 @@
...
+                    },
+                },
+                {
+                    "apiVersion": "v1",
+                    "kind": "ConfigMap",
+                    "metadata": {
+                        "name": "artifact-repositories",
+                        "namespace": namespace,
+                        "annotations": {
+                            "workflows.argoproj.io/default-artifact-repository": "default-seaweedfs-namespaced"
```
juliusvonkohout, line 147: what about jsut "default-namespaced" ?

experimental/seaweedfs/base/pipeline-profile-controller/sync.py
```diff
@@ -0,0 +1,445 @@
...
+            desired_resources = [
+                {
+                    "apiVersion": "v1",
+                    "kind": "ConfigMap",
+                    "metadata": {
+                        "name": "kfp-launcher",
+                        "namespace": namespace,
+                    },
+                    "data": {
+                        "defaultPipelineRoot": f"minio://kubeflow/{namespace}/v2/artifacts",
```
juliusvonkohout, line 124: {namespace}/artifacts/ should be private-artifacts/{namespace}/

experimental/seaweedfs/base/pipeline-profile-controller/sync.py
```diff
@@ -0,0 +1,445 @@
...
+            print('Received request:\n', json.dumps(parent, sort_keys=True))
+            print('Desired resources except secrets:\n', json.dumps(desired_resources, sort_keys=True))
+
+            # Moved after the print argument because this is sensitive data.
+
+            # This could lead to the creation of multiple access keys for the same profile
+            # which is not an imminent problem, but yet not that elegant. But it is simple
+            # and ensures that there is always a secret with valid keys.
+            # That is likely why metacontroller docs recommend not having any side effects in the
+            # hook function.
```
juliusvonkohout, line 386: Please take a look at https://github.com/kubeflow/pipelines/pull/7725/files#diff-c947a4fd2e05218edeaa4bf511854f1dbb7c41e893eec9aaeffb59d7cda634d6 cany out try to first tread an existing key from the secret in the users namespaces? If it already exists, just use the existing one.

```
        def get_password(self, namespace):
            secret_name = 'mlpipeline-minio-artifact'
            secret_names = [secret.metadata.name for secret in self.kubernetes_api_v1.list_namespaced_secret(namespace).items]
            print(secret_names)
            if secret_name in secret_names:
                print('found secret %s', secret_name)
                print('Using existing password for %s', namespace)
                secret = str(self.kubernetes_api_v1.read_namespaced_secret(secret_name, namespace).data)
                secret = json.loads(secret.replace('\'', '\"'))
                return base64.b64decode(secret['secretkey']).decode('utf-8')
            else:
                print('Generating new password for %s', namespace)
                alphabet = string.ascii_letters + string.digits
                return ''.join(secrets.choice(alphabet) for i in range(16))
```

juliusvonkohout, line 386: Can you elaborate a bit on that ?

experimental/seaweedfs/base/pipeline-profile-controller/sync.py
```diff
@@ -0,0 +1,445 @@
...
+                            "metadata": {
+                                "labels": {
+                                    "app": "ml-pipeline-visualizationserver"
+                                },
+                                "annotations": disable_istio_sidecar and {
+                                    "sidecar.istio.io/inject": "false"
+                                } or {},
+                            },
+                            "spec": {
+                                "containers": [{
```
juliusvonkohout, line 197: please add a securitycontext as in https://github.com/kubeflow/manifests/pull/3050 for the deprecated visualization-server and artifact-proxy

experimental/seaweedfs/base/seaweedfs/seaweedfs-admin-secret.yaml
juliusvonkohout: can you use the existing minio admin secret in the namespace kubeflow ? I think this one is also used by ml-pipeline-ui and other deployments.

experimental/seaweedfs/base/seaweedfs/seaweedfs-create-admin-user-job.yaml
```diff
@@ -0,0 +1,52 @@
+kind: Job
+apiVersion: batch/v1
+metadata:
+  name: init-seaweedfs
+spec:
+  template:
+    metadata:
+      name: init-seaweedfs
+    spec:
+      restartPolicy: Never
```
juliusvonkohout, line 10: I think it should restart until it is successful

experimental/seaweedfs/base/seaweedfs/seaweedfs-admin-secret.yaml
juliusvonkohout: We can also do it directly in this PR here. The visualization-server can just be removed and the artifact proxy can be removed from sync.py if you add

```
patches:
- target:
    kind: Deployment
    name: ml-pipeline-ui
    namespace: kubeflow
  patch: |-
    apiVersion: apps/v1
    kind: Deployment
    metadata:
      name: ml-pipeline-ui
      namespace: kubeflow
    spec:
      template:
        spec:
          containers:
          - name: ml-pipeline-ui
            env:
            - name: ARTIFACTS_SERVICE_PROXY_ENABLED
              value: 'false'
```

experimental/seaweedfs/base/seaweedfs/seaweedfs-networkpolicy.yaml
```diff
@@ -11,18 +11,20 @@ spec:
...
+    ports:
+    - port: 8333
+  - from:
     - namespaceSelector:
         matchExpressions:
         - key: kubernetes.io/metadata.name
           operator: In
           values:
           - istio-system
-    - podSelector: {}
```
juliusvonkohout, line 20: ml-pipeline-ui and maybe others need it as well without the artifact proxy

experimental/seaweedfs/istio/istio-authorization-policy.yaml
juliusvonkohout, line 1: Lets make sure that it is in line with

https://github.com/kubeflow/pipelines/blob/8d0ae5381e8366905c90009c56fd0e4807e94f0f/manifests/kustomize/third-party/minio/options/istio/istio-authorization-policy.yaml#L1-L30

experimental/seaweedfs/istio/istio-authorization-policy.yaml
```diff
@@ -12,11 +12,23 @@ spec:
...
+kind: AuthorizationPolicy
+metadata:
+  name: pipeline-profile-controller
+spec:
+  action: ALLOW
+  selector:
+    matchLabels:
+      app: kubeflow-pipelines-profile-controller
+  rules:
   - {}
```
juliusvonkohout, line 32: What is this policy for ? Who should have access to pipelines-profile-controller? I think only metacontoller should call it. Please also compare with https://github.com/kubeflow/pipelines/blob/master/manifests/kustomize/base/installs/multi-user/istio-authorization-config.yaml and https://github.com/kubeflow/pipelines/blob/8d0ae5381e8366905c90009c56fd0e4807e94f0f/manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/deployment.yaml#L10

experimental/seaweedfs/base/pipeline-profile-controller/sync.py
juliusvonkohout, line 1: Please also highlight the changes to the existing sync.py somwhow.

experimental/seaweedfs/base/kustomization.yaml
```diff
@@ -3,7 +3,5 @@ kind: Kustomization
 namespace: kubeflow
 
 resources:
-- seaweedfs-deployment.yaml
-- seaweedfs-pvc.yaml
-- seaweedfs-service.yaml
-- seadweedfs-networkpolicy.yaml
+- seaweedfs/
+- pipeline-profile-controller/
```
juliusvonkohout, line 7: - pipeline-profile-controller/ should be a patch

juliusvonkohout:
https://github.com/kubeflow/manifests/actions/runs/13866549907/job/38806718063?pr=3051 means that probably the pipeline is stuck because the secrets are not there. Let me add a tests from https://github.com/kubeflow/manifests/blob/13a72b79e6f107118bfaeeba2bb26fc21e9244b6/.github/workflows/pipeline_test.yaml#L51-L77

experimental/seaweedfs/base/pipeline-profile-controller/deployment.yaml
juliusvonkohout: You can also target the latest 1.32.x version instead of 1.29.13. please also do not use docker.io but Pleae also use https://explore.ggcr.dev/?repo=registry.k8s.io registry.k8s.io

juliusvonkohout:
> @juliusvonkohout I need to do a bit more testing (for example UI components), but the integration between seaweedfs and kf-pipelines works that way.

Thank you very much for the PR. I have added a lot of comments and clarifications.

experimental/seaweedfs/istio/istio-authorization-policy.yaml
juliusvonkohout: We might need

```
  - from:
    - source:
        principals:
        - cluster.local/ns/kubeflow/sa/ml-pipeline-ui
```

as well if we remove the artifact proxy.

experimental/seaweedfs/base/seaweedfs/seaweedfs-admin-secret.yaml
pschoen-itsc: That should be no problem.

experimental/seaweedfs/base/pipeline-profile-controller/sync.py
pschoen-itsc: During development I observed that the sync endpoint was called multiple times when creating a namespace, which resulted in creating multiple keypairs in seaweedfs. That wasn't a problem by itself, just a bunch of access keys which will never be used. During my last test (after the rebase) I didn't observe this behauviour anymore. I will keep this in mind and will remove this comment if this was resolved  just by itself.

experimental/seaweedfs/base/pipeline-profile-controller/deployment.yaml
pschoen-itsc: @juliusvonkohout Do you know if any of theese images has botocore / boto3 included? I'm only using that image because it has awscli / botocore installed. 
An alternative would be to build an own image, but I would need some directions for that process in the kubeflow project.

experimental/seaweedfs/base/pipeline-profile-controller/sync.py
```diff
@@ -0,0 +1,445 @@
...
+                        "annotations": {
+                            "workflows.argoproj.io/default-artifact-repository": "default-seaweedfs-namespaced"
+                        }
+                    },
+                    "data": {
+                        "default-seaweedfs-namespaced": json.dumps({
+                            "archiveLogs": True,
+                            "s3": {
+                                "endpoint": "minio-service.kubeflow:9000",
+                                "bucket": "kubeflow",
```
juliusvonkohout, line 155: Can we stick with mlpipeline for backwards compatibility ?

juliusvonkohout:
Do you mind squashing and signing your commits  (DCO) and rebasing to master?

juliusvonkohout:
I think you can copy from our new pipeline install script. See also https://github.com/kubeflow/manifests/pull/3093/files and https://github.com/kubeflow/manifests/blob/master/.github/workflows/pipeline_test.yaml as reference.

juliusvonkohout:
Thank you, i just re-enabled the tests

juliusvonkohout:
Ah i broke the DCO, but you can just squash my test enabling commit and take ownership of it if needed.

juliusvonkohout:
I think https://github.com/pschoen-itsc/kf-manifests/blob/integrate-profiles-with-seaweedfs/tests/gh-actions/pipeline_test.py or https://github.com/kubeflow/manifests/actions/runs/14669563838/job/41172716531?pr=3051 needs some debugging. The large testhere https://github.com/kubeflow/manifests/actions/runs/14669563828/job/41172716516?pr=3051 usually has all logs and recently i saw that seaweedfs was not able to create the bucket or so.

.github/workflows/pipeline_swfs_test.yaml
milinddethe15, line 1: For this action, use `ubuntu-latest-16-cores` runner:
```yaml
    runs-on:
      labels: ubuntu-latest-16-cores
```

pschoen-itsc:
The End-to-End Integration test succeded before, now it fails without changing anything...
@juliusvonkohout can you just retrigger it?

.github/workflows/pipeline_swfs_test.yaml
pschoen-itsc: Thanks

experimental/seaweedfs/base/pipeline-profile-controller/sync.py
milinddethe15: workaround: https://www.diffchecker.com/h34UmQrY/

experimental/seaweedfs/base/seaweedfs/seaweedfs-deployment.yaml
```diff
@@ -17,16 +17,22 @@ spec:
     metadata:
       labels:
         app: seaweedfs
+        application-crd-id: kubeflow-pipelines
     spec:
       containers:
       - name: seaweedfs
-        image: 'chrislusf/seaweedfs:3.69'
+        image: 'chrislusf/seaweedfs:3.85'
```
milinddethe15, line 24: Can we use latest release v3.86? 
https://github.com/seaweedfs/seaweedfs/releases/

pschoen-itsc: I don't think so https://hub.docker.com/r/chrislusf/seaweedfs/tags
The Docker image is still missing. But this version does not seem the bring significant changes for our usecase, so 3.85 should be fine for now.

juliusvonkohout:
Amazing, thank you very much. It it such an improvement of the status quo.

In a follow up PR we should
1.  check whether ubuntu-latest-16-cores is really needed instead of the smaller nodes
2. test it also in https://github.com/kubeflow/manifests/blob/master/.github/workflows/full_kubeflow_integration_test.yaml
3. Se how we can upstream that to KFP

Thank you everyone, it started all in 2020 here https://github.com/kubeflow/pipelines/issues/4649 and went via https://github.com/kubeflow/pipelines/pull/7725 and https://github.com/kubeflow/manifests/pull/2826 (October 2024) to https://github.com/kubeflow/manifests/pull/3051, Without that experimental are and extended tests it would have been very hard to pull of and coordinate.

/lgtm
/approve

juliusvonkohout:
Tracked in https://github.com/kubeflow/manifests/issues/3119

PR #3052, opened 2025-03-14
Limit Istio Sidecar Scope to reduce memory and make cluster more scalable

han-steve:

Adding a new [istio sidecar resource](https://istio.io/latest/docs/reference/config/networking/sidecar/) to limit the sidecar's egress visibility to unnecessary services. 

We (Roblox) have been running Kubeflow in production for a long time, and we are noticing that the istio sidecar memory is almost 1GB now due to the amount of services in the cluster that has to be cached in each sidecar. This adds up to over 2 TB of memory in total. This change limits the caching of cluster services in each sidecar, thus helping the scalability of the cluster. 

This change can save TBs of memory and spare our DNS services. But I want to ask the community to see if there are any istio-enabled egress communication from kubeflow pods that we haven't considered. As far as we know
Communications to Notebook and Pipeline backends go through the ingress gateway instead of directly inside the cluster, so that won't matter
Communications to kserve models go through cluster ingress gateway
All other CRD-based workloads don't need any egress communication

https://github.com/knative/serving/issues/12917 We are facing this issue where each sidecar is pinging DNS to resolve the cluster ingressgateway ip, essentially DDOSing our DNS. Removing the ExternalName service for cluster ingress gateway from the sidecars would resolve this problem. 

## ✅ Contributor Checklist

---     
 
Slack message link: https://cloud-native.slack.com/archives/C073W572LA2/p1741893411623659


juliusvonkohout:
Thank you for the PR.

/ok-to-test

common/istio-1-24/istio-install/base/sidecar-prune-egress.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: networking.istio.io/v1alpha3
+kind: Sidecar
+metadata:
+  name: prune-egress
+spec:
+  egress:
+  - hosts:
+    - "./*" # use mTLS within the namespace
```
juliusvonkohout, line 8: Does that block stuff that is not on the servicemesh by default ? E.g. a deployment in the same namespace with istio-injection disabled. Or egress to a non-istio service in another namespaces. Because that must be still allowed.

juliusvonkohout: We have many mtls rules for the namespace "kubeflow". Please check the destination rules in the manifests first.

han-steve: These destination rules all point to other services in the "kubeflow" namespace, which this policy allows. If you want, we can also have this policy only apply to user namespaces (kubeflow-*) by using a kyverno policy

juliusvonkohout: There is "kubeflow", knative-serving, istio-system, oauth2-proxy etc whil the user namespaces are named arbitrarily. They could be thisismycoolusernamespace. Also kyverno is not yet part of the default platform and we need to work with the tools that are there by default.

"These destination rules all point to other services in the "kubeflow" namespace, which this policy allows" if that is true it might work. but please add a test in this PR here to verify the functionality (connection and mtls).

juliusvonkohout:
@tarekabouzeid do you mind testing this ?

juliusvonkohout:
> So we've actually been running this in production for quite a while. The only caveat is that the notebook controller cannot directly talk to notebooks in user namespaces to update the kernel activity. We fixed this by having another Sidecar CRD that only applies to the notebook controller and allow it to cache services in all namespaces

Do you mind providing implementation details ?

han-steve:
Sure. 
```
apiVersion: networking.istio.io/v1beta1
kind: Sidecar
metadata:
  name: istio-sidecar-prune-egress
  namespace: istio-system
spec:
  egress:
  - hosts:
    - ./*
    - ingress-nginx/*
    - ingress-nginx-serving/*
    - istio-system/*
    - kubeflow/*
    - kube-system/*
    - argocd/*
    - cert-manager/*
    - external-dns/*
    - monitoring/*
```
and
```
apiVersion: networking.istio.io/v1beta1
kind: Sidecar
metadata:
  name: istio-sidecar-prune-egress
  namespace: kubeflow
spec:
  egress:
  - hosts:
    - '*/*'
  workloadSelector:
    labels:
      app: notebook-controller
```

juliusvonkohout:
CC also @kimwnasptd @mvlassis since this could help with the performance as well. I am too busy right now, but will follow up.

juliusvonkohout:
closed in favor of https://github.com/kubeflow/manifests/pull/3206

PR #3064, opened 2025-03-20
Update trivy_scan.py

juliusvonkohout:
Tekton is unmaintained for a long time. so we remove it and add spark.


juliusvonkohout:
@tarekabouzeid for /lgtm

PR #3030, opened 2025-03-03
Update README.md

juliusvonkohout:
@tarekabouzeid @varodrig for /lgtm

PR #3080, opened 2025-03-30
Ensure manifests description is vendor-neutral

thesuperzapper:
## What this PR does

For the same reasons as discussed in https://github.com/kubeflow/website/pull/4071, it's is critical that the description of the manifests remain vendor-neutral and true to what we provide as a community.

This PR proposes an update to the manifests introduction paragraph that clarifies what the manifests actually are.

See [diff](xxxx) for the current proposal, but it updates to use the same wording as in https://github.com/kubeflow/website/pull/4071.

## Motivation

@kubeflow/kubeflow-steering-committee  Up until 30 days ago (PR: https://github.com/kubeflow/manifests/pull/3030), the main README of `kubeflow/manifests` did not include any statements that implied support for specific (kubernetes) distributions by the manifests.

However, the README now includes the following introduction:

> The **Kubeflow Manifests** are a collection of community-maintained manifests for installing Kubeflow in popular Kubernetes clusters such as Kind, Minikube, Rancher, EKS, AKS, and GKE. 
> The manifests include all Kubeflow components (Pipelines, Kserve, etc.), the **Kubeflow Central Dashboard**, and other applications that comprise the **Kubeflow Platform**.
> This installation is beneficial for users wanting to explore the end-to-end capabilities of the Kubeflow Platform.

The wording of the first sentence is misleading, as the manifests are currently only tested to deploy a minimum-viable Kubeflow Platform on Kind, with no guarantees for other platforms.

It's also important to note that the [Manifests WG charter](https://github.com/kubeflow/community/blob/master/wg-manifests/charter.md#out-of-scope) explicitly prohibits the inclusion of (kubernetes) distribution-specific manifests, so even mentioning specific distributions verges on violating the charter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,10 @@
 # Kubeflow Manifests
 
-The **Kubeflow Manifests** are a collection of community-maintained manifests for installing Kubeflow in popular Kubernetes clusters such as Kind, Minikube, Rancher, EKS, AKS, and GKE. The manifests include all Kubeflow components (Pipelines, Kserve, etc.), the **Kubeflow Central Dashboard**, and other applications that comprise the **Kubeflow Platform**. This installation is beneficial for users wanting to explore the end-to-end capabilities of the Kubeflow Platform.
+The Kubeflow Manifests are community maintained kustomize manifests which are tested to deploy a minimum-viable Kubeflow Platform on **[Kind](https://kind.sigs.k8s.io/)** clusters.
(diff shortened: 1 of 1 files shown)


juliusvonkohout:
You can install it on the popular Kubernetes clusters. Which is amazing and what many users do. Even if you rephrase it to "The Kubeflow Manifests are a collection of community-maintained manifests for installing Kubeflow on popular Kubernetes clusters  in the cloud an on-premises". That is also why the KSC cut down the text in other places. We are a best-effort open-source project by default, no need present ourseve worse than we are.

PR #3056, opened 2025-03-18
Document secure Kserve authentication via automated tests

madmecodes:
# Pull Request Template for Kubeflow Manifests

## ✏️ Summary of Changes
This PR addresses a security issue with KServe inference endpoints by applying oauth2-proxy authentication to the cluster-local-gateway. Currently, KServe endpoints lack proper authentication because they use the cluster-local-gateway which has an `ALLOW` policy instead of proper oauth2-proxy authentication.

Changes include:
1. Modified the cluster-local-gateway AuthorizationPolicy to use oauth2-proxy authentication instead of allowing all traffic without authentication
2. Updated the KServe test workflow to:
   - Use proper Kubeflow user namespaces instead of generic test namespaces
   - Test with tokens from the appropriate service accounts
   - Apply predictor-specific AuthorizationPolicy for tests

This ensures KServe inference endpoints are properly secured and require authentication, just like other Kubeflow services accessed through the istio-ingressgateway.

## 📦 Dependencies
No direct dependencies on other PRs. This PR combines improvements that were planned for PR #2936 with the security fix.

## 🐛 Related Issues
- Fixes #2811 (Secure KServe inferencing with proper authentication)
- Related to #2936 (Update KServe tests with real Kubeflow namespaces)
- Helps unblock #2907 (ISTIO CNI by default)

## ✅ Contributor Checklist
  - [x] I have tested these changes with kustomize. See [Installation Prerequisites](https://github.com/kubeflow/manifests#prerequisites).
  - [x] All commits are [_signed-off_](https://github.com/kubeflow/community/tree/master/dco-signoff-hook#signing-off-commits) to satisfy the DCO check.

---     
 
> You can join the CNCF Slack and access our meetings at the [Kubeflow Community](https://www.kubeflow.org/docs/about/community/) website. Our channel on the CNCF Slack is here [**#kubeflow-platform**](https://app.slack.com/client/T08PSQ7BQ/C073W572LA2).

PR #3082, opened 2025-04-02
Fix: flaky Dex login test by improving authentication flow handling

madmecodes:
## ✏️ Summary of Changes

1. Make the Dex test more robust (fix flakiness)

## Solution:

**403 Handling - When receiving a 403 Forbidden response during login, we attempt to restart the OAuth flow, which helps when sessions expire.**

## Reproduce the issue:
```
# Run multiple concurrent requests to increase system load
kubectl run load-test --image=busybox --restart=Never -- /bin/sh -c "while true; do wget -q -O- http://localhost:8080; done"
```

You will get this:
Screenshot 2025-04-02 at 10 54 44 AM

After implementing 403 handling:
```
# Handle 403 specifically - might need to restart oauth flow
          if resp.status_code == 403:
                    # Let's try one more approach - go through the oauth2 flow again
                    oauth_url = f"{urlsplit(self._endpoint_url).scheme}://{urlsplit(self._endpoint_url).netloc}/oauth2/start"
                    resp = s.get(
                        oauth_url,
                        headers=headers,
                        allow_redirects=True,
                        verify=not self._skip_tls_verify,
                        timeout=30
                    )
                    
                    # Continue with normal flow after restart
                    if resp.status_code == 200:
                        # If we have cookies now, we're good
                        if s.cookies:
                            return "; ".join([f"{c.name}={c.value}" for c in s.cookies])
```

Successfully passed, after main 403 retry. (load test pod still running and if i comment out the 403 mechanism, it again throws the error)

Screenshot 2025-04-02 at 11 00 38 AM

## 📦 Dependencies
Asked to resolve the issue in [PR 3056](https://github.com/kubeflow/manifests/pull/3056#issuecomment-2752093237)

## 🐛 Related Issues
none

## ✅ Contributor Checklist

---     
 
### DIFF
--- a/tests/gh-actions/test_dex_login.py
+++ b/tests/gh-actions/test_dex_login.py
@@ -52,94 +52,155 @@ class DexSessionManager:
         Get the session cookies by authenticating against Dex
         :return: a string of session cookies in the form "key1=value1; key2=value2"
         """
+        max_retries = 3
+        retry_delay = 2
+        
+        for attempt in range(max_retries):
+            try:
+                # use a persistent session (for cookies)
+                s = requests.Session()
+                
+                # Add user-agent to avoid some security blocks or {} brackets also worked
+                headers = {
+                    'User-Agent': 'Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36'
+                }
 
-        # use a persistent session (for cookies)
-        s = requests.Session()
-
-        # GET the endpoint_url, which should redirect to Dex
-        resp = s.get(
-            self._endpoint_url, allow_redirects=True, verify=not self._skip_tls_verify
-        )
-        if resp.status_code == 200:
-            pass
-        elif resp.status_code == 403:
-            # if we get 403, we might be at the oauth2-proxy sign-in page
-            # the default path to start the sign-in flow is `/oauth2/start?rd=<url>`
-            url_obj = urlsplit(resp.url)
-            url_obj = url_obj._replace(
-                path="/oauth2/start", query=urlencode({"rd": url_obj.path})
-            )
-            resp = s.get(
-                url_obj.geturl(), allow_redirects=True, verify=not self._skip_tls_verify
-            )
-        else:
-            raise RuntimeError(
-                f"HTTP status code '{resp.status_code}' for GET against: {self._endpoint_url}"
-            )
-
-        # if we were NOT redirected, then the endpoint is unsecured
-        if len(resp.history) == 0:
-            # no cookies are needed
-            return ""
-
-        # if we are at `../auth` path, we need to select an auth type
-        url_obj = urlsplit(resp.url)
-        if re.search(r"/auth$", url_obj.path):
-            url_obj = url_obj._replace(
-                path=re.sub(r"/auth$", f"/auth/{self._dex_auth_type}", url_obj.path)
-            )
-
-        # if we are at `../auth/xxxx/login` path, then we are at the login page
-        if re.search(r"/auth/.*/login$", url_obj.path):
-            dex_login_url = url_obj.geturl()
-        else:
-            # otherwise, we need to follow a redirect to the login page
-            resp = s.get(
-                url_obj.geturl(), allow_redirects=True, verify=not self._skip_tls_verify
-            )
-            if resp.status_code != 200:
-                raise RuntimeError(
-                    f"HTTP status code '{resp.status_code}' for GET against: {url_obj.geturl()}"
+                # GET the endpoint_url, which should redirect to Dex
+                resp = s.get(
+                    self._endpoint_url, 
+                    headers=headers,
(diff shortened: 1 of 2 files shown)


madmecodes:
@juliusvonkohout looking forward for the feedback

madmecodes:
> @juliusvonkohout looking forward for the feedback

The key addition is just the 403 handling block after the login POST request, which restarts the OAuth flow. The retry logic is also minimal but effective.

tests/gh-actions/test_dex_login.py
```diff
@@ -52,94 +52,125 @@ def get_session_cookies(self) -> str:
...
+        max_retries = 3
+        retry_delay = 2
+        
+        for attempt in range(max_retries):
+            try:
+                # use a persistent session (for cookies)
+                s = requests.Session()
+
+                # GET the endpoint_url, which should redirect to Dex
+                resp = s.get(
```
juliusvonkohout, line 64: Lest use clean coding standards, so proper long pronounceable variable names. session instead of s, response instead of resp, url_object instead of url_obj

tests/gh-actions/test_dex_login.py
```diff
@@ -52,94 +52,125 @@ def get_session_cookies(self) -> str:
...
+                    )
+                    if resp.status_code != 200:
+                        raise RuntimeError(
+                            f"HTTP status code '{resp.status_code}' for POST against: {url_obj.geturl()}"
+                        )
+
+                return "; ".join([f"{c.name}={c.value}" for c in s.cookies])
+                
+            except Exception as e:
+                if attempt < max_retries - 1:
```
juliusvonkohout, line 164: Why do you use -1  instead of <= ?

tests/gh-actions/test_dex_login.py
madmecodes: okay i will be more verbose

tests/gh-actions/test_dex_login.py
madmecodes: I used attempt < max_retries - 1 to ensure we retry only up to the second-to-last attempt (e.g., attempts 0 and 1 for max_retries = 3). 

On the final attempt (attempt 2), if it fails, we fall into the else block to raise the exception without an extra retry. 

Using <= max_retries would imply retrying beyond the intended limit since attempt only goes up to max_retries - 1 in the loop.

attempt <= max_retries would be incorrect since attempt never reaches max_retries in a range(max_retries)=> (0,1,2) and maxRetry=3 loop.

I will update the code, for more readability

tests/gh-actions/test_dex_login.py
```diff
@@ -52,94 +52,128 @@ def get_session_cookies(self) -> str:
...
+                # Handle 403 specifically - might need to restart oauth flow
+                if response.status_code == 403:
+                    # Try one more approach - go through the oauth2 flow again
+                    oauth_url = f"{urlsplit(self._endpoint_url).scheme}://{urlsplit(self._endpoint_url).netloc}/oauth2/start"
+                    response = session.get(
+                        oauth_url,
+                        allow_redirects=True,
+                        verify=not self._skip_tls_verify,
+                    )
+                    # Continue with normal flow after restart
```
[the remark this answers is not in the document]
juliusvonkohout: Why does a bit of load fail the test so easily ?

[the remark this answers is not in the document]
madmecodes: That bit of load did fail the test at that time, but when I checked again the next day, it was able to handle it. So I deliberately triggered a 403 using this code block, and as seen in the output, the old version exited with the error, while the new one worked.


```
def patched_post(original_post):
    def _patched_post(*args, **kwargs):
        if random.random() < 0.2:
            mock_response = requests.Response()
            mock_response.status_code = 403
            mock_response._content = b"Simulated 403 error"
            mock_response.url = args[0]
            return mock_response
        return original_post(*args, **kwargs)
    return _patched_post

original_post = requests.Session.post
requests.Session.post = patched_post(original_post)
```
Screenshot 2025-04-13 at 5 50 33 PM

**You're right that this is more of a workaround than a solution to the underlying problem.** The 403 errors happen when the Dex authentication flow gets disrupted under load. While this fix makes the test more reliable, we should investigate why these errors occur in the first place. 

The error occurs randomly and is a bit hard to reproduce. But yes, when it fails, the same error occurs — a 403 against the URL.

Shall I investigate the logs of these pods and run it again to find the root cause?

kubectl logs -n auth $(kubectl get pods -n auth -l app=dex -o name) -f
kubectl logs -n oauth2-proxy $(kubectl get pods -n oauth2-proxy -o name) -f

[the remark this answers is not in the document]
juliusvonkohout: Yes please investigate, but i will merge it for now

PR #3134, opened 2025-05-16
istio namespace missing

juliusvonkohout:
@tarekabouzeid 

see https://github.com/kubeflow/manifests/blob/533ca9f8c1c97836fae6f5946c76ce31f42dd2c9/common/istio-1-24/istio-install/base/kustomization.yaml#L4


juliusvonkohout:
needs a second round fir the istio-cni daemonset namespace

tarekabouzeid:
> needs a second round fir the istio-cni daemonset namespace

Yeah, i think we can add labels to it , so its set to a more permissive level

Issue #3061, opened 2025-03-20
Istio CNI by default


juliusvonkohout:
### Validation Checklist

- [x] I confirm that this is a Kubeflow-related issue.
- [x] I am reporting this in the appropriate repository.
- [x] I have followed the [Kubeflow installation guidelines](https://github.com/kubeflow/manifests/blob/master/README.md).
- [x] The issue report is detailed and includes version numbers where applicable.
- [x] This issue pertains to Kubeflow development.
- [x] I am available to work on this issue.
- [x] You can join the CNCF Slack and access our meetings at the [Kubeflow Community](https://www.kubeflow.org/docs/about/community/) website. Our channel on the CNCF Slack is here [**#kubeflow-platform**](https://app.slack.com/client/T08PSQ7BQ/C073W572LA2).

### Version

master

### Detailed Description

Istio CNI by default

### Steps to Reproduce

Istio CNI by default

juliusvonkohout:
After fixing the linked issues I just need to switch all workflows from tests/gh-actions/install_istio.sh to tests/gh-actions/install_istio-cni.sh. Then check whether there are problems on common clusters.

juliusvonkohout:
Part of rootless kubeflow https://github.com/kubeflow/manifests/issues/2528

juliusvonkohout:
https://github.com/kubeflow/manifests/pull/3062 is just the first part.

juliusvonkohout:
Check the GCP hostpath permissions

Issue #2528, opened 2023-09-13
Rootless Kubeflow


juliusvonkohout:
# Rootless Kubeflow

This is an issue to track the progress of the following proposal https://github.com/kubeflow/manifests/blob/master/proposals/20200913-rootlessKubeflow.md.
### Goals

We want to run Kubeflow as rootless as possible according to CNCF/Kubernetes best practices.
Most enterprise environments will require this as well.

### Implementation details
The main steps are adding an additional profile for istio-cni and later ambient mesh, updating the documentation and manifest generation process.
Only istio-cni or istio ambient mesh can run rootless as explained here https://istio.io/latest/docs/setup/additional-setup/cni/. 
Istio-cni will still nedd a deamonset in kube-system, but that is completly isolated from user workloads. 
The ambient mesh should get rid of this as well and also has the benefit of removing the istio initcontainers and sidecars altogether.
Then adding the baseline and restricted PSS as kustomize component to /contrib and extending the profile controller to annotate user namespaces with configurable PSS labels.

We want to use a staged approach.

First Stage:
1. Implement Istio 1.17.5 and use it by default, because 1.17. is what we have planned to use for Kubeflow 1.8.
2. Implement istio-cni (--set components.cni.enabled=true --set components.cni.namespace=kube-system) as second option.
3. Add simple tests similar to tests/gh-actions/install_istio.sh and tests/gh-actions/install_knative.sh for istio-cni and support both rootfull and rootless istio at the same time and give users one release to test

Second stage:
4. Add pod security standards (https://kubernetes.io/docs/concepts/security/pod-security-standards/) base/restricted to manifests/contrib
5. Enforce PSS baseline (here you can still build OCI containers via Podman and buildah). It works with any istio
6. Enable Warnings for violations of restricted PSS
7. Add tests to make sure that the PSS are used and tested in the CICD
8. Optionally Enforce PSS restricted (this is where minor corner cases are affected)

Third stage:
9. Upgrade Istio to 1.19 to make the ambient mesh available
10. Add istio-ambient as an option to the next Kubeflow release.

fourth stage:
11. Use the ambient service mesh by default in Kubeflow 1.10.

### Non-Goals
This does not cover Application level CVEs, only cluster level security.

### Does this break any existing functionality?
So far not. Only PSS restricted may block the incredibly dangerous and unprofessional Docker in Docker.
This is a rarely used feature from the KFP SDK.
With PSS baseline you can still build OCI images with Podman for example. 
[...]

PR #3129, opened 2025-05-11
Updated the V2 pipeline test  to be more robust

kunal-511:
# Pull Request Template for Kubeflow Manifests

## ✅ Contributor Checklist
  - [ ] I have tested these changes with kustomize. See [Installation Prerequisites](https://github.com/kubeflow/manifests#prerequisites).
  - [x] All commits are [_signed-off_](https://github.com/kubeflow/community/tree/master/dco-signoff-hook#signing-off-commits) to satisfy the DCO check.
  - [ ] I have considered adding my company to the adopters page to support Kubeflow and help the community, since I expect help from the community for my issue (see [1.](https://github.com/kubeflow/community/issues/833) and [2.](https://github.com/kubeflow/community/blob/master/ADOPTERS.md#adopters-of-kubeflow-platform)).

---     
 
> You can join the CNCF Slack and access our meetings at the [Kubeflow Community](https://www.kubeflow.org/docs/about/community/) website. Our channel on the CNCF Slack is here [**#kubeflow-platform**](https://app.slack.com/client/T08PSQ7BQ/C073W572LA2).

PR #3135, opened 2025-05-17
Switch to Istio CNI by default

madmecodes:
Switch to Istio CNI by default

This PR changes the default Istio installation to use Istio CNI instead of standard Istio.
Key benefits include:
- Eliminates the need for privileged Istio init containers
- Improves compatibility with Pod Security Standards (PSS)
- Enables native sidecars support for better init container network access

Changes include:
- Updated example/kustomization.yaml to use Istio CNI paths
- Updated README.md to reflect new installation paths
- Added note about Ray operator being configured for Istio CNI compatibility

This change is part of the broader Rootless Kubeflow initiative https://github.com/kubeflow/manifests/issues/2528
and follows up on previous work https://github.com/kubeflow/manifests/issues/3061.


madmecodes:
@juliusvonkohout For GCP hostPath check, https://github.com/kubeflow/manifests/blob/9825950c84fd5d29617bc8ad6c0d15a6432e7635/common/istio-cni-1-24/istio-install/base/install.yaml#L3009

am i suppose to spin the cluster in GCP/GKE ad check the paths? will that cost me?

juliusvonkohout:
> @juliusvonkohout For GCP hostPath check,
> 
> https://github.com/kubeflow/manifests/blob/9825950c84fd5d29617bc8ad6c0d15a6432e7635/common/istio-cni-1-24/istio-install/base/install.yaml#L3009
> 
> am i suppose to spin the cluster in GCP/GKE ad check the paths? will that cost me?

I know the paths and that it is correct for gcp. The question is how we can enable both cni directories at the same time. Since the default ones are correct for Azure, Kind and others.

juliusvonkohout:
See https://github.com/kubeflow/manifests/issues/3061#issuecomment-2813289604. We need to support both paths at the same time somehow.

madmecodes:
Okay, looking into this

> See [#3061 (comment)](https://github.com/kubeflow/manifests/issues/3061#issuecomment-2813289604). We need to support both paths at the same time somehow.

madmecodes:
Why This Multi-Path Approach woukd Work?

The patch adds both the standard path (/opt/cni/bin) and the GCP-specific path (/home/kubernetes/bin) as separate volume mounts in the Istio CNI DaemonSet.

1. The CNI installer in the container will attempt to install the CNI binary to both paths
2. Regardless of whether the node is running on standard Kubernetes or GCP/GKE, one of the paths will exist and be the correct one.
3. If a path doesn't exist, mounting a non-existent hostPath simply results in an empty directory, which causes no harm.

@juliusvonkohout what do you think, can this be a potential approach?

juliusvonkohout:
> Why This Multi-Path Approach woukd Work?
> 
> The patch adds both the standard path (/opt/cni/bin) and the GCP-specific path (/home/kubernetes/bin) as separate volume mounts in the Istio CNI DaemonSet.
> 
>     1. The CNI installer in the container will attempt to install the CNI binary to both paths
> 
>     2. Regardless of whether the node is running on standard Kubernetes or GCP/GKE, one of the paths will exist and be the correct one.
> 
>     3. If a path doesn't exist, mounting a non-existent hostPath simply results in an empty directory, which causes no harm.
> 
> 
> @juliusvonkohout what do you think, can this be a potential approach?

Yes, could work. Do you mind testing it on GCP? I think there is a small free 4GB node available by default if you have a gmail adress. You jus tneed to install Istio, so it should be enough. I can then later also test on some GCP clusters.

juliusvonkohout:
Do you mind fixing  python3: can't open file '/home/runner/work/manifests/manifests/tests/gh-actions/test_pipeline.py': [Errno 2] No such file or directory in https://github.com/kubeflow/manifests/actions/runs/15084697684/job/42405783624?pr=3135 in a separate PR ? I think the file has just been renamed since we have v1 and v2 kfp tests. CC @kunal-511 to help

kunal-511:
@madmecodes The test_pipeline.py has been changed to test_pipeline_v2.py in #3129

madmecodes:
> @madmecodes The test_pipeline.py has been changed to test_pipeline_v2.py in #3129

ok let me rebase

> The test_pipeline.py has been changed to test_pipeline_v2.py in #3129

this is updated https://github.com/kubeflow/manifests/pull/3136

madmecodes:
# GKE Istio CNI Multi-Path Testing Report

## Problem Statement
Istio CNI fails on Google Kubernetes Engine (GKE) because `/opt/cni/bin` is mounted read-only, preventing the CNI installer from writing the `istio-cni` binary.

## Solution Approach
Implemented multi-path support by mounting both standard (`/opt/cni/bin`) and GCP-specific (`/home/kubernetes/bin`) CNI binary directories.

## Testing Environment
- **Platform**: Google Kubernetes Engine (GKE)
- **Cluster**: `istio-cni-test` (2 nodes, e2-standard-2)
- **Kubernetes**: v1.32.3-gke.1927009
- **Istio CNI**: 1.24.3

## Commands Used and Results

### 1. Cluster Setup
```bash
gcloud container clusters create istio-cni-test \
  --zone us-central1-a \
  --num-nodes 2 \
  --machine-type e2-standard-2 \
  --disk-size 20GB
```

### 2. Install Kubeflow with Istio CNI
```bash
while ! kustomize build example | kubectl apply --server-side --force-conflicts -f -; do 
  echo "Retrying..."; sleep 10; 
done
```

### 3. Observed Failure
```bash
kubectl logs -n kube-system istio-cni-node-8fc7j
```

**Output:**
```
CNIBinTargetDirs: /host/opt/cni/bin
error: failed file copy of /opt/cni/bin/istio-cni to /host/opt/cni/bin: 
open /host/opt/cni/bin/istio-cni.tmp.586267941: read-only file system
```

### 4. Verified Multi-Path Volume Mounts
```bash
kubectl describe pod -n kube-system istio-cni-node-8fc7j
```

**Output:**
```
Mounts:
  /host/opt/cni/bin from cni-bin-dir (rw)
  /host/home/kubernetes/bin from cni-bin-dir-gcp (rw)

Volumes:
  cni-bin-dir:
    Type: HostPath (bare host directory volume)
    Path: /opt/cni/bin
  cni-bin-dir-gcp:
    Type: HostPath (bare host directory volume)  
    Path: /home/kubernetes/bin
```

### 5. Attempted Configuration Fix
```bash
kubectl patch daemonset istio-cni-node -n kube-system --type='json' \
  -p='[{"op": "add", "path": "/spec/template/spec/containers/0/env/-", 
       "value": {"name": "CNI_BIN_TARGET_DIRS", 
                "value": "/host/opt/cni/bin,/host/home/kubernetes/bin"}}]'
```

**Result:** Environment variable set but ignored by CNI installer.

### 6. Verified GCP Path Functionality
```bash
kubectl debug node/gke-istio-cni-test-default-pool-6c013e50-k2bx -it --image=busybox
```

**Commands in debug pod:**
```bash
ls -la /host/home/kubernetes/bin/    # Shows CNI binaries present
touch /host/home/kubernetes/bin/test-file && rm /host/home/kubernetes/bin/test-file
echo "GCP path is writable"
```

**Output:** GCP path is writable ✓

## Key Findings

### ✅ What Works
1. **Multi-path volume mounts**: Both `/opt/cni/bin` and `/home/kubernetes/bin` successfully mounted
2. **GCP path accessibility**: `/home/kubernetes/bin` is writable and contains CNI binaries
3. **Infrastructure setup**: All volume configurations applied correctly

### ❌ Current Limitation
**Istio CNI installer hardcoded behavior**: Despite environment variable `CNI_BIN_TARGET_DIRS`, installer only attempts standard path:
```
CNIBinTargetDirs: /host/opt/cni/bin  (single path only)
```

## Root Cause Analysis
- GKE mounts `/opt/cni/bin` as read-only for security
- Istio CNI installer lacks native multi-directory support
- Installer ignores `CNI_BIN_TARGET_DIRS` environment variable
```

juliusvonkohout:
/lgtm
/approve

thank you

PR #3165, opened 2025-06-11
Update install_KinD_create_KinD_cluster_install_kustomize.sh

juliusvonkohout:
@tarekabouzeid for /lgtm


juliusvonkohout:
It should allow us to use https://kubernetes.io/blog/2025/04/25/userns-enabled-by-default/ / https://kubernetes.io/docs/concepts/workloads/pods/user-namespaces/#integration-with-pod-security-admission-checks

PR #3190, opened 2025-07-21
PSS restricted for static namespaces

juliusvonkohout:
@tarekabouzeid


juliusvonkohout:
@akagami-harsh can you check why this is failing ?

juliusvonkohout:
Interesting, so only Ray is failing if we retry
image

juliusvonkohout:
I am wondering whether the securitycontext was cont completely upstreamed https://github.com/kubeflow/manifests/commit/b99375fd999fd44de27f6f709a963908a0042430

juliusvonkohout:
@kunal-511 the ray test is missing from https://github.com/kubeflow/manifests/blob/master/.github/workflows/full_kubeflow_integration_test.yaml, that is why PSS restricred problems were not detected before.

PR #3221, opened 2025-08-19
Update kserve_test.yaml

juliusvonkohout:
I thought this is covered in https://github.com/kubeflow/manifests/blob/master/.github/workflows/kserve_secure_test.yaml @madmecodes


juliusvonkohout:
@kunal-511 can you help investigating ?

PR #3238, opened 2025-09-09
fix: override dex deployment with custom image and env specific for c42 env

Avnshrai:
### Summary
- Added new overlay patch (`patch-deployment.yaml`) in `common/dex/overlays/oauth2-proxy` to:
  - Override Dex container image with `avnshrai/dex-with-ca:v1`
  - Add `SSL_CERT_FILE=/etc/ssl/certs/ca-certificates.crt`

- Updated `kustomization.yaml` in the same overlay to include this patch.

### Why
- Ensures Dex uses the custom image and required SSL_CERT_FILE env var across all installs.
- Prevents drift when uninstalling/reinstalling Kubeflow.

### DIFF
--- a/common/dex/overlays/oauth2-proxy/kustomization.yaml
+++ b/common/dex/overlays/oauth2-proxy/kustomization.yaml
@@ -11,4 +11,10 @@ patches:
       kind: ConfigMap
       name: dex
       namespace: auth
+  - path: patch-deployment.yaml
+    target:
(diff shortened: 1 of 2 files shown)


juliusvonkohout:
Hello, i do not understand what you want exactly (why do you want to add more certificates in DEX) and it looks very dangerous to add an image that is not from the official vendor, so dex itself.

PR #3262, opened 2025-10-23
Tests for Model Catalog

kunal-511:

## ✅ Contributor Checklist
  - [ ] I have tested these changes with kustomize. See [Installation Prerequisites](https://github.com/kubeflow/manifests#prerequisites).
  - [ ] All commits are [_signed-off_](https://github.com/kubeflow/community/tree/master/dco-signoff-hook#signing-off-commits) to satisfy the DCO check.
  - [ ] I have considered adding my company to the adopters page to support Kubeflow and help the community, since I expect help from the community for my issue (see [1.](https://github.com/kubeflow/community/issues/833) and [2.](https://github.com/kubeflow/community/blob/master/ADOPTERS.md#adopters-of-kubeflow-platform)).

---     
 
### DIFF
--- a/.github/workflows/full_kubeflow_integration_test.yaml
+++ b/.github/workflows/full_kubeflow_integration_test.yaml
@@ -109,6 +109,9 @@ jobs:
       run: |
         kustomize build applications/model-registry/upstream/options/ui/overlays/istio | kubectl apply -n kubeflow -f -
 
+    - name: Install KF Model Catalog
+      run: ./tests/model_catalog_install.sh
+
     - name: Install Spark
       run: chmod u+x tests/*.sh && ./tests/spark_install.sh
 
@@ -230,6 +233,9 @@ jobs:
         curl -s --fail \
           "localhost:8080/model-registry/api/v1/model_registry?namespace=${KF_PROFILE}" \
           -H "Authorization: Bearer ${KF_TOKEN}"
+
+    - name: Run Model Catalog API tests
+      run: ./tests/model_catalog_test.sh
 
     - name: Test Model Registry API with Unauthorized Token
       run: |


.github/workflows/model_catalog_test.yaml
```diff
@@ -0,0 +1,71 @@
...
+    - name: Run Model Catalog API tests
+      run: ./tests/model_catalog_test.sh
+
+    - name: Test Model Catalog API endpoint
+      run: |
+        nohup kubectl port-forward svc/model-catalog-model-catalog-service -n kubeflow 8082:8080 &
+        sleep 5
+        curl -v -X 'GET' \
+          'http://localhost:8082/api/model_catalog/v1alpha1/models?pageSize=100' \
+          -H 'accept: application/json' || echo "Note: Empty catalog is expected for basic test"
```
tarilabs: not sure if this last step is needed, given tests/model_catalog_test.sh already curl on this API endpoint?

kunal-511: You're absolutely right! The last step is redundant

juliusvonkohout:
Should we just merge this with the model registry files or do you want to have it explicitly separate @tarilabs ?

tarilabs:
> Should we just merge this with the model registry files or do you want to have it explicitly separate @tarilabs ?

I'd prefer to keep these separate GHA and scripts, if that is okay with you, as we'd like to showcase Catalog also independently of Registry. Further will help pin-point if either fails their check. Wdyt?

tests/model_catalog_install.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -euxo pipefail
+
+cd applications/model-registry/upstream/options/catalog && kustomize build . | kubectl apply -n kubeflow -f - && cd ../../../../../
```
pboyd: Minor thing, but a sub-shell would save you from counting directories:

```suggestion
(
    cd applications/model-registry/upstream/options/catalog
    kustomize build . | kubectl apply -n kubeflow -f -
)
```

Also, when we bring the changes from upstream, this path will need to be `applications/model-registry/upstream/options/catalog/base` (unless you want the istio components, in which case it's `applications/model-registry/upstream/options/catalog/overlay`).

pboyd (approved):
Looks good to me.

Fair warning though, we've tweaked a few things upstream and we'll need some changes to this test when we bring those changes back down. Namely, we fixed some of the redundant names and added an optional istio overlay.

juliusvonkohout:
Strange, i see 
```
{"level":"info","ts":"2025-10-24T13:50:52.002205822Z","logger":"setup","caller":"trainer-controller-manager/main.go:93","msg":"Loading configuration","configFile":""}
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x13def43]

goroutine 1 [running]:
github.com/kubeflow/trainer/v2/pkg/config.addTo(0xc0003b6fd0, 0xc000231040, 0x40?)
	/workspace/pkg/config/config.go:71 +0x1c3
github.com/kubeflow/trainer/v2/pkg/config.Load(_, {_, _}, _)
	/workspace/pkg/config/config.go:128 +0x216
main.main()
	/workspace/cmd/trainer-controller-manager/main.go:94 +0x3a5
```
in kubeflow-trainer-controller-manager in kubeflow-system
probably because of https://github.com/kubeflow/manifests/blob/65245b29afe55d5d7b699fdf33903b32e21db9f5/applications/trainer/upstream/overlays/manager/kustomization.yaml#L18

juliusvonkohout:
/lgtm
/unhold
thank you

PR #3268, opened 2025-11-04
Fix kustomize warnings

Misha6Sharma:
This PR updates several deprecated Kustomize fields across multiple components to reduce warnings during kustomize build
Changes included in the following field:
Replacing commonLabels with labels
Replacing bases with resources
Replacing patchesStrategicMerge with patches
I didn’t convert vars to replacements yet since that feature is still experimental and could affect how the manifests build.

### DIFF
--- a/applications/admission-webhook/upstream/base/kustomization.yaml
+++ b/applications/admission-webhook/upstream/base/kustomization.yaml
@@ -8,11 +8,6 @@ resources:
 - service-account.yaml
 - service.yaml
 - crd.yaml
-commonLabels:
-  app: poddefaults
-  kustomize.component: poddefaults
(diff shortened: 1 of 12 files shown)


juliusvonkohout:
Hello @Misha6Sharmathese and thank you for the PR. The files are synchronized here via scripts. You need to fix it upstream in kubeflow/dashboard, kubeflow/pipelines kubeflow/katib etc. to change the source. But you can use this PR to test your upstream changes platform wide.

Misha6Sharma:
Thanks for clarifying , I didnt know these files are synced from the upstream repos
Will move the updates to the respective repo and will use the same to verify

PR #3282, opened 2025-11-24
Synchronize trainer manifests v2.1.0

juliusvonkohout:
@andreyvelich @yiminghub2024 please review.

### DIFF
--- a/scripts/synchronize-trainer-manifests.sh
+++ b/scripts/synchronize-trainer-manifests.sh
@@ -9,7 +9,7 @@ setup_error_handling
 COMPONENT_NAME="trainer"
 REPOSITORY_NAME="kubeflow/trainer"
 REPOSITORY_URL="https://github.com/kubeflow/trainer.git"
-COMMIT="f12a6d3"
+COMMIT="v2.1.0"
 REPOSITORY_DIRECTORY="trainer"
 SOURCE_DIRECTORY=${SOURCE_DIRECTORY:=/tmp/kubeflow-${COMPONENT_NAME}}
 BRANCH_NAME=${BRANCH_NAME:=synchronize-${COMPONENT_NAME}-manifests-${COMMIT?}}


applications/trainer/upstream/overlays/data-cache/kustomization.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+# Patch namespace-rbac resources to install in default namespace
+patches:
+  - target:
+      kind: ServiceAccount
+      name: kubeflow-trainer-cache-initializer
+    patch: |-
+      - op: add
+        path: /metadata/namespace
+        value: default
```
juliusvonkohout, line 25: That is a hard blocker from the security perspective. Why do you want the default namespace ?

applications/trainer/upstream/overlays/data-cache/cluster_role.yaml
juliusvonkohout, line 1: Should we not use kubeflow-edit and default-editor?

juliusvonkohout: i think this is just an example that we do not install by default

juliusvonkohout:
@andreyvelich do we still need the patches from here https://github.com/kubeflow/manifests/blob/cab3e83e4ee54cd31e41940c084908189d33da39/applications/trainer/overlays/kustomization.yaml#L8-L48 or are they already upstreamed ?

andreyvelich:
> securityContext

SecurityContext should be already there: https://github.com/kubeflow/trainer/blob/master/manifests/base/manager/manager.yaml#L26
I am not sure why do you need `traffic.sidecar.istio.io/excludeInboundPort` ?

andreyvelich:
Can we also update ClusterRoleBinding to allow everyone access ClusterTrainingRuntime, as we discussed here: https://cloud-native.slack.com/archives/C073N7BMLB1/p1764001097918489?thread_ts=1763585362.957169&cid=C073N7BMLB1

cc @kimwnasptd

yiminghub2024:
i think maybe trainer and trainer-operator leave one in manifest is ok , because use one key install ,it will only install trainer , many users will very confused why have two in list but only install one ?  what's the differents between trainer-controller and trainer-operator?

yiminghub2024:
another suggestion is  in manifest readme.md   component list  version do not have f12a6d3 like this ,it need display exactly version like v2.0.1 or many users do not know what's the version ,they need use docker images to check the install version ,its very confused ,thanks

juliusvonkohout:
> Can we also update ClusterRoleBinding to allow everyone access ClusterTrainingRuntime, as we discussed here: https://cloud-native.slack.com/archives/C073N7BMLB1/p1764001097918489?thread_ts=1763585362.957169&cid=C073N7BMLB1
> 
> cc @kimwnasptd

Yes, but that is something for a seperate PR.

/approve

juliusvonkohout:
> > securityContext
> 
> SecurityContext should be already there: https://github.com/kubeflow/trainer/blob/master/manifests/base/manager/manager.yaml#L26 I am not sure why do you need `traffic.sidecar.istio.io/excludeInboundPort` ?

If there is an istio sidecar this will help. We need to check whether it is still needed in a separate PR.

juliusvonkohout:
> > > securityContext
> > 
> > 
> > SecurityContext should be already there: https://github.com/kubeflow/trainer/blob/master/manifests/base/manager/manager.yaml#L26 I am not sure why do you need `traffic.sidecar.istio.io/excludeInboundPort` ?
> 
> If there is an istio sidecar this will help. We need to check whether it is still needed in a separate PR.

@kunal-511 can you look into that. Nevertheless we should still merge this PR here.

PR #3283, opened 2025-11-25
Synchronize kubeflow pipelines manifests 2.15.0

juliusvonkohout:
@akagami-harsh @pschoen-itsc please review.

I had to manually remove applications/pipeline/upstream/base/pipeline/allow-same-namespace-networkpolicy.yaml
and modify applications/pipeline/upstream/base/pipeline/kustomization.yaml to get rid of the duplicated networkpolicy. allow-same-namespace-networkpolicy.yaml Should probably only be in the KFP tests but not in the release manifests.


juliusvonkohout:
SO lets merge this one here first, but @pschoen-itsc or @akagami-harsh  can you take care of removing the duplicated networkpolicy from upstream KFP?

akagami-harsh:
> SO lets merge this one here first, but @pschoen-itsc or @akagami-harsh can you take care of removing the duplicated networkpolicy from upstream KFP?

opened a pr https://github.com/kubeflow/pipelines/pull/12492

PR #3319, opened 2026-01-05
fix: variable namespaces for networkpolicies

juhyeon-cha:
# Pull Request Template for Kubeflow Manifests

## ✏️ Summary of Changes

The problem is that all NetworkPolicies are being created in the kubeflow namespace because the `namespace` value is present in kustomization.yaml.

## ✅ Contributor Checklist
  - [x] I have tested these changes with kustomize. See [Installation Prerequisites](https://github.com/kubeflow/manifests#prerequisites).
  - [x] All commits are [_signed-off_](https://github.com/kubeflow/community/tree/master/dco-signoff-hook#signing-off-commits) to satisfy the DCO check.
  - [ ] I have considered adding my company to the adopters page to support Kubeflow and help the community, since I expect help from the community for my issue (see [1.](https://github.com/kubeflow/community/issues/833) and [2.](https://github.com/kubeflow/community/blob/master/ADOPTERS.md#adopters-of-kubeflow-platform)).

---     
 
> You can join the CNCF Slack and access our meetings at the [Kubeflow Community](https://www.kubeflow.org/docs/about/community/) website. Our channel on the CNCF Slack is here [**#kubeflow-platform**](https://app.slack.com/client/T08PSQ7BQ/C073W572LA2).

PR #3317, opened 2026-01-04
fix(knative): update PDBs to allow node drain

aniketpati1121:
## ✏️ Summary of Changes
- Updated PodDisruptionBudgets (PDBs) for Knative core components (activator, webhook, eventing-webhook)
- Adjusted `minAvailable` and replicas to allow safe node drain for single-replica deployments
- Ensures cluster maintenance can proceed without manually evicting pods or blocking drains
## 🐛 Related Issues
Fixes #3313

PR #3321, opened 2026-01-11
documentation: propose kubeflow-system namespace migration

aniketpati1121:
This PR adds a design proposal for gradually migrating Kubeflow control-plane 
components from the hard-coded `kubeflow` namespace to `kubeflow-system`. 

It outlines:
- Current namespace usage
- Risks and motivation for change
- Proposed component-by-component migration
- Testing and rollout strategy

### DIFF
--- a/proposals/kubeflow-system-namespace.md
+++ b/proposals/kubeflow-system-namespace.md
@@ -1,0 +1,207 @@
+# Proposal: kubeflow-system Namespace Migration
+
+**Author:** Aniket Patil (@aniketpatil1121)  
+**Status:** Draft – Seeking Maintainer Feedback
+
+---
+
+## Overview
+
+This document proposes a gradual and safe migration of Kubeflow control-plane
+components from the hard-coded `kubeflow` namespace to `kubeflow-system`,
+or alternatively making the namespace configurable where required.
+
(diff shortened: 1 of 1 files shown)


juliusvonkohout:
I think there are more pressing issues such as https://github.com/kubeflow/manifests/pull/3319, https://github.com/kubeflow/manifests/pull/3317 and many more that i will prioritize.

juliusvonkohout:
We should actually consider decoupling, so kubeflow-katib, kubeflow-pipelines, kubeflow-dashboard etc. as namespace names.

andreyvelich:
> We should actually consider decoupling, so kubeflow-katib, kubeflow-pipelines, kubeflow-dashboard etc. as namespace names.

@juliusvonkohout Any particular reason you want to separate system namespaces by the project? I think, it would be easier and better to install all control plane components into single `kubeflow-system` namespace.

juliusvonkohout:
> > We should actually consider decoupling, so kubeflow-katib, kubeflow-pipelines, kubeflow-dashboard etc. as namespace names.
> 
> @juliusvonkohout Any particular reason you want to separate system namespaces by the project? I think, it would be easier and better to install all control plane components into single `kubeflow-system` namespace.

CC @thesuperzapper I am mostly fine either way.

PR #3348, opened 2026-02-16
Kserve path based routing

grudloffev:
# Pull Request Template for Kubeflow Manifests

KServe supports native path-based routing via the `pathTemplate` configuration, eliminating manual VirtualService management.

**KServe Configuration** (`kserve.yaml`, `kserve_kubeflow.yaml`):
- Added `pathTemplate` to inferenceservice-config ingress settings:
  ```json
  "pathTemplate": "/serving/{{ .Namespace }}/{{ .Name }}"
  ```

**Test Updates** (`tests/kserve_test.sh`):
- Removed manual VirtualService creation (27 lines)
- Updated paths: `/kserve/${NAMESPACE}/isvc-sklearn/` → `/serving/${NAMESPACE}/isvc-sklearn/`
- Retained AuthorizationPolicy for cluster-local-gateway security

KServe now auto-generates VirtualServices with the configured path template when InferenceServices are created.

Discussion: https://github.com/kubeflow/manifests/issues/3263#issuecomment-3898938747

N/A

## ✅ Contributor Checklist
  - [x] I have tested these changes with kustomize. See [Installation Prerequisites](https://github.com/kubeflow/manifests#prerequisites).
  - [x] All commits are [_signed-off_](https://github.com/kubeflow/community/tree/master/dco-signoff-hook#signing-off-commits) to satisfy the DCO check.
  - [ ] I have considered adding my company to the adopters page to support Kubeflow and help the community, since I expect help from the community for my issue (see [1.](https://github.com/kubeflow/community/issues/833) and [2.](https://github.com/kubeflow/community/blob/master/ADOPTERS.md#adopters-of-kubeflow-platform)).

---     
 
> You can join the CNCF Slack and access our meetings at the [Kubeflow Community](https://www.kubeflow.org/docs/about/community/) website. Our channel on the CNCF Slack is here [**#kubeflow-platform**](https://app.slack.com/client/T08PSQ7BQ/C073W572LA2).

PR #3352, opened 2026-02-18
refactor(tests): consolidate KServe test infrastructure into two-file pattern

Raakshass:
## ✏️ Summary of Changes
Consolidates scattered KServe test scripts into the canonical `kserve_install.sh` + `kserve_test.sh` two-file pattern, matching every other component's test structure in this repository.
### Consolidated `tests/kserve_test.sh`
Merged three separate test scripts into a single structured script with **6 tests**:
| # | Test | Source |
|---|------|--------|
| 1 | Model inference via KServe SDK (pytest) | existing `kserve_test.sh` |
| 2 | Ingress gateway security — AuthorizationPolicy enforcement | existing `kserve_test.sh` (removed manual VirtualService) |
| 3 | KServe Models Web Application API — XSRF + auth + unauthorized access | `kserve_models_web_application_test.sh` |
| 4 | Knative Service authentication via cluster-local-gateway | existing `kserve_test.sh` |
| 5 | Namespace isolation — cross-namespace token rejection | `kserve_complete_authentication_test.sh` |
| 6 | Unauthenticated request rejection | existing `kserve_test.sh` |
**Key changes to `kserve_test.sh`:**
- Removed manual VirtualService creation (now handled natively by KServe via `pathTemplate` — see #3348)
- Updated prediction paths from `/kserve/` to `/serving/` to align with #3348
- Added explicit `echo "FAIL: ..."` error messages for all failure cases
### Deleted obsolete files
| File | Reason |
|------|--------|
| `tests/kserve_complete_authentication_test.sh` | Logic merged into `kserve_test.sh` |
| `tests/kserve_models_web_application_test.sh` | Logic merged into `kserve_test.sh` |
| `tests/kserve_setup_external_access.sh` | Obsoleted by #3348 native `pathTemplate` support |
| `tests/kserve/kserve-external-access.yaml` | Obsoleted by #3348 native `pathTemplate` support |
| `tests/kserve_test.yaml` | Unused CI artifact (not to be confused with the workflow) |
### Updated CI workflows
- `.github/workflows/kserve_models_web_application_test.yaml`:
  - Calls consolidated `kserve_test.sh` instead of deleted `kserve_models_web_application_test.sh`
  - Added `setup-python@v4` step (required by pytest in consolidated script)
  - Added `tests/kserve/**` path trigger
  - Increased `timeout-minutes` from 25 to 45 (now runs all 6 tests)
- `.github/workflows/full_kubeflow_integration_test.yaml`:
  - Removed redundant `Test KServe Models Web Application API` step (already covered by `kserve_test.sh` on the preceding step)
### Updated dependencies
- `tests/kserve/requirements.txt`: `kserve>=0.15.0` → `kserve>=0.16.0`
## 🐛 Related Issues
- Addresses test infrastructure fragmentation across multiple KServe test scripts
- Aligns with the two-file pattern (`*_install.sh` + `*_test.sh`) used by all other components
## ✅ Contributor Checklist
  - [ ] I have considered adding my company to the adopters page to support Kubeflow and help the community, since I expect help from the community for my issue (see [1.](https://github.com/kubeflow/community/issues/833) and [2.](https://github.com/kubeflow/community/blob/master/ADOPTERS.md#adopters-of-kubeflow-platform)).
/cc @juliusvonkohout

### DIFF
--- a/common/oauth2-proxy/README.md
+++ b/common/oauth2-proxy/README.md
@@ -275,6 +275,141 @@ for listing notebooks:
 
 ## KServe Authentication
 
+KServe inference endpoints are secured through a layered approach using
+Istio `RequestAuthentication` and `AuthorizationPolicy` resources.
+The examples below focus on machine-to-machine (M2M) access using service
+account tokens. Browser-based user access follows the general Kubeflow
+`oauth2-proxy` flow described above.
+
+### Traffic Flow and Security Layers
+
+Inference requests to KServe models pass through two security checkpoints:
+
+```
+client ──► istio-ingressgateway (istio-system) ──► predictor pod sidecar ──► predictor container
+              │                                         │
+              ▼                                         ▼
+         RequestAuthentication                   AuthorizationPolicy
+         (JWT validation)                        (access control)
+```
+
+1. **Ingress gateway** (`istio-system`): A `RequestAuthentication` resource
+   validates the JWT in the `Authorization: Bearer <token>` header. Requests
+   with an invalid token are rejected with `401`. Requests without any token
+   pass through (handled by the `AuthorizationPolicy` at the next layer).
+2. **Predictor pod sidecar**: An `AuthorizationPolicy` controls which
+   requests reach the model container.
+
+### Configuring AuthorizationPolicy for Predictor Pods
+
+Kubeflow ships a `global-deny-all` `AuthorizationPolicy` in `istio-system` that
+blocks all mesh traffic by default. To allow inference traffic to reach a
+predictor pod, you must create an `AuthorizationPolicy` in the model's
+namespace.
+
+The **intended** configuration uses `requestPrincipals: ["*"]`, which matches
+any request carrying a validated JWT principal:
+
+```yaml
+apiVersion: security.istio.io/v1beta1
+kind: AuthorizationPolicy
+metadata:
+  name: allow-isvc-sklearn
+  namespace: kubeflow-user-example-com
+spec:
+  action: ALLOW
+  rules:
+  - from:
+    - source:
+        requestPrincipals: ["*"]
+  selector:
+    matchLabels:
+      serving.knative.dev/service: isvc-sklearn-predictor
+```
+
+`requestPrincipals: ["*"]` relies on the JWT principal being propagated from
+the ingress gateway to the predictor pod's Envoy sidecar. This propagation
+depends on the Istio mTLS and `PeerAuthentication` configuration of the
+cluster. In environments where the principal is not propagated (e.g., some
+KinD-based CI setups), `requestPrincipals` will always be empty at the
+sidecar, and the rule will never match.
+
+In such environments, the CI tests use a permissive fallback:
+
+```yaml
+spec:
+  action: ALLOW
+  rules:
+  - {}   # allow-all: security is enforced at the ingress gateway
+  selector:
+    matchLabels:
+      serving.knative.dev/service: isvc-sklearn-predictor
+```
+
+> **Important:** `rules: [{}]` allows **all** traffic to the predictor pod,
+> including unauthenticated requests that bypass the ingress gateway.
+> This is acceptable in CI because the ingress gateway's
+> `RequestAuthentication` is the primary security boundary — it validates
+> the JWT **before** forwarding traffic to the predictor. However, in
+> production clusters with proper mTLS configuration, prefer
+> `requestPrincipals: ["*"]` for defense in depth.
+
+### Path-Based and Host-Based Routing
+
+KServe supports two routing modes for inference requests, both secured by the
+same authentication flow:
+
+| Mode | URL pattern | Configuration |
+|------|-------------|---------------|
+| Path-based | `http://<gateway>/serving/<namespace>/<name>/v1/models/<name>:predict` | `pathTemplate` in `inferenceservice-config` ConfigMap |
+| Host-based | `http://<gateway>/v1/models/<name>:predict` with `Host: <name>.<namespace>.example.com` | `domainTemplate` in `inferenceservice-config` ConfigMap |
+
+Path-based routing is configured via a kustomize patch on the
+`inferenceservice-config` ConfigMap
+(`applications/kserve/kserve/kustomization.yaml`):
+
+```json
+{
+  "pathTemplate": "/serving/{{ .Namespace }}/{{ .Name }}"
+}
+```
+
+KServe auto-generates a `VirtualService` on the `kubeflow-gateway` for each
+`InferenceService`, enabling both routing modes simultaneously.
+
+### KServe Models Web Application Authentication
+
+The KServe Models Web Application (`kserve-models-web-app`) uses the same
+XSRF + Bearer token authentication pattern as other Kubeflow web applications.
+API calls require:
+
+1. A valid XSRF token (obtained via cookie on the initial page load)
+2. A valid `Authorization: Bearer <token>` header
+3. The token's identity must have RBAC permissions in the target namespace
+
+Unauthorized service accounts (e.g., `default` SA from a different namespace)
+receive `401`/`403` when attempting to list `InferenceService` resources in a
+namespace they do not have access to.
+
+### CI Test Coverage
+
+The KServe test suite (`tests/kserve_test.sh`) validates the following
+authentication and security scenarios end-to-end in a KinD cluster:
+
+| # | Test | What is verified |
+|---|------|-----------------|
+| 1 | Model prediction via KServe Python SDK | InferenceService deployment, prediction, and cleanup using the `kserve` SDK with M2M token |
+| 2a | Path-based routing without token | Unauthenticated request returns `403`/`302` |
+| 2b | Host-based routing without token | Unauthenticated request returns `403`/`302` |
+| 2c | Path-based routing with valid token | Authenticated request returns `200` |
+| 2d | Host-based routing with valid token | Authenticated request returns `200` |
+| 3 | KServe Models Web App API | XSRF + auth flow, unauthorized SA gets `401`/`403` |
+| 4 | Knative Service auth via cluster-local-gateway | Unauthenticated and invalid-token requests are rejected |
+| 5 | Cluster-local-gateway authentication | Direct access without token returns `403` |
+| 6 | Namespace isolation | Cross-namespace attacker token is rejected |
+
+### Architecture Analysis (Future Improvements)
+
 The analysis of KServe auth capabilities suggests that while it's possible to limit access to only authenticated agents,
 there might be some improvements required to enable access only to authorized agents.
 
@@ -297,6 +432,11 @@ This is based on the following:
    > create an [Istio AuthorizationPolicy](https://istio.io/latest/docs/reference/config/security/authorization-policy/) to grant access to the pods or disable it
 
    Most probably some work is needed to enable authorized access to kserve models.
+3. Potential improvement: adding `source.namespaces` to the `AuthorizationPolicy`
+   to restrict access to traffic originating from specific namespaces (e.g.,
+   `istio-system`). This would provide an additional layer of security but
+   requires proper mTLS/PeerAuthentication configuration to propagate SPIFFE
+   identities correctly.
 
 ## Links
 


tests/kserve/requirements.txt
juliusvonkohout, line 1: may you inline this as well to remove the kserve folder altogether ?

.github/workflows/kserve_models_web_application_test.yaml
```diff
@@ -21,7 +22,7 @@ env:
 jobs:
   build:
     runs-on: ubuntu-latest
-    timeout-minutes: 25
+    timeout-minutes: 45
```
juliusvonkohout: why 45 minutes

tests/kserve_test.sh
```diff
@@ -13,39 +13,20 @@ export KSERVE_INGRESS_HOST_PORT=${KSERVE_INGRESS_HOST_PORT:-localhost:8080}
...
-            request:
-              set:
-                Host: isvc-sklearn-predictor.${NAMESPACE}.svc.cluster.local
-          weight: 100
-      timeout: 300s
-EOF
+# ============================================================
+# Test 2: Ingress Gateway Security (AuthorizationPolicy)
+# ============================================================
+# NOTE: Path-based routing is now handled natively by KServe via pathTemplate
```
juliusvonkohout: where is pathTemplate ?

tests/kserve/requirements.txt
Raakshass: Done — inlined the pip dependencies directly in [kserve_test.sh](cci:7://file:///c:/Users/asus%20vivoBook/Desktop/New%20folder%20%282%29/kubeflow-manifests/tests/kserve_test.sh:0:0-0:0) (`pip install "pytest>=7.0.0" "kserve>=0.16.0" "kubernetes>=18.20.0" "requests>=2.18.4"`) and deleted [requirements.txt](cci:7://file:///c:/Users/asus%20vivoBook/Desktop/New%20folder%20%282%29/kubeflow-manifests/tests/kserve/requirements.txt:0:0-0:0). The `tests/kserve/` folder still contains [test_sklearn.py](cci:7://file:///c:/Users/asus%20vivoBook/Desktop/New%20folder%20%282%29/kubeflow-manifests/tests/kserve/test_sklearn.py:0:0-0:0), [utils.py](cci:7://file:///c:/Users/asus%20vivoBook/Desktop/New%20folder%20%282%29/kubeflow-manifests/tests/kserve/utils.py:0:0-0:0), and [data/iris_input.json](cci:7://file:///c:/Users/asus%20vivoBook/Desktop/New%20folder%20%282%29/kubeflow-manifests/tests/kserve/data/iris_input.json:0:0-0:0) which are needed by the pytest SDK test (these predate this PR), so the folder itself stays.

.github/workflows/kserve_models_web_application_test.yaml
Raakshass: The original workflow ran only the kserve_models_web_application test (1 test) in 25 min. The consolidated script now runs all 6 tests sequentially — each with `kubectl wait` + `sleep` timeouts and model deployment waits. Looking at the successful run though, it completed in ~10 min, so happy to reduce this to 30 min if you prefer.

tests/kserve_test.sh
Raakshass: `pathTemplate` is configured in `kserve_kubeflow.yaml` as part of PR #3348 (listed under "Depends-On" in the PR description). That PR adds native KServe path-based routing — `pathTemplate: /serving/{{ .Namespace }}/{{ .Name }}` — which replaces the manual VirtualService creation that this PR removes. The `/serving/` prefix paths in the test script match that template.

.github/workflows/kserve_models_web_application_test.yaml
juliusvonkohout: Let's keep it at 15-20 then

tests/kserve_test.sh
juliusvonkohout: Please add this as inline documentation to kserve_test

tests/kserve/requirements.txt
juliusvonkohout: Can you merge test_sklearn.py, utils.py, and data/iris_input.json  as well into a single python file and move it next to kserve_test to get rid of the folder ?

Raakshass: Done — merged test_sklearn.py, utils.py, and data/iris_input.json into a single tests/kserve_sklearn_test.py. The iris input data is inlined as a Python dict, all utility functions are in the same file. The tests/kserve/ folder is completely eliminated.

.github/workflows/kserve_models_web_application_test.yaml
Raakshass: Reduced to 20 minutes. Actual CI runs completed in ~10 min, so this leaves sufficient headroom.

tests/kserve_test.sh
Raakshass: Added expanded inline documentation at Test 2 in [kserve_test.sh](cci:7://file:///c:/Users/asus%20vivoBook/Desktop/New%20folder%20%282%29/manifests-fix/tests/kserve_test.sh:0:0-0:0). The comment block now documents:

- The exact pathTemplate pattern: `/serving/{{ .Namespace }}/{{ .Name }}`
- Where it is configured: `kserve_kubeflow.yaml` (introduced in PR #3348)
- What it means for routing: every InferenceService is automatically reachable at `http://<gateway>/serving/<namespace>/<isvc-name>/...`
- Why no manual VirtualService creation is needed
- Why all curl URLs in the test use the `/serving/` prefix

tests/kserve_sklearn_test.py
```diff
@@ -123,3 +130,35 @@ def predict_str(
...
+    )
+
+    kserve_client = KServeClient(
+        config_file=os.environ.get("KUBECONFIG", "~/.kube/config")
+    )
+    kserve_client.create(isvc)
+    kserve_client.wait_isvc_ready(service_name, namespace=KSERVE_TEST_NAMESPACE)
+    res = predict(service_name, IRIS_INPUT)
+    assert res["predictions"] == [1, 1]
+    kserve_client.delete(service_name, KSERVE_TEST_NAMESPACE)
```
juliusvonkohout:
```suggestion
    response = predict(service_name, IRIS_INPUT)
    assert response["predictions"] == [1, 1]
    kserve_client.delete(service_name, KSERVE_TEST_NAMESPACE)
```

juliusvonkohout:
image
The test must fail if the a single part fails, also the python part.

tests/kserve_test.sh
```diff
@@ -3,49 +3,36 @@ set -euxo pipefail
...
-                Host: isvc-sklearn-predictor.${NAMESPACE}.svc.cluster.local
-          weight: 100
-      timeout: 300s
-EOF
+# ============================================================
+# Test 1: Model Inference via KServe SDK
+# ============================================================
+# Runs the consolidated kserve_sklearn_test.py which deploys an
+# sklearn InferenceService and validates prediction output.
+pytest "${SCRIPT_DIRECTORY}/kserve_sklearn_test.py" -vs --log-level info || true
```
juliusvonkohout: This test failed, with 
```
/opt/hostedtoolcache/Python/3.12.12/x64/lib/python3.12/site-packages/requests/models.py:1026: HTTPError
------------------------------ Captured log call -------------------------------
INFO     root:kserve_sklearn_test.py:110 M2M Token Found.
INFO     root:kserve_sklearn_test.py:121 Sending Header = {'Host': 'isvc-sklearn.kubeflow-user-example-com.example.com', 'Content-Type': 'application/json', 'Authorization': '***'}
INFO     root:kserve_sklearn_test.py:122 Sending url = http://localhost:8080/v1/models/isvc-sklearn:predict
INFO     root:kserve_sklearn_test.py:123 Sending request data: {"instances": [[6.8, 2.8, 4.8, 1.4], [6.0, 3.4, 4.5, 1.6]]}
INFO     root:kserve_sklearn_test.py:125 Got response code 403, content b'RBAC: access denied'
=========================== short test summary info ============================
FAILED tests/kserve_sklearn_test.py::test_sklearn_kserve - requests.exceptions.HTTPError: 403 Client Error: Forbidden for url: http://localhost:8080/v1/models/isvc-sklearn:predict
======================== 1 failed in 101.69s (0:01:41) =========================
+ true
```

but it did not fail the whole workflow, which is dangerous

tests/kserve_test.sh
```diff
@@ -3,49 +3,36 @@ set -euxo pipefail
...
+
+# ============================================================
+# Test 2: Ingress Gateway Security (AuthorizationPolicy)
+# ============================================================
+# Path-based routing is handled natively by KServe via the pathTemplate
+# configuration in kserve_kubeflow.yaml (introduced in PR #3348):
```
juliusvonkohout: This PR is not merged. Please use a permanent link to our master branch.

Do you not have to write a patch for kserve to set the /applications/kserve/kserve/kserve_kubeflow.yaml#L32682 ? Or do you rely on an implicit default pathtemplate ?

juliusvonkohout: For example here https://github.com/kubeflow/manifests/blob/master/applications/kserve/kserve/kserve_kubeflow.yaml#L72531 I do not see the pathtemplate. Note never ever modify the upstream folder, use patches instead.

tests/kserve_test.sh
juliusvonkohout, line 1: an additional test for a raw deployment would also be interesting, see https://github.com/kserve/kserve/pull/5090
see also https://github.com/kserve/kserve/blob/b8fa0fbf8310c07b775e5c7407043253a2a0e6d0/config/configmap/inferenceservice.yaml#L389

tests/kserve_test.sh
juliusvonkohout: see also https://github.com/kserve/kserve/blob/b8fa0fbf8310c07b775e5c7407043253a2a0e6d0/config/configmap/inferenceservice.yaml#L389 patching the configmap could be useful as well

tests/kserve_test.sh
Raakshass: Agreed, this was dangerous. Two fixes applied:

1. Removed `|| true` — pytest failures are now fatal.
2. The 403 root cause was Host-header routing in [predict_str](cci:1://file:///c:/Users/asus%20vivoBook/Desktop/New%20folder%20%282%29/kubeflow-manifests/tests/kserve_sklearn_test.py:88:0-131:35) — it set `Host: isvc-sklearn.kubeflow-user-example-com.example.com` and hit `/v1/models/...`, which Istio's default-deny AuthorizationPolicy blocks. Switched to path-based routing via `/serving/<namespace>/<name>/v1/models/<model>:predict`, which routes through the KServe-generated VirtualService.

tests/kserve_test.sh
Raakshass: You're right — pathTemplate was only in the `_example` docs block, not in the active `ingress` config. The `/serving/` URLs were silently failing.

Added a kustomize strategic merge patch in [applications/kserve/kserve/kustomization.yaml](cci:7://file:///c:/Users/asus%20vivoBook/Desktop/New%20folder%20%282%29/kubeflow-manifests/applications/kserve/kserve/kustomization.yaml:0:0-0:0) that replaces the `ingress` data block in the `inferenceservice-config` ConfigMap with the full config including:
```json
"pathTemplate": "/serving/{{ .Namespace }}/{{ .Name }}"

File: tests/kserve_sklearn_test.py [hunk 6]
```diff
@@ -0,0 +1,247 @@
...
+        raise M2mTokenNotAvailable(env_name)
+
+
+def create_allow_authorization_policy(service_name, namespace):
+    """Create an Istio AuthorizationPolicy that allows authenticated requests
+    to reach the InferenceService predictor pods.
+
+    Required because Kubeflow installs a mesh-wide global-deny-all policy
+    (common/istio/istio-install/base/deny_all_authorizationpolicy.yaml).
+    Without an explicit ALLOW policy, the sidecar blocks all traffic.
```
juliusvonkohout: Are you sure. Could it be that it only blocks traffic from other namespaces? Did you try it from the same namespace ?

tests/kserve_test.sh
```diff
@@ -3,94 +3,134 @@ set -euxo pipefail
...
+# The AuthorizationPolicy (allow-isvc-sklearn) was created by pytest above.
+# It allows requests with any valid requestPrincipal to reach the predictor.
+#
+# This test validates:
+#   a) Requests WITHOUT a token → should be blocked (403/302)
+#   b) Requests WITH a valid token → should succeed (200)
+#
+# We use Host-header routing to match the Knative-managed VirtualService.
+
+sleep 10
```
juliusvonkohout: why did you add sleep 10 ?

tests/kserve_test.sh
```diff
@@ -3,94 +3,134 @@ set -euxo pipefail
...
+
+HOST_HEADER="Host: isvc-sklearn.${NAMESPACE}.example.com"
 
 # Request without token should be rejected
 RESPONSE_NO_TOKEN=$(curl -s -o /dev/null -w "%{http_code}" \
+ -H "${HOST_HEADER}" \
  -H "Content-Type: application/json" \
- "http://${KSERVE_INGRESS_HOST_PORT}/kserve/${NAMESPACE}/isvc-sklearn/v1/models/isvc-sklearn:predict" \
+ "http://${KSERVE_INGRESS_HOST_PORT}/v1/models/isvc-sklearn:predict" \
```
juliusvonkohout: Ok you added host based routing. Yes this needs to be tested as well, but where is now the test for path based routing ?

juliusvonkohout: We need both. host based and path based for python and host based and path based for bash/curl

tests/kserve_sklearn_test.py
```diff
@@ -0,0 +1,247 @@
...
+
+    # Create an AuthorizationPolicy that allows authenticated requests to
+    # reach this InferenceService. Without this, Kubeflow's mesh-wide
+    # global-deny-all policy blocks all pod-level traffic.
+    create_allow_authorization_policy(service_name, KSERVE_TEST_NAMESPACE)
+
+    try:
+        response = predict(service_name, IRIS_INPUT)
+        assert response["predictions"] == [1, 1]
+    finally:
```
juliusvonkohout, line 262: Actually you should delete it here and recreate it with plain bash+yaml in follow up tests.

tests/kserve_sklearn_test.py
Raakshass: honestly no, i didn't test from the same namespace — i assumed the global-deny-all with
empty spec `{}` in istio-system would deny everything mesh-wide (which is what the istio
docs say for an ALLOW policy with no rules). but digging deeper, i realized the actual
problem isn't even the global-deny-all.
the real issue is that the m2m `RequestAuthentication` (`m2m-token-issuer`) only validates
tokens at the ingress gateway (`selector: app: istio-ingressgateway`), not at the sidecar
level. so even though i was creating an AP with `requestPrincipals: ["*"]` on the predictor
pods, the sidecar couldn't extract any requestPrincipal from the token because there's no
RequestAuthentication configured on the sidecars — meaning the condition would never match
regardless of whether global-deny-all was the culprit or not.
i've removed the AP creation from pytest entirely now. prediction testing is done via curl
in bash using path-based routing through the cluster-local-gateway, which is the same
proven approach from master.

tests/kserve_test.sh
Raakshass: my thinking was that istio needs some time to propagate the AuthorizationPolicy to all
envoy sidecars after creation. the master branch uses `sleep 60` for this, so i thought
10 seconds might be enough as a faster alternative.
in hindsight this was wrong for two reasons: (1) the AP wasn't even the right fix (see
my other reply about requestPrincipals), and (2) even if it was, 10 seconds is too
aggressive — istio docs recommend allowing at least 30-60s for policy propagation in
production-like setups.
removed it. the only sleep now is `sleep 60` in the bash curl test section, right after
creating the AP + VirtualService — same as master.

tests/kserve_test.sh
Raakshass: yeah, i initially switched to host-based routing because i was trying to avoid the
manual VirtualService creation that master does — i thought the Knative-managed
VirtualService would handle it natively. but that approach broke because:
1. the Knative VirtualService is attached to `knative-serving/knative-ingress-gateway`,
   not `kubeflow/kubeflow-gateway`  
2. without the pathTemplate actually being configured (it was only in the _example docs
   block, not the active config), path-based routing wasn't available either
so i ended up with only host-based and no path-based, which is exactly the gap you caught.
fixed now — Test 2a does path-based routing (with a manually-created VirtualService
through `kubeflow-gateway` → `cluster-local-gateway`, same as master), and Test 2b does
host-based routing with `Host: isvc-sklearn.<namespace>.example.com`. both test
with-token and without-token scenarios.

tests/kserve_sklearn_test.py
Raakshass: makes sense — keeping the tests independent is much cleaner. i was originally trying to
do everything in one pytest run (deploy + predict + cleanup) to minimize test duration,
but that created a tight coupling between the python SDK test and the routing/auth
infrastructure.
done now:
- pytest deploys the isvc, waits for Ready, then deletes it in cleanup
- `kserve_test.sh` recreates it with `kubectl apply` (plain bash+yaml)
- prediction is tested entirely via curl in bash — no python predict function needed
this way if the python SDK deploy works but curl prediction fails (or vice versa), we
know exactly which layer broke.

tests/kserve_test.sh
juliusvonkohout: yes then its fine

tests/kserve_test.sh
juliusvonkohout: Are you sure that you pushed the changes to this branch ?

tests/kserve_sklearn_test.py
juliusvonkohout: No, do not remove the prediction test from python. It must be there as well. Both python and bash must work indepedently

tests/kserve/test_sklearn.py
```diff
@@ -55,6 +67,11 @@ def test_sklearn_kserve():
     )
     kserve_client.create(isvc)
     kserve_client.wait_isvc_ready(service_name, namespace=KSERVE_TEST_NAMESPACE)
-    res = predict(service_name, "./data/iris_input.json")
```
juliusvonkohout, line 58: No, do not remove the prediction test from python. It must be there as well. Both python and bash must work indepedently. keep it with proper variable naming.

tests/kserve_test.sh
juliusvonkohout: But the goal is to remove the manually created virtualservice as in https://github.com/kubeflow/manifests/pull/3348

tests/kserve_sklearn_test.py
juliusvonkohout: What is an AP? Do you mean authorization policy ?

tests/kserve/utils.py
juliusvonkohout, line 1: as requested before please merge this file into kserve_sklearn_test.py and remove the kserve folder.

tests/requirements.txt
juliusvonkohout, line 1: as requested before please merge this file into kserve_sklearn_test.py and remove the kserve folder.

.github/workflows/kserve_models_web_application_test.yaml
juliusvonkohout:
```suggestion
    timeout-minutes: 25
```

tests/kserve_sklearn_test.py
```diff
@@ -0,0 +1,167 @@
...
+    V1beta1InferenceServiceSpec,
+    V1beta1PredictorSpec,
+    V1beta1SKLearnSpec,
+)
+
+# Add tests/kserve to sys.path so we can import utils
+sys.path.insert(
+    0, os.path.join(os.path.dirname(os.path.abspath(__file__)), "kserve")
+)
+from utils import KSERVE_TEST_NAMESPACE, predict
```
juliusvonkohout: no please inline utils.py

tests/kserve_sklearn_test.py
```diff
@@ -0,0 +1,167 @@
...
+
+logging.basicConfig(level=logging.INFO)
+
+AP_NAME = "allow-isvc-sklearn"
+SERVICE_NAME = "isvc-sklearn"
+
+
+def create_predictor_authz_policy(namespace):
```
juliusvonkohout:
```suggestion
def create_predictor_authorization_policy(namespace):
```

please use proper variable names.

tests/kserve_sklearn_test.py
```diff
@@ -0,0 +1,167 @@
...
+
+# Add tests/kserve to sys.path so we can import utils
+sys.path.insert(
+    0, os.path.join(os.path.dirname(os.path.abspath(__file__)), "kserve")
+)
+from utils import KSERVE_TEST_NAMESPACE, predict
+
+logging.basicConfig(level=logging.INFO)
+
+AP_NAME = "allow-isvc-sklearn"
```
juliusvonkohout:
```suggestion
AUTHORIZATION_POLICY_NAME = "allow-isvc-sklearn"
```

Use proper variable names. Long expressive pronounceable

tests/kserve_sklearn_test.py
```diff
@@ -0,0 +1,167 @@
...
+        group="security.istio.io",
+        version="v1beta1",
+        namespace=namespace,
+        plural="authorizationpolicies",
+        body=ap_body,
+    )
+    logging.info("Created AuthorizationPolicy %s in %s", AP_NAME, namespace)
+
+
+def delete_predictor_authz_policy(namespace):
```
juliusvonkohout:
```suggestion
def delete_predictor_authorization_policy(namespace):
```
use proper names. Long expressive and pronounceable

tests/kserve_sklearn_test.py
```diff
@@ -0,0 +1,167 @@
...
+        ),
+        spec=V1beta1InferenceServiceSpec(predictor=predictor),
+    )
+
+    kserve_client = KServeClient(
+        config_file=os.environ.get("KUBECONFIG", "~/.kube/config")
+    )
+
+    try:
+        # Create AuthorizationPolicy BEFORE the ISVC so it's ready
```
juliusvonkohout:
```suggestion
        # Create the AuthorizationPolicy BEFORE the Inference service
```

tests/kserve_test.sh
```diff
@@ -13,41 +13,45 @@ export KSERVE_INGRESS_HOST_PORT=${KSERVE_INGRESS_HOST_PORT:-localhost:8080}
...
+          memory: "128Mi"
+        limits:
+          cpu: "100m"
+          memory: "256Mi"
 EOF
 
-# WARNING: This policy allows ANY valid token from ANY kubeflow namespace to access this InferenceService.
+kubectl wait --for=condition=Ready inferenceservice/isvc-sklearn -n ${NAMESPACE} --timeout=300s
+
+# Allow traffic to the predictor pod. The ingress gateway and CLG
```
juliusvonkohout:
```suggestion
# Allow traffic to the predictor pod. The ingress gateway and Cluster local gateway
```

please be expressive.

tests/kserve_test.sh
```diff
@@ -13,41 +13,45 @@ export KSERVE_INGRESS_HOST_PORT=${KSERVE_INGRESS_HOST_PORT:-localhost:8080}
...
+# have already validated the JWT before forwarding.
+# NOTE: requestPrincipals:["*"] does NOT work here because the predictor
+# sidecar has no RequestAuthentication to extract the requestPrincipal.
+# NOTE: source.namespaces requires mTLS identity which may not be
+# established in all environments (e.g. KinD CI).
```
juliusvonkohout: Why do you assume this ? We rely on mtls. Restricting by namespace source can make a lot of sense.

tests/kserve_test.sh
```diff
@@ -57,40 +61,135 @@ metadata:
...
+kubectl get inferenceservice sklearn-iris -n ${NAMESPACE} || exit 1
+READY=$(kubectl get isvc sklearn-iris -n ${NAMESPACE} -o jsonpath='{.status.conditions[?(@.type=="Ready")].status}')
+[[ "$READY" == "True" ]] || {
+  echo "FAILURE: InferenceService sklearn-iris Ready status is: $READY"
+  exit 1
+}
+
+kubectl delete inferenceservice sklearn-iris -n ${NAMESPACE} || exit 1
+
+# Test unauthorized access to models web app
```
juliusvonkohout:
```suggestion
# Test unauthorized access to models web application
```

tests/kserve_test.sh
```diff
@@ -57,40 +61,135 @@ metadata:
...
+UNAUTH_TOKEN="$(kubectl -n default create token default)"
+HTTP_CODE=$(curl -s -o /dev/null -w "%{http_code}" "${BASE_URL}/api/namespaces/${NAMESPACE}/inferenceservices" -H "Authorization: Bearer ${UNAUTH_TOKEN}")
+[[ "$HTTP_CODE" == "403" || "$HTTP_CODE" == "401" ]] || { echo "FAILURE: Expected 401/403, got $HTTP_CODE"; exit 1; }
+echo "Models Web App: Token from unauthorized ServiceAccount cannot list InferenceServices in $NAMESPACE namespace."
```
juliusvonkohout:
```suggestion
echo "Models Web Application: Token from unauthorized ServiceAccount cannot list InferenceServices in $NAMESPACE namespace."
```

kubeflow-all.yaml
juliusvonkohout, line 1: where is this file coming from ?

test-fail.yaml
juliusvonkohout, line 1: where is this file coming from ?

test-pass.yaml
juliusvonkohout, line 1: where is this file coming from ?

Raakshass:
@juliusvonkohout I followed your guidance to use `source.namespaces: ["istio-system"]`
in the AuthorizationPolicy, but the CI KinD tests are consistently failing with
`403 RBAC: access denied`.

**CI logs** (both `test-kserve` and `build` jobs, commit `1ce0284`):

    INFO  Created AuthorizationPolicy allow-isvc-sklearn in kubeflow-user-example-com
    INFO  M2M Token Found.
    INFO  Sending url = http://localhost:8080/v1/models/isvc-sklearn:predict
    INFO  Got response code 403, content b'RBAC: access denied'

**What I applied** (as you directed):

    spec:
      action: ALLOW
      rules:
      - from:
        - source:
            namespaces: ["istio-system"]
      selector:
        matchLabels:
          serving.knative.dev/service: isvc-sklearn-predictor

**Observation**: The current master branch uses `requestPrincipals: ["*"]` in the same
AuthorizationPolicy (kserve_test.sh L66-L67), and those tests pass.

**Traffic flow**:
client → istio-ingressgateway (istio-system) → predictor pod sidecar → predictor container

Since the ingress gateway lives in `istio-system`, I expected `source.namespaces`
to match via mTLS SPIFFE identity. But the sidecar rejects it, suggesting the
identity isn't established in KinD CI.

**Question**: Should I:
1. Keep `source.namespaces` and investigate the mTLS/PeerAuthentication config in KinD CI?
2. Revert to `requestPrincipals: ["*"]` to match master?
3. Something else?

juliusvonkohout:
Revert to request principals: [*] for the time being. Source namespace is useful in real-world scenarios where you have an application in the same namespace. But I would rather keep it simple and merge the PR sooner than later. We do not have to be perfect, but just significantly improve the status quo without regressions.

.github/workflows/kserve_test.yaml
```diff
@@ -6,7 +6,7 @@ on:
     - .github/workflows/kserve_test.yaml
     - applications/kserve/**
     - apps/kserve/**
-    - tests/kserve/**
+    - tests/kserve_*
```
juliusvonkohout:
```suggestion
    - tests/kserve*
```

tests/kserve_test.sh
```diff
@@ -3,51 +3,51 @@ set -euxo pipefail
 
 NAMESPACE=${1:-kubeflow-user-example-com}
 SCRIPT_DIRECTORY="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-TEST_DIRECTORY="${SCRIPT_DIRECTORY}/kserve"
 
 if ! command -v pytest &> /dev/null; then
-  pip install -r ${TEST_DIRECTORY}/requirements.txt
+  pip install -r ${SCRIPT_DIRECTORY}/requirements.txt
 fi
```
juliusvonkohout: You can drop that as well as the requirements.txt file and inline it into kserve_sklearn_test.py

tests/requirements.txt
juliusvonkohout: So the file must be deleted.

tests/kserve_test.sh
juliusvonkohout: For example `os.system("pip install pytest>=7.0.0 kserve>=0.16.0 kubernetes>=18.20.0 requests>=2.18.4")`

tests/kserve_sklearn_test.py
```diff
@@ -0,0 +1,270 @@
...
+to the predictor pod. Security is maintained because the
+ingress gateway and cluster-local-gateway both validate the
+JWT via RequestAuthentication before forwarding.
+"""
+
+import os
+import sys
+
+# Install dependencies inline (replaces the deleted requirements.txt).
+# This ensures pytest, kserve SDK, and other deps are available when
```
juliusvonkohout:
```suggestion
# This ensures pytest, kserve SDK, and other dependencies are available when
```

pleas use proper spelling. deps is not correct.

tests/kserve_test.sh
```diff
@@ -3,51 +3,48 @@ set -euxo pipefail
...
-if cd ${TEST_DIRECTORY}; then
-  pytest . -vs --log-level info || true
-fi
+# ============================================================
+# Test 1: Model Prediction via KServe Python SDK
+# ============================================================
+# Runs kserve_sklearn_test.py which independently deploys an sklearn
+# InferenceService, predicts via host-based routing, asserts the
+# output, and deletes the InferenceService.
+pip install -q "pytest>=7.0.0" "kserve>=0.16.0" "kubernetes>=18.20.0" "requests>=2.18.4"
```
juliusvonkohout: why is this still here if it s inline din the python script ?
```suggestion
```

juliusvonkohout:
I think the raw deployment mode is for a follow up PR. This one here is long-lived enough.

"an additional test for a raw deployment would also be interesting, see https://github.com/kserve/kserve/pull/5090
see also https://github.com/kserve/kserve/blob/b8fa0fbf8310c07b775e5c7407043253a2a0e6d0/config/configmap/inferenceservice.yaml#L389"

juliusvonkohout:
And you need to update the documentation here https://github.com/kubeflow/manifests/blob/master/common/oauth2-proxy/README.md#kserve-authentication

common/oauth2-proxy/README.md
```diff
@@ -275,6 +275,141 @@ for listing notebooks:
...
+              │                                         │
+              ▼                                         ▼
+         RequestAuthentication                   AuthorizationPolicy
+         (JWT validation)                        (access control)
+```
+
+1. **Ingress gateway** (`istio-system`): A `RequestAuthentication` resource
+   validates the JWT in the `Authorization: Bearer <token>` header. Requests
+   with an invalid token are rejected with `401`. Requests without any token
+   pass through (handled by the `AuthorizationPolicy` at the next layer).
```
juliusvonkohout, line 299: Maybe for kserve only

juliusvonkohout:
Thank you, lets cut down the documentation (removing code and documentation is often more important than adding.) and add tests for the raw deployment in a follow up PR. https://github.com/kubeflow/manifests/pull/3352#issuecomment-3945980534

And please print actual prediction result on the console, do not hide them, such that from to test output its easier to see results.

/lgtm
/approve

Raakshass:
Thank you for the thorough review and merge, @juliusvonkohout. Noted the follow-up items:

Trim the KServe auth documentation — remove redundant code blocks
Add raw deployment mode test (ref: kserve/kserve#5090)
Print prediction results to console instead of suppressing them
I'll open the follow-up PR against the updated master.

PR #3357, opened 2026-02-25
fix: replace deprecated gcr.io minio image with docker.io in v1.9.1

ferduz:
## Problem
The image `gcr.io/ml-pipeline/minio:RELEASE.2019-08-14T20-37-41Z-license-compliance` is no longer available on gcr.io, causing `minio` pod to fail with `ImagePullBackOff` on fresh installs of v1.9.1.

## Fix
Replace with the equivalent official image from Docker Hub:
`minio/minio:RELEASE.2019-08-14T20-37-41Z`

## Testing
Verified working on a local kind cluster running Kubeflow v1.9.1.

### DIFF
--- a/apps/pipeline/upstream/third-party/minio/base/minio-deployment.yaml
+++ b/apps/pipeline/upstream/third-party/minio/base/minio-deployment.yaml
@@ -30,7 +30,7 @@ spec:
             secretKeyRef:
               name: mlpipeline-minio-artifact
               key: secretkey
-        image: minio/minio:RELEASE.2019-08-14T20-37-41Z
+        image: docker.io/minio/minio:RELEASE.2019-08-14T20-37-41Z
         name: minio
         ports:
         - containerPort: 9000


juliusvonkohout:
Is there a reason you are not using Kubeflow 1.11 ? 1.9 is really old.

juliusvonkohout:
And you are not specifying dockerhub for the oci image. It must start with docker.io to do so.

ferduz:
> Is there a reason you are not using Kubeflow 1.11 ? 1.9 is really old.

I thought that can be more stable that 1.11, but, if you invite me to do use it, sounds good for me

ferduz:
> And you are not specifying dockerhub for the oci image. It must start with docker.io to do so.

fixed now in case you want to fix this in this version

juliusvonkohout:
> > Is there a reason you are not using Kubeflow 1.11 ? 1.9 is really old.
> 
> I thought that can be more stable that 1.11, but, if you invite me to do use it, sounds good for me

No, most stable and supported is the latest release 1.11. using older releases is less stable as you can see from this minio problem for example.

PR #3360, opened 2026-02-26
Synchronize kubeflow pipelines manifests 2.16.0

juliusvonkohout:
@tarekabouzeid

### DIFF
--- a/.github/workflows/full_kubeflow_integration_test.yaml
+++ b/.github/workflows/full_kubeflow_integration_test.yaml
@@ -158,7 +158,7 @@ jobs:
 
     - name: V2 Pipeline Test
       run: |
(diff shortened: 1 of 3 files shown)


applications/pipeline/upstream/base/installs/multi-user/pipelines-profile-controller/test_sync.py
juliusvonkohout, line 1: need to review

applications/pipeline/upstream/base/installs/multi-user/pipelines-profile-controller/sync.py
juliusvonkohout, line 1: need to review

PR #3364, opened 2026-02-27
synchronize(dex): bump Dex manifests to v2.45.0

danish9039:
## ✏️ Summary of Changes
- Bump Dex sync script pin to `v2.45.0`
- Regenerate Dex image tag in `common/dex/base/deployment.yaml`
- Update Dex version entry in root `README.md`

## 📦 Dependencies
- None

## 🐛 Related Issues
- Follow-up component sync after cert-manager sync script updates

## ✅ Contributor Checklist
- [x] I have tested these changes with kustomize.
- [x] All commits are signed-off to satisfy the DCO check.
- [ ] I have considered adding my company to the adopters page.


juliusvonkohout:
Please check https://github.com/dexidp/dex/blob/master/examples/k8s/dex.yaml, why the tests are failing and whether 2 replicas would also work.

juliusvonkohout:
Thank you for the PR. Please check also whether https://github.com/danish9039/manifests/blob/sync/dex-next/common/dex/base/crds.yaml should be updated. And adjust the script to do so if needed.

juliusvonkohout:
Please also rebase to master, the new copilot instructions should block pr titles that say sync instead of synchronisation

danish9039:
@juliusvonkohout I checked the Dex `v2.45.0` failure locally,

first blocking issue was not replicas. The new Dex image`v2.45.0`   now runs as named user `dex:dex`, while our deployment keeps `runAsNonRoot: true` without a numeric UID/GID. Kubernetes then rejects the pod with `CreateContainerConfigError` because it cannot verify that the named image user is non-root, so i have added  numeric `runAsUser: 1001` and `runAsGroup: 1001` , and the Dex login flow passed

scripts/synchronize-dex-manifests.sh
```diff
@@ -27,6 +39,55 @@ else
...
+- kind: ServiceAccount
+  name: dex                 # Service account assigned to the dex pod.
+  namespace: auth           # The namespace dex is running in.
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: dex
+  namespace: auth
+EOF
```
juliusvonkohout: why is this in the synchronization script. That is for sure not the right place.

juliusvonkohout: should it be a resource overlay in common/dex ?

juliusvonkohout: We need to have an upstream folder for Dex as for the other components, to seperate what we get from upstream and what we add and patch on top. Also for cert-manager and others.

juliusvonkohout:
Thank you
/lgtm
/approve

PR #3372, opened 2026-03-01
block-impersonating-serviceaccounts

juliusvonkohout:
@tarekabouzeid


sendorrr:
Thank you for addressing and merging the fix.
Given that this change removes the ability to impersonate ServiceAccounts due to an RBAC misconfiguration, could you please confirm whether a CVE will be requested for this issue?
Additionally, please let me know how reporter credit will be handled in the advisory.
Thank you.

PR #3375, opened 2026-03-01
scripts: fix follow-up issues from #3371

danish9039:
## ✏️ Summary of Changes
This PR follows up on review feedback from `#3371` and fixes the remaining correctness and portability issues in the sync scripts.

Original PR:
- https://github.com/kubeflow/manifests/pull/3371

Changes included:
- fix `create_branch` in `scripts/library.sh` by removing the fragile empty-test branch check
- fix `scripts/synchronize-kubeflow-manifests.sh` so the CRUD web app source paths match the current upstream Kubeflow layout:
  - `components/crud-web-apps/jupyter/manifests`
  - `components/crud-web-apps/volumes/manifests`
  - `components/crud-web-apps/tensorboards/manifests`
- fix the CRUD web app README replacement paths in `scripts/synchronize-kubeflow-manifests.sh`
- narrow `commit_changes` usage so the Kubeflow and Knative sync scripts stage only their intended outputs instead of the whole repo
- make `replace_in_file()` in `scripts/synchronize-knative-manifests.sh` Darwin-safe

Review comments addressed:
- `scripts/library.sh`: fragile branch existence check
  - https://github.com/kubeflow/manifests/pull/3371#discussion_r2867782078
- `scripts/synchronize-kubeflow-manifests.sh`: CRUD web app README replacement mismatch
  - https://github.com/kubeflow/manifests/pull/3371#discussion_r2867782087
- `scripts/library.sh`: overly broad staging through `commit_changes`
  - https://github.com/kubeflow/manifests/pull/3371#discussion_r2867782064
- `scripts/synchronize-knative-manifests.sh`: non-portable `sed -i`
  - https://github.com/kubeflow/manifests/pull/3371#discussion_r2867782090

## 📦 Dependencies
- Follow-up to `#3371`

## 🐛 Related Issues
- Addresses remaining review feedback on `#3371`

## ✅ Contributor Checklist
- [x] I have tested these changes locally.
- [x] All commits are [_signed-off_](https://github.com/kubeflow/community/tree/master/dco-signoff-hook#signing-off-commits) to satisfy the DCO check.
- [ ] I have considered adding my company to the adopters page to support Kubeflow and help the community for my issue.

## Local Validation
Validated in a clean worktree and disposable repos:
- `bash -n` on all modified scripts
- actual run of `scripts/synchronize-kubeflow-manifests.sh` after fixing CRUD source paths
- actual run of `scripts/synchronize-knative-manifests.sh` with local `yq` available
[...]

PR #3378, opened 2026-03-03
knative: drop dead configmap filters

danish9039:
## ✏️ Summary of Changes
This PR is a small follow-up to   https://github.com/kubeflow/manifests/pull/3375#issuecomment-3986377704 and removes two dead `yq eval` filters from `scripts/synchronize-knative-manifests.sh`:

- `config-observability` filter on `common/knative/knative-eventing/base/upstream/in-memory-channel.yaml`
- `config-tracing` filter on `common/knative/knative-eventing/base/upstream/in-memory-channel.yaml`

These lines were originally intended to filter redundant Eventing ConfigMaps from the in-memory channel manifest, but for the current Knative inputs they are no longer affecting the final generated output.

I did **not** remove the other remaining `yq` calls:
- the job-name stabilization still has to stay, because removing it breaks the post-install job bases
- the comment-stripping calls are not needed for builds, but removing them rewrites several tracked upstream YAML files and would turn this into a larger churn PR

## 📦 Dependencies
- Follow-up to `#3375`
- Addresses mentor follow-up comment: `#issuecomment-3986377704`

## 🐛 Related Issues
- Knative sync-script cleanup after the `explode(.)` removal in `#3375`

## ✅ Contributor Checklist
- [x] I have tested these changes locally with kustomize.
- [ ] I have considered adding my company to the adopters page to support Kubeflow and help the community, since I expect help from the community for my issue.

## Local Validation
Validated in a clean worktree and disposable repos:
- inventory of all remaining `yq eval` calls in `scripts/synchronize-knative-manifests.sh`
- disposable-repo A/B runs against current `upstream/master`
- `kustomize v5.8.1 build` on:
  - `common/knative/knative-serving/base`
  - `common/knative/knative-eventing/base`
  - `common/knative/knative-serving-post-install-jobs/base`
  - `common/knative/knative-eventing-post-install-jobs/base`

Observed result:
- removing these two filter lines caused no output drift
- all 4 Knative builds still passed
- removing the job-name stabilization lines broke both post-install job base builds
- removing comment stripping caused large tracked YAML churn, so that was intentionally left out of this minimal follow-up

### DIFF
--- a/scripts/synchronize-knative-manifests.sh
+++ b/scripts/synchronize-knative-manifests.sh
@@ -39,8 +39,6 @@ yq eval -i '... comments=""' $DESTINATION_DIRECTORY/knative-eventing/base/upstre
 yq eval -i '... comments=""' $DESTINATION_DIRECTORY/knative-eventing/base/upstream/mt-channel-broker.yaml
 yq eval -i '... comments=""' $DESTINATION_DIRECTORY/knative-eventing-post-install-jobs/base/eventing-post-install.yaml
 yq eval -i 'select(.kind == "Job" and .metadata.generateName == "storage-version-migration-eventing-") | .metadata.name = "storage-version-migration-eventing"' $DESTINATION_DIRECTORY/knative-eventing-post-install-jobs/base/eventing-post-install.yaml
-yq eval -i 'select((.kind == "ConfigMap" and .metadata.name == "config-observability") | not)' $DESTINATION_DIRECTORY/knative-eventing/base/upstream/in-memory-channel.yaml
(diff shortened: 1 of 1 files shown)


danish9039:
@juliusvonkohout Addressed the comment : https://github.com/kubeflow/manifests/pull/3375#issuecomment-3986377704 ,  I checked the remaining `yq eval` calls more closely. The two filters for `config-observability` and `config-tracing` in `in-memory-channel.yaml` are redundant now and can be removed, but the job-name stabilization still has to stay because removing it breaks the post-install job bases. I did not remove the comment-stripping calls because they are not needed for builds, but dropping them rewrites several tracked upstream YAML files and would turn this into a much larger churn PR.

juliusvonkohout:
Thank you
/lgtm
/approve

PR #3237, opened 2025-09-08
Pipeline Helm Charts

kunal-511:
# Pull Request Template for Kubeflow Manifests

## ✅ Contributor Checklist
  - [ ] I have tested these changes with kustomize. See [Installation Prerequisites](https://github.com/kubeflow/manifests#prerequisites).
  - [ ] All commits are [_signed-off_](https://github.com/kubeflow/community/tree/master/dco-signoff-hook#signing-off-commits) to satisfy the DCO check.
  - [ ] I have considered adding my company to the adopters page to support Kubeflow and help the community, since I expect help from the community for my issue (see [1.](https://github.com/kubeflow/community/issues/833) and [2.](https://github.com/kubeflow/community/blob/master/ADOPTERS.md#adopters-of-kubeflow-platform)).

---     
 
> You can join the CNCF Slack and access our meetings at the [Kubeflow Community](https://www.kubeflow.org/docs/about/community/) website. Our channel on the CNCF Slack is here [**#kubeflow-platform**](https://app.slack.com/client/T08PSQ7BQ/C073W572LA2).

PR #3386, opened 2026-03-08
Improve Helm vs Kustomize resource comparison logic

ozalakshay:
✏️ Summary of Changes

Improved the manifest comparison utility used for validating Helm vs Kustomize outputs.

Key improvements include:

Namespace-aware resource identity logic to prevent cross-namespace key collisions when comparing manifests.

Namespace-aware handling of expected Helm-only resources, improving correctness for cases such as Katib webhook certificates and KServe AuthorizationPolicy resources.

Added a CLI interface using argparse with a --verbose mode that prints detailed per-resource and per-field differences to make debugging easier when comparing large manifest sets.

Code cleanup by removing unused imports and keeping only required typing symbols to improve maintainability.

These changes make the comparison logic more robust and provide better debugging capabilities when validating generated manifests.

📦 Dependencies

None.

🐛 Related Issues

No specific issue linked. This PR improves correctness and developer usability of the Helm vs Kustomize comparison script.

✅ Contributor Checklist

 I have tested these changes with kustomize. See Installation Prerequisites.

Verified Python syntax with python3 -m py_compile tests/helm_kustomize_compare.py

Full script execution was limited due to environment constraints (kustomize binary and PyYAML not installed).

 All commits are signed-off to satisfy the DCO check.

 I have considered adding my company to the adopters page.


juliusvonkohout:
@dhanishaphadate

Ahad-Dngwala:
@juliusvonkohout Sir As I was Going through this PR , the namespace-aware resource identity makes a lot of sense and should avoid cross-namespace collisions during comparison.

I had a quick question regarding the namespace parameter in get_expected_helm_extras():
If the CLI is run without specifying a namespace (since the argparse default is empty), this would generate keys like:
Secret//katib-webhook-cert

Would it be safer to either require namespace for certain components (like Katib) or derive the namespace directly from the manifests to avoid mismatches?

Just asking to understand the intended behavior....

Also the argparse CLI with --verbose is a nice addition. It should make debugging large manifest differences much easier when validating Helm vs Kustomize outputs.

tests/helm_kustomize_compare.py
```diff
@@ -279,21 +298,28 @@ def compare_manifests(kustomize_file: str, helm_file: str, component: str, scena
...
-    namespace = sys.argv[5] if len(sys.argv) > 5 and not sys.argv[5].startswith('--') else ""
-    
+    parser = argparse.ArgumentParser(
+        description="Compare manifests generated by Kustomize and Helm"
+    )
+    parser.add_argument("kustomize_file")
+    parser.add_argument("helm_file")
+    parser.add_argument("component")
+    parser.add_argument("scenario")
+    parser.add_argument("namespace", nargs="?", default="")
```
dhanishaphadate, line 308: If the namespace is not passed as arg, it defaults to an empty string (""). Is this case handled when resource keys are created for comparison?

tests/helm_kustomize_compare.py
```diff
@@ -279,21 +298,28 @@ def compare_manifests(kustomize_file: str, helm_file: str, component: str, scena
...
     if component not in ["katib", "model-registry", "kserve-models-web-app", "notebook-controller"]:
         print(f"ERROR: Unknown component: {component}")
         print("Supported components: katib, model-registry, kserve-models-web-app, notebook-controller")
         sys.exit(1)
     
     success = compare_manifests(kustomize_file, helm_file, component, scenario, namespace)
-    sys.exit(0 if success else 1) 
\ No newline at end of file
+    sys.exit(0 if success else 1) 
+    success = compare_manifests(kustomize_file, helm_file, component, scenario, namespace, args.verbose)
```
dhanishaphadate, line 322: This function call will never execute if the script exits with 0 after the previous successful comparison. Should we instead have a single compare_manifests function with a verbose flag, or conditionally call compare_manifests depending on whether the verbose argument is passed?

```
success = compare_manifests(kustomize_file, helm_file, component, scenario, namespace, args.verbose)
```

Ashish-Kumar-Dash:
Hey @ozalakshay , here are my notes for this PR where maybe you can take a look

* **Empty namespace bug:** (Echoing Copilot and dhanishaphadate here!) If `namespace` defaults to `""`, `get_expected_helm_extras` creates keys like `Secret//katib-webhook-cert`. This will never match the actual keys, which means unexpected Helm resources will fail silently. 
* **Unreachable code:** On line ~322, the second `compare_manifests` call (the one that actually uses `args.verbose`) will never run because the `sys.exit` right above it kills the script first. Let's just remove the first call and keep the verbose-aware one.
* **Missing component:** The CLI validation currently rejects any component not explicitly listed. However, `kubeflow-pipelines` is the most actively worked-on chart right now(PR #3237 ).  Although, this change can be taken after discussion with the maintainers(@juliusvonkohout ) whether it's a priority or not. I personally want to work on PR #3237 so I might be a bit biased :)
* **Testing:** I saw that you couldn't fully execute the script due to constraints. Since the namespace key changes touch the core  logic, could you run at least one full test and drop the results ?
I'm not a maintainer, but do let me know your thoughts on my notes!

PR #3390, opened 2026-03-13
Synchronize kserve models web application manifests v0.16.0

juliusvonkohout:
@tarekabouzeid @LogicalGuy77 
I have to wait for the OCI image builds here https://github.com/kserve/models-web-app/actions/runs/23042998303/job/66925611518 such that the tests can succeed.


juliusvonkohout:
/approve
lets wait for the tests

PR #3314, opened 2025-12-30
cleanup trainer overlay

juliusvonkohout:
/hold 
to be rebased and merged once https://github.com/kubeflow/trainer/pull/3259 is merged and synchronized
CC @andreyvelich

PR #3395, opened 2026-03-17
Update kubeflow/trainer manifests from v2.2.0-rc.0

juliusvonkohout:
@tarekabouzeid afterwards we might be able to merge https://github.com/kubeflow/manifests/pull/3314


Raakshass:
## CI Failure Root Cause Analysis — Update

I traced the CI failures through the `kubeflow-test-logs` diagnostic artifact and found two layered issues:

### Issue 1: Istio Sidecar Breaks Webhook Communication

The Trainer controller's reconciliation fails with `context deadline exceeded` because Istio sidecars intercept webhook traffic:

```json
{
  "error": "Internal error occurred: failed calling webhook \"mjobset.kb.io\": failed to call webhook: Post \"https://jobset-webhook-service.kubeflow-system.svc:443/mutate-jobset-x-k8s-io-v1alpha2-jobset?timeout=10s\": context deadline exceeded"
}
```
**Fix:** Added `sidecar.istio.io/inject: "false"` to both JobSet and Trainer controller pod templates in the upstream patches.

### Issue 2: Local Overlay Overrides Upstream Annotations

After the first fix, CI still failed — this time the **Trainer validation webhook** timed out:
```
Error from server (InternalError): failed calling webhook "validator.clustertrainingruntime.trainer.kubeflow.org": context deadline exceeded
```

**Root cause:** The local [applications/trainer/overlays/kustomization.yaml](cci:7://file:///C:/Users/asus%20vivoBook/Desktop/New%20folder%20%282%29/manifests-trainer-fix/applications/trainer/overlays/kustomization.yaml:0:0-0:0) has inline patches that set `spec.template.metadata.annotations` for both deployments. These patches override the annotations from the upstream `kubeflow-platform` overlay, **stripping `sidecar.istio.io/inject: "false"`** from the rendered output.

I verified this by running `kustomize build applications/trainer/overlays` locally — `sidecar.istio.io/inject` was absent from the rendered Deployment YAML despite being set in the upstream patches.

**Fix:** Propagated `sidecar.istio.io/inject: "false"` into the local overlay inline patches as well, and updated the Trainer's `excludeInboundPorts` to `"9443,10443"` (new status-server port from v2.2.0-rc.0).

### PR

#3396 (2 commits against this branch):
1. `fix: disable Istio sidecar injection on webhook-serving controllers` — upstream patches
2. `fix: propagate sidecar.istio.io/inject=false in local overlay patches` — local overlay

juliusvonkohout:
I think we also have to adjust our job. yaml according to what https://github.com/kubeflow/trainer/blob/master/examples/pytorch/image-classification/mnist.ipynb produces and make sure it satisfies pss restricted.

juliusvonkohout:
closed in favor of https://github.com/kubeflow/manifests/pull/3413

PR #3460, opened 2026-05-14
Fix ClusterRoleBinding namespace patch operation for workflow controller

### LINKED ISSUE #3459: workflow-controller ClusterRoleBinding patch fails on Kubernetes v1.35
### Version

* Kubeflow manifests: master
* Kubernetes: v1.35.0
* Distribution: MicroK8s
* OS: Ubuntu 26.04 LTS

### Detailed Description

While deploying Kubeflow using the manifests repository on a MicroK8s Kubernetes v1.35 cluster, the deployment repeatedly failed during the `kustomize build ... | kubectl apply` stage.

The issue was traced to the following file:

```text
applications/pipeline/upstream/third-party/argo/installs/cluster/workflow-controller-clusterrolebinding-patch.json
```

Current patch:

```json
[
  {
    "op": "replace",
    "path": "/subjects/0/namespace",
    "value": "kubeflow"
  }
]
```

The deployment fails because `/subjects/0/namespace` does not exist at patch application time.

According to RFC6902 JSON Patch semantics, the `replace` operation requires the target path to already exist. Since the field is absent, patch application fails.

Changing the operation from `replace` to `add` resolves the issue successfully:

```json
[
  {
    "op": "add",
    "path": "/subjects/0/namespace",
    "value": "kubeflow"
  }
]
```

After applying this change, the Kubeflow deployment completed successfully and all major components reached healthy running states.

### Steps to Reproduce

1. Prepare a Kubernetes v1.35 cluster (MicroK8s in this case).
2. Follow the Kubeflow manifests installation guide.
3. Run:

```bash
while ! kustomize build example | kubectl apply --server-side --force-conflicts -f -; do
  echo "Retrying to apply resources"
  sleep 20
done
```

4. Observe repeated failures during patch application.
5. Modify:

```text
applications/pipeline/upstream/third-party/argo/installs/cluster/workflow-controller-clusterrolebinding-patch.json
```

from:

```json
"op": "replace"
```

to:

```json
"op": "add"
```

6. Re-run deployment and observe successful completion.

### Expected Behavior

Kubeflow manifests should deploy successfully without requiring manual modification of the Argo ClusterRoleBinding patch.

### Additional Context

This issue appears to occur because the `namespace` field is absent before patch application, making the `replace` operation invalid for this path.

I am available to submit a PR for this fix.

astro-dally:
# ✏️ Summary of Changes

This PR updates the JSON patch operation used in:

```text
applications/pipeline/upstream/third-party/argo/installs/cluster/workflow-controller-clusterrolebinding-patch.json
```

The existing patch uses:

```json
"op": "replace"
```

for:

```text
/subjects/0/namespace
```

In a Kubernetes v1.35 MicroK8s environment, the `namespace` field was absent during patch application, causing the deployment to fail because RFC6902 requires the target path to exist for the `replace` operation.

This PR changes the operation to:

```json
"op": "add"
```

which allows the namespace field to be created when absent and resolves the deployment issue.

The Kubeflow deployment completed successfully after this change.

# 📦 Dependencies

None.

# 🐛 Related Issues

Fixes #3459 

# ✅ Contributor Checklist

* [x] I have tested these changes with kustomize. See [[Installation Prerequisites](https://github.com/kubeflow/manifests#prerequisites)](https://github.com/kubeflow/manifests#prerequisites).
* [x] All commits are [*[signed-off](https://github.com/kubeflow/community/tree/master/dco-signoff-hook#signing-off-commits)*](https://github.com/kubeflow/community/tree/master/dco-signoff-hook#signing-off-commits) to satisfy the DCO check.
* [ ] I have considered adding my company to the adopters page to support Kubeflow and help the community, since I expect help from the community for my issue.

---

### DIFF
--- a/applications/pipeline/upstream/third-party/argo/installs/cluster/workflow-controller-clusterrolebinding-patch.json
+++ b/applications/pipeline/upstream/third-party/argo/installs/cluster/workflow-controller-clusterrolebinding-patch.json
@@ -1,6 +1,6 @@
 [
   {
-    "op": "replace",
+    "op": "add",
     "path": "/subjects/0/namespace",
     "value": "kubeflow"
   }


juliusvonkohout:
@danish9039 please verify

danish9039:
@astro-dally  Thanks for the fix. I verified the PR on local MicroK8s `v1.35.0`: the new `add` patch renders and applies successfully with
  `kubectl apply --server-side --force-conflicts`, and `argo-binding` correctly points to `ServiceAccount` `argo` in namespace `kubeflow` and `kind cluster` validation has already been covered in the CI .

  One thing I could not reproduce locally is the original failure with the old `replace` operation. In my setup, both the focused Argo install and `kustomize build example` rendered successfully even before this change. Could you share the exact `kustomize`/`kubectl` versions and any relevant error output from the failing run? That would help confirm which patch path or  environment difference triggers the failure.

astro-dally:
Thanks @danish9039 for validating the patch and testing it locally.

The versions on my setup were:

```bash id="7dkq22"
kubectl: v1.35.5
kustomize: v5.7.1
```

I’m currently unable to reproduce the original failure again, as the Kubeflow setup is now running correctly.

danish9039:
> Thanks @danish9039 for validating the patch and testing it locally.
> 
> The versions on my setup were:
> 
> ```shell
> kubectl: v1.35.5
> kustomize: v5.7.1
> ```
> 
> I’m currently unable to reproduce the original failure again, as the Kubeflow setup is now running correctly.

Thanks for confirming. I also could not reproduce the original failure locally, including on MicroK8s `v1.35.0` with Kustomize `v5.7.1`; the old `replace` path rendered successfully here too.

  Given that, this seems environment/state-dependent rather than consistently reproducible. The change still looks safe to me because `add` is the more appropriate operation when `/subjects/0/namespace` may be absent, @juliusvonkohout can give the final decision.

juliusvonkohout:
Ah, you are modifying upstream folders. Please raise the PR against kubeflow/pipelines where this file is synchronized from.

PR #3505, opened 2026-06-15
Update kserve/kserve manifests to version v0.19.0

christian-heusel:
# Pull Request Template for Kubeflow Manifests

## ✏️ Summary of Changes

Update KServe to version v0.19.0.

## 📦 Dependencies

_None_

## 🐛 Related Issues

_None_

## ✅ Contributor Checklist

- [x] I have tested these changes with kustomize. See [Installation Prerequisites](https://github.com/kubeflow/community-distribution#prerequisites).
- [x] All commits are [_signed-off_](https://github.com/kubeflow/community/tree/master/dco-signoff-hook#signing-off-commits) to satisfy the DCO check.
- [x] I have considered adding my company to the adopters page to support Kubeflow and help the community, since I expect help from the community for my issue (see [1.](https://github.com/kubeflow/community/issues/833) and [2.](https://github.com/kubeflow/community/blob/master/ADOPTERS.md#adopters-of-kubeflow-platform)).

PR #3515, opened 2026-06-29
kserve: move platform resources to kserve namespace

danish9039:
## Summary of Changes

This follow-up moves KServe-owned platform resources from `kubeflow` to the
upstream `kserve` namespace.

- Install the standard upstream KServe bundle in `kserve`, including
  controllers, webhooks, certificates, namespaced RBAC, and
  `inferenceservice-config`.
- Move the KServe Models Web Application and its NetworkPolicy to `kserve`
  while preserving routing through `kubeflow/kubeflow-gateway`.
- Keep user `InferenceService` resources and serving workloads in Profile
  namespaces.
- Keep direct Kustomize installation self-contained with `Namespace/kserve`;
  add equivalent namespace ownership and restricted Pod Security labels to the
  `kubeflow-namespaces` foundation chart.
- Disable Istio sidecar injection explicitly on KServe controller pod templates
  while retaining explicit injection for the Models Web Application.
- Preserve Kubeflow aggregate roles and disable the restricted-Pod-Security
  incompatible local model node agent.
- Update installation tests, workflow diagnostics, chart parity, synchronized
  version handling, and upgrade cleanup guidance.

## Ownership Boundaries

The component Kustomization owns `Namespace/kserve` for direct Kustomize
installation. Helm component charts do not create namespaces; the
`kubeflow-namespaces` foundation chart owns `Namespace/kserve` for Helm
installation.

This pull request only aligns the existing Models Web Application chart with
the relocation. Generated KServe control-plane charts and their synchronization
pipeline remain separate future Helm work.

## Upgrade Notes

Controllers in the old and new namespaces use different leader-election
Leases, so the old `kubeflow` controllers must be scaled to zero before the new
controllers start. The old Models Web Application route and namespaced
resources must also be removed to prevent duplicate routing and stale
resources.

The documented cleanup does not delete KServe CRDs, user `InferenceService`
objects, Profile namespaces, or model-serving workloads. Existing Helm
installations must uninstall the Models Web Application release from
`kubeflow` and install a new release in `kserve`, because Helm cannot move a
release between namespaces.

## Related Issues

Follow-up to #3505 and the namespace recommendation in the pull request
discussion.

## Contributor Checklist

- [x] I have tested these changes with Kustomize. See [Installation Prerequisites](https://github.com/kubeflow/community-distribution#prerequisites).
- [x] All commits are [signed off](https://github.com/kubeflow/community/tree/master/dco-signoff-hook#signing-off-commits) to satisfy the DCO check.
- [ ] I have considered adding my company to the adopters page to support Kubeflow and help the community, since I expect help from the community for my issue (see [1](https://github.com/kubeflow/community/issues/833) and [2](https://github.com/kubeflow/community/blob/master/ADOPTERS.md#adopters-of-kubeflow-platform)).

## Verification

- `python3 tests/kserve_manifest_test.py`
- pre-commit hooks for every changed or added file
- `git diff --check upstream/master...HEAD`
- shell syntax validation for the affected test and synchronization scripts
- Kustomize rendering for the KServe control plane, Models Web Application,
  shared namespaces, and complete example
- strict `kubeconform` validation of both KServe renderings
- Helm linting for the namespace foundation and Models Web Application charts
- complete `tests/helm_kustomize_compare_all.sh all` parity matrix
- independent adversarial review of RBAC, webhooks, certificates, routing,
  NetworkPolicies, Pod Security, Istio injection, upgrade cleanup, and Helm
  ownership

Not run locally: live-cluster admission, runtime, and in-place migration tests
because no active Kubernetes context was available.

## Related Issues

Tracking issue:
https://github.com/kubeflow/community-distribution/issues/3561


danish9039:
@juliusvonkohout  for review

README.md
```diff
@@ -351,6 +351,8 @@ Optionally, you can install Knative Eventing, which can be used for inference re
 ./tests/kserve_install.sh
 ```
 
+The KServe control plane is installed in the `kserve` namespace. The KServe UI remains in the `kubeflow` namespace.
```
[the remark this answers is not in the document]
danish9039: Fixed: that sentence now states that the control plane and the Models Web Application are both installed in `kserve`, that user `InferenceService` resources remain in the profile namespaces, and that the Models Web Application is still reached through the Kubeflow gateway.

tests/helm_kustomize_compare.sh
```diff
@@ -161,6 +161,7 @@ case "$COMPONENT" in
             "$ROOT_DIRECTORY/common/istio/istio-namespace/base"
             "$ROOT_DIRECTORY/common/oauth2-proxy/base"
             "$ROOT_DIRECTORY/common/dex/base"
+            "$ROOT_DIRECTORY/applications/kserve/kserve"
```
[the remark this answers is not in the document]
danish9039: Fixed: `applications/kserve/kserve/**` is now watched by `.github/workflows/helm-kustomize-comparison.yml`, so a change to that rendering alone no longer skips the comparison.

tests/kserve_manifest_test.py
```diff
@@ -0,0 +1,270 @@
...
+    def test_relocation_leaves_no_resource_in_the_kubeflow_namespace(self):
+        """A resource left behind in kubeflow still installs and still runs.
+
+        Nothing in the live installation fails, so only a static sweep of the
+        rendered output can prove the relocation is complete.
+        """
+        for resource in self.kserve_resources:
+            self.assertNotEqual(
+                resource.get("metadata", {}).get("namespace"),
+                "kubeflow",
```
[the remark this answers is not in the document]
danish9039: Fixed: the sweep now asserts the exact namespace instead of the absence of the old one, so a resource that declares no namespace fails rather than passing. Verified by mutation: a namespaced resource with the namespace removed passes the old assertion and fails the new one. The cluster scope rule is now shared with the cluster scope test.

danish9039:
@juliusvonkohout  Addressed the Copilot review comments

danish9039:
@juliusvonkohout ready to merge

applications/kserve/kserve/kustomization.yaml
```diff
@@ -1,110 +1,96 @@
...
-      value: kubeflow/llmisvc-serving-cert
+    - op: add
+      path: /spec/container/securityContext
+      value:
+        allowPrivilegeEscalation: false
+        capabilities:
+          drop:
+          - ALL
+        runAsNonRoot: true
+        runAsUser: 1000
```
juliusvonkohout:
```suggestion
```

applications/kserve/README.md
```diff
@@ -16,4 +20,107 @@ When using KServe with path-based routing in a KubeFlow deployment, you may enco
...
+  deployment.apps/kserve-localmodel-controller-manager \
+  deployment.apps/llmisvc-controller-manager \
+  daemonset.apps/kserve-localmodelnode-agent \
+  lease.coordination.k8s.io/kserve-controller-manager-leader-lock \
+  lease.coordination.k8s.io/llminferenceservice-kserve-controller-manager \
+  certificate.cert-manager.io/serving-cert \
+  certificate.cert-manager.io/llmisvc-serving-cert \
+  certificate.cert-manager.io/localmodel-serving-cert \
+  issuer.cert-manager.io/selfsigned-issuer
+```
```
christian-heusel: Why are these notes duplicated here from the main README.md? 🤔 I think they can be dropped here. 

Also I think such notes should always be versioned, otherwise it's unsure on which upgrades they apply 😅

applications/kserve/kserve/aggregated-roles.yaml
christian-heusel, line 1: Where are those roles coming from and why do we need to introduce them? 🤔 AFAIU the upstream overlay already creates rules like this in the kubeflow overlay 😅

juliusvonkohout: The kubeflow overlay is not that well maintained. This is why we wanted to switch to the regular one, also in light of the helm efforts.

applications/kserve/README.md
juliusvonkohout: Yes please remove duplicates

juliusvonkohout:
Please check https://github.com/kubeflow/community-distribution/pull/3515#pullrequestreview-4897645229

juliusvonkohout:
Please rebase after https://github.com/kubeflow/community-distribution/pull/3576/

applications/kserve/kserve/kustomization.yaml
danish9039: Removed. To record why it is safe: the restricted Pod Security Standard requires `runAsNonRoot`, not a particular identifier, and `kserve/storage-initializer:v0.19.0` already declares `USER 1000`, which is the numeric non-root user that `runAsNonRoot` needs in order to be verified. Upstream applies the same rule: of the twelve serving runtimes it ships, only `tensorflow-serving`, `torchserve` and `tritonserver` set `runAsUser`, and those are exactly the three whose images do not declare a numeric non-root user.

applications/kserve/README.md
danish9039: Removed, and the section now points at the single copy in the root README. Your second point is right as well: these steps are version-specific, which is a further reason to keep them with the other release upgrade notes rather than in the component directory.

applications/kserve/kserve/aggregated-roles.yaml
danish9039: Both observations are correct, and the resolution sits between them. These roles previously arrived embedded in `upstream/kserve_kubeflow.yaml`, three of them, and that file is no longer installed. The standard `upstream/kserve.yaml` embeds none, so dropping the Kubeflow flavour drops the roles with it.

The standalone `upstream/aggregated-roles.yaml` is not a substitute: it grants only `inferenceservices` plus the Knative resources, and omits `inferencegraphs`, `llminferenceservices`, `servingruntimes` and `trainedmodels`. Adopting it would silently remove permissions that users have today. I have added a comment to the file recording this so the next reader does not have to work it out again.

danish9039:
@juliusvonkohout addressed the comments

juliusvonkohout:
/lgtm

Although it is still a bit too verbose. Some information in the readme for example is just llm verbosity and not minimal enough. Every line you add has a cost on someone else and one of the best qualities is deleting and removing instead of adding.

danish9039:
> /lgtm
> 
> 
> 
> Although it is still a bit too verbose. Some information in the readme for example is just llm verbosity and not minimal enough. Every line you add has a cost on someone else and one of the best qualities is deleting and removing instead of adding.

I will keep this in mind

PR #3529, opened 2026-07-05
Synchronize kserve UI manifests v1.0.0

juliusvonkohout:
@danish9039 @Raakshass  for lgtm


scripts/synchronize-kserve-ui-manifests.sh
```diff
@@ -1,12 +1,12 @@
 #!/usr/bin/env bash
-# This script helps to create a PR to update the KServe Models Web App manifests
+# This script helps to create a PR to update the KServe UI manifests
 SCRIPT_DIRECTORY=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 source "${SCRIPT_DIRECTORY}/library.sh"
 setup_error_handling
-COMPONENT_NAME="models-web-app"
+COMPONENT_NAME="kserve-ui"
```
christian-heusel, line 6: Why are we renaming this given that the upstream repository is still named the same? 🤔

AGENTS.md
christian-heusel, line 1: Why does this PR include an agents.md? 🤔 This seems unrelated to the change itself

CLAUDE.md
christian-heusel, line 1: If there are no additional contents in this file we should rather just symlink it.

scripts/synchronize-kserve-ui-manifests.sh
juliusvonkohout: I am changing the upstream one too.

AGENTS.md
juliusvonkohout: Because this structure should work for most AI assistants now.

CLAUDE.md
juliusvonkohout: Depending on the software consuming it it could lead to problems. I am not against it, if it works reliably.

juliusvonkohout:
/hold
i need to change the image tag

PR #3530, opened 2026-07-05
Fix kserve-ui modified upstream folders

juliusvonkohout:
@Raakshass @danish9039 for lgtm


applications/kserve/kserve-ui/kustomization.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- ./upstream/overlays/kubeflow
+
+images:
+  - name: ghcr.io/kserve/models-web-app
+    newTag: 1.0.0
```
christian-heusel, line 8: Maybe add a comment here to remove this on next sync 🤔

christian-heusel (approved):

applications/kserve/kserve-ui/kustomization.yaml
juliusvonkohout: That depends on which direction we  want to take upstream. If others can help with that upstream we could do so. Are you interested ? Right now it is in line with upstream.

PR #3536, opened 2026-07-07
Update kubeflow/pipelines manifests from 9a1a11a

juliusvonkohout:
Test PR


christian-heusel:
@juliusvonkohout what are we testing here? 👀

PR #3539, opened 2026-07-09
proper build version

juliusvonkohout:
@Raakshass @danish9039 for lgtm


christian-heusel:
@juliusvonkohout why are those changes done with regards to the Pipeline overlay and not dashboard? 🤔

juliusvonkohout:
Because the application crd is owned by upstream pipelines. I do not like it myself. It should be moved into the dashboard repository. Or better removed and just get the build string from an environment variable.

PR #3581, opened 2026-08-12
fix(pipeline): allow kube-apiserver to reach the PipelineVersion admission webhook

### LINKED ISSUE #3580: ml-pipeline NetworkPolicy blocks kube-apiserver calls to the PipelineVersion admission webhook (8443)
## Title
`ml-pipeline` NetworkPolicy blocks kube-apiserver calls to the PipelineVersion admission webhook (8443)

## Kubeflow Manifests version
commit `b88f4eec472070d7d58c58468a37791cad76e9bf` (master, 2026-08-04) — "Update OAuth2-Proxy to v7.15.3, Cert Manager to v1.21.1 and Trainer to v2.3.0-rc.3 (#3572)"

## Environment
- `kind` v0.32.0, node image Kubernetes v1.36.1, kustomize v5.8.1
- KFP api-server image: `ghcr.io/kubeflow/kfp-api-server:2.17.0`
- Overlay: `applications/pipeline/upstream/env/dev-kind` (built on `env/cert-manager/platform-agnostic-k8s-native`), K8s-native pipelines mode (`Pipeline`/`PipelineVersion` CRDs, `pipelines.kubeflow.org/v2beta1`)
- cert-manager for webhook TLS

## What happened
`kubectl apply` of a `Pipeline` + `PipelineVersion` pair: the `Pipeline` object is created immediately (no admission webhook targets it), but `PipelineVersion` creation fails:

```
Error from server (InternalError): error when creating "workshop-pipeline-v1.yaml": Internal error occurred:
failed calling webhook "pipelineversions.pipelines.kubeflow.org": failed to call webhook:
Post "https://ml-pipeline.kubeflow.svc:8443/webhooks/mutate-pipelineversion?timeout=30s":
net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
```

Same symptom shows up internally at `ml-pipeline` startup when it tries to self-register the bundled sample pipelines (`config.go:247`), which also create `PipelineVersion` objects and hit the same webhook.

## Root cause
`common/kubeflow-namespace/base/kubeflow/ml-pipeline.yaml` defines a single `NetworkPolicy` named `ml-pipeline` that gates **all** ingress to the `ml-pipeline` pod (all 3 ports: 8888/8887/8443), restricted to:

```yaml
spec:
  podSelector:
    matchExpressions:
      - key: app
        operator: In
        values: [ml-pipeline]  # comment says "just the apiserver" — refers to the pod's own role
  ingress:
    - from:
        - namespaceSelector: {app.kubernetes.io/part-of In [kubeflow-profile]}
        - namespaceSelector: {kubernetes.io/metadata.name In [istio-system]}
```

This predates the K8s-native pipelines feature, where `ml-pipeline` itself now serves the `pipelineversions.pipelines.kubeflow.org` mutating/validating admission webhooks on port 8443. Admission webhook calls are made directly by `kube-apiserver`, which is not a pod and therefore does not match any `podSelector`/`namespaceSelector`-based ingress rule. Once any `NetworkPolicy` with `policyTypes: [Ingress]` selects a pod, unmatched traffic is dropped by default — silently, which is why the symptom is a **timeout**, not a connection refusal.

`default-allow-same-namespace` (also in `kubeflow` namespace) does not help either, since it only allows traffic `from: [{podSelector: {}}]` — i.e. from other pods, which again excludes the apiserver.

### This is inconsistent with every other webhook-serving component in the same manifests tree
All other components that serve their own admission webhooks correctly scope their `NetworkPolicy` to **port-only** rules (no `from`, i.e. explicitly open to any source, since the webhook's own TLS/CA is the actual trust boundary):

- `kubeflow` ns `katib-controller` → `{ingress: [{ports: [{port: 8443}]}], podSelector: ...}`
- `kubeflow` ns `spark-operator-webhook` → `{ingress: [{ports: [{port: 9443}]}], podSelector: ...}`
- `kubeflow-system` ns `jobset-webhook` → `{ingress: [{ports: [{port: 9443}]}], podSelector: ...}`
- `kubeflow-system` ns `trainer-webhook` → `{ingress: [{ports: [{port: 9443}]}], podSelector: ...}`

`ml-pipeline` is the only one that folds its webhook port into the general-purpose API NetworkPolicy instead of getting the same port-only treatment.

Also notable: `applications/pipeline/upstream/env/dev-kind/kustomization.yaml` already patches the `seaweedfs` NetworkPolicy in this same overlay to add an `ipBlock: 0.0.0.0/0` rule scoped to its port — so the "open this one port broadly, on this overlay" pattern is already precedented right next to the file that needs the same treatment for `ml-pipeline`.

## Verification
Reproduced and bisected directly against the live cluster:
- Same-node / same-namespace pod → `ml-pipeline:8443`: succeeds.
- `kube-apiserver`'s own network path (control-plane node's host network) → `ml-pipeline` pod IP:8443: hangs until timeout.
- Same request → a pod in `kube-system` (no `NetworkPolicy` applied) on the same worker node: succeeds instantly.
- After patching the `ml-pipeline` `NetworkPolicy` to add a port-only ingress rule for 8443 (matching the pattern used by `katib-controller`/`spark-operator-webhook`/`jobset-webhook`/`trainer-webhook`), the same request succeeds instantly and `kubectl apply` of the `PipelineVersion` succeeds.

## Suggested fix
Add a port-only ingress rule to `common/kubeflow-namespace/base/kubeflow/ml-pipeline.yaml`, consistent with the other webhook `NetworkPolicy`s:

```yaml
spec:
  ingress:
    - from:
        - namespaceSelector: {app.kubernetes.io/part-of In [kubeflow-profile]}
        - namespaceSelector: {kubernetes.io/metadata.name In [istio-system]}
    - ports:
        - protocol: TCP
          port: 8443
```

Happy to open a PR with this change if useful.

## Repro steps
1. `kind` cluster (3 nodes), cert-manager installed, `NetworkPolicy` enforcement active (e.g. via `kindnetd` netpol support or any CNI enforcing them).
2. Apply `applications/pipeline/upstream/env/cert-manager/cluster-scoped-resources`, then `applications/pipeline/upstream/env/dev-kind`.
3. `kubectl apply -f` any `Pipeline` + `PipelineVersion` manifest.
4. Observe the `PipelineVersion` create call time out with the error above, while the `Pipeline` object is created successfully.

tyetladd:
Fixes #3580

## Problem
The `ml-pipeline` NetworkPolicy only allows ingress from `kubeflow-profile` namespaces and `istio-system`. This predates the K8s-native pipelines feature, where `ml-pipeline` itself serves the `pipelineversions.pipelines.kubeflow.org` mutating/validating admission webhooks on port 8443. Admission webhook calls come directly from `kube-apiserver`, which is not a pod and cannot be matched via `podSelector`/`namespaceSelector`, so those calls are silently dropped and time out — every `PipelineVersion` create/update fails while the parent `Pipeline` object (no webhook) succeeds:

```
failed calling webhook "pipelineversions.pipelines.kubeflow.org": failed to call webhook:
Post "https://ml-pipeline.kubeflow.svc:8443/webhooks/mutate-pipelineversion?timeout=30s":
net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
```

## Fix
Add a port-only ingress rule for 8443 to `common/kubeflow-namespace/base/kubeflow/ml-pipeline.yaml`, matching the existing pattern already used by `katib-controller`, `spark-operator-webhook`, `jobset-webhook` and `trainer-webhook` — all of them scope their webhook port with a `ports`-only rule (no `from`), since the webhook's own TLS/CA is the actual trust boundary, not network segmentation.

## Verification
Reproduced and bisected against a live cluster (see linked issue for full detail): confirmed `kube-apiserver`'s network path to the `ml-pipeline` pod's webhook port hangs until timeout while identical traffic to pods without a `NetworkPolicy` succeeds instantly; after applying the equivalent live patch, the same request succeeds immediately and `kubectl apply` of a `PipelineVersion` succeeds.


christian-heusel (commented):
Thanks a lot for your contribution! 🤗 

Did you use AI in any capacity while creating these changes? I see that your commits do not carry any `Assisted-By: <...>` / `Co-authored-by: <...>` tags, so I wanted to make sure that you're aware of the [Kubeflow AI Policy](https://www.kubeflow.org/docs/about/ai_policy/). 🤗 🤖

/ok-to-test

christian-heusel (approved):

common/kubeflow-namespace/base/kubeflow/ml-pipeline.yaml
```diff
@@ -24,5 +24,14 @@ spec:
                 operator: In
                 values:
                   - istio-system
```
juliusvonkohout, line 26: is this still needed then or can it be dropped?

tyetladd: My read: it's probably still needed, but flagging my reasoning in case I'm missing context.

The `ml-pipeline` pod serves three ports — `8888` (HTTP API), `8887` (gRPC), and `8443` (admission webhook, added by the k8s-native overlay). This `from:` rule has no `ports` field, so as far as I can tell it's the only thing restricting `8888`/`8887` to `kubeflow-profile`/`istio-system` namespaces. The new `ports: [8443]` rule is scoped only to the webhook port.

If that's right, dropping the `from:` block would open the API/gRPC ports to any namespace, which seems like an unrelated regression rather than cleanup here. Let me know if I'm missing something about how those ports are otherwise protected.

PR #3590, opened 2026-08-25
test(spark): add Spark Connect from Workspace integration test

vikas-saxena02:
**What this PR does / why we need it**:

Adds an integration test covering interactive Spark from a Kubeflow Workspace: the Workspace creates a `SparkConnect` session through the Kubeflow SDK, runs a distributed query against it, and cleans up. This is the Spark and Workspaces combination discussed on Slack, modelled on `workspaces_pipeline_run_test.yaml`.

It also gives the aggregated Spark RBAC from #3586 an end-to-end exercise rather than only the `kubectl auth can-i` assertions in `spark_test.sh`.

**Files**

- `.github/workflows/workspaces_spark_connect_test.yaml` — installs Istio, cert-manager, the Profile Controller, multi-tenancy, Spark Operator and Workspaces, then runs the test. Includes the log collection and artifact upload from the workspaces pipeline workflow.
- `tests/workspaces_spark_connect_test.sh` — creates the Workspace, waits for `Running`, installs the SDK, copies the script in and executes it.
- `tests/spark_connect_from_workspace.py` — runs inside the Workspace: creates the session, runs a grouped aggregation, asserts the result, deletes the session.
- `tests/workspacekind.spark.test.yaml` — a `jupyterlab-spark` WorkspaceKind granting `kubeflow-spark-edit`.
- `tests/workspace.spark.test.kubeflow-user-example-com.yaml` — the Workspace and its PVC.

**Design notes**

*Separate WorkspaceKind.* `jupyterlab-spark` rather than extending the existing `jupyterlab` kind, so the two tests do not collide. It is a copy of `workspacekind.test.yaml` with one functional change: `serviceAccount.clusterRoles` grants `kubeflow-spark-edit` instead of the two pipelines roles. That follows the note in the existing fixture — since Workspaces v2.0.0-beta.0 the per-Workspace ServiceAccount no longer inherits the Kubeflow ClusterRoles, so each application binds what it needs, minimally.

*Two identities.* The Workspace runs as `ws-spark-test`, which `kubeflow-spark-edit` lets create `SparkConnect` resources. The Spark driver pod is separate: it creates and watches executor pods, which that role does not cover, so it runs as `default-editor` — the same choice `applications/spark/sparkapplication_example.yaml` makes for its driver. Verified `default-editor` holds `create`, `delete`, `get`, `list` and `watch` on pods in a profile namespace.

*Sidecar exclusion.* The driver and executors carry `sidecar.istio.io/inject: "false"`, as `sparkapplication_example.yaml` does. Spark's driver-executor traffic on ports 7078/7079 does not survive the mesh — executors register with the driver and then fail fetching broadcast blocks, so tasks never complete. Excluding just those ports from interception is not sufficient; the pods have to be outside the mesh.

*Version pinning.* The SDK itself is installed unpinned, then the Spark Connect client is pinned to `constants.DEFAULT_SPARK_VERSION` read back from the installed SDK. So the client always matches the server the SDK will provision, and neither has to be hard-coded here. If that constant moves in a future SDK release the test fails at that step, which is a clear place to find out.

*RBAC assertion before the session.* `kubectl auth can-i create sparkconnects` as the Workspace ServiceAccount runs first, so an RBAC regression fails in seconds with a clear message rather than as a Spark timeout several minutes later.

*Executor pod template.* The template declares a container named `spark-kubernetes-executor` even though the operator supplies the image, because Spark submits with `spark.kubernetes.executor.podTemplateContainerName` and looks it up by name. Without it the driver exits 1 at startup with nothing useful on stdout. There is a comment in the script explaining this.

**Verification**

Run locally against Kubeflow v1.10.0 on Kubernetes v1.35.0 (KinD), Spark Operator 2.5.2:

```
+ kubectl auth can-i create sparkconnects --as=system:serviceaccount:kubeflow-user-example-com:ws-spark-test -n kubeflow-user-example-com
+ grep -qx yes
SDK default Spark version: 4.0.4
INFO - Creating a Spark Connect session in namespace kubeflow-user-example-com
INFO - Session ready: kubeflow-user-example-com/spark-connect-849053c9 state=SparkConnectState.READY (2s)
INFO - In-cluster connect URL: sc://spark-connect-849053c9-server.kubeflow-user-example-com.svc.cluster.local:15002
INFO - Connected to Spark 4.0.4 (client 4.0.4)
INFO - Aggregated 100000 rows into 8 buckets
INFO - Deleting Spark Connect session spark-connect-849053c9
INFO - Spark Connect session from Workspace verified.
```

The server and executor pods come up `1/1`, confirming the sidecar exclusion applies, and the executor reaches `Completed`. The namespace is left with no `SparkConnect` resources.

**Checklist**:

- [x] I have conducted a self-review of my own code.
- [ ] I have updated documentation accordingly.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] Existing unit tests pass locally with my changes.


.github/workflows/workspaces_spark_connect_test.yaml
```diff
@@ -0,0 +1,100 @@
+name: Test Spark Connect from Workspace
+on:
+  pull_request:
```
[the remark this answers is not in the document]
vikas-saxena02: fixed applies

tests/workspaces_spark_connect_test.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+set -euxo pipefail
+
+KF_PROFILE=${1:-kubeflow-user-example-com}
+
+kubectl apply -f tests/workspacekind.spark.test.yaml
+kubectl apply -f tests/workspace.spark.test.kubeflow-user-example-com.yaml
```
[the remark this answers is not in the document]
vikas-saxena02: fixed applied

tests/workspacekind.spark.test.yaml
```diff
@@ -0,0 +1,113 @@
+# Test-only copy of the upstream JupyterLab WorkspaceKind sample
+# (applications/workspaces/upstream/controller/samples/jupyterlab_v1beta1_workspacekind.yaml),
+# trimmed to what the Spark Connect test needs. The only functional difference
+# from the upstream sample is the `serviceAccount.clusterRoles` grant below.
```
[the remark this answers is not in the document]
vikas-saxena02: fixed applied

.github/workflows/workspaces_spark_connect_test.yaml
```diff
@@ -0,0 +1,100 @@
...
+    - tests/workspace.spark.test.kubeflow-user-example-com.yaml
+    - tests/istio*
+    - tests/multi_tenancy_install.sh
+    - tests/profile_controller_install.sh
+    - tests/oauth2-proxy_install.sh
+    - tests/cert_manager_install.sh
+    - common/cert-manager/**
+    - common/oauth2-proxy/**
+    - common/istio*/**
+    - common/kubeflow-namespace/**
```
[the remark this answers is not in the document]
vikas-saxena02: fixed applies

.github/workflows/workspaces_spark_connect_test.yaml
```diff
@@ -0,0 +1,105 @@
+name: Test Spark Connect from Workspace
+on:
+  push:
+    branches:
+    - mster
+  pull_request:
```
christian-heusel, line 3: This is not how the other actions trigger, also there is a typo in `branches`:
```
$ yq ".on | keys" .github/workflows/workspaces*.yaml
- pull_request
- pull_request
```

.github/workflows/workspaces_spark_connect_test.yaml
```diff
@@ -0,0 +1,105 @@
...
+
+    - name: Install KF Multi Tenancy
+      run: ./tests/multi_tenancy_install.sh
+
+    - name: Create KF Profile
+      run: kustomize build common/user-namespace/base | kubectl apply -f -
+
+    - name: Install Spark Operator
+      run: |
+        chmod u+x tests/*.sh
```
christian-heusel: Why is this needed? 🤔 Please remove it so that the file is directly checked out with `+x` permissions!

tests/spark_connect_from_workspace.py
```diff
@@ -0,0 +1,104 @@
...
+# excluded from injection, as sparkapplication_example.yaml already does.
+NO_SIDECAR = {"sidecar.istio.io/inject": "false"}
+
+
+def main() -> None:
+    client = SparkClient(backend_config=KubernetesBackendConfig(namespace=NAMESPACE))
+
+    logger.info("Creating a Spark Connect session in namespace %s", NAMESPACE)
+    spark = None
+    try:
```
christian-heusel: Why are we wrapping the connect in a `try` block here? 🤔

tests/workspaces_spark_connect_test.sh
```diff
@@ -0,0 +1,46 @@
...
+# so an RBAC regression fails here with a clear message rather than inside Spark.
+kubectl auth can-i create sparkconnects \
+  --as="system:serviceaccount:${KF_PROFILE}:ws-spark-test" \
+  -n "${KF_PROFILE}" | grep -qx yes
+
+# The jupyter-scipy image does not ship the Kubeflow SDK. Installing it here rather
+# than baking a dedicated image keeps this test self-contained. The resolver reports
+# protobuf conflicts against the pre-installed kfp packages; they do not affect Spark.
+kubectl -n "${KF_PROFILE}" exec "${WORKSPACE_POD}" -- \
+  python -m pip install --quiet "kubeflow[spark]"
```
christian-heusel: This install should be versioned, otherwise the test could be flaky. Also the jupyter-scipy image ships the kubeflow sdk with the 1.11 release, the samples have just not been updated yet: 
https://github.com/kubeflow/notebooks/blob/177b098656264a36af71ea27d098c2ad14999eec/components/example-notebook-servers/jupyter-scipy/requirements.txt#L3

So you can just use the following in the workspacekind:
```diff
-              image: "ghcr.io/kubeflow/kubeflow/notebook-servers/jupyter-scipy:v1.10.0"
+              image: "ghcr.io/kubeflow/kubeflow/notebook-servers/jupyter-scipy:v1.11.0"
```

tests/workspaces_spark_connect_test.sh
```diff
@@ -0,0 +1,46 @@
...
+
+# Pin the Spark Connect client to the server version the SDK will provision, so the two
+# cannot drift when either the SDK default or the PyPI resolution changes. If this
+# constant moves in a future SDK release the test fails here, which is the right place
+# to find out.
+SPARK_VERSION="$(kubectl -n "${KF_PROFILE}" exec "${WORKSPACE_POD}" -- python -c \
+  'from kubeflow.spark.backends.kubernetes import constants; print(constants.DEFAULT_SPARK_VERSION)')"
+echo "SDK default Spark version: ${SPARK_VERSION}"
```
christian-heusel: Does `kubeflow[spark]` not already do this? 🤔

christian-heusel (changes requested):

tests/workspaces_spark_connect_test.sh
vikas-saxena02: `kubeflow[spark]` has a default version which was coming up as 4.0.4 but `pip install pyspark` was going to the latest version of pyspark. so i added the extra logic of  installing pyspark binaries which are the same version as default spark verstion used by sdk to align the two and avoid any issues in future

.github/workflows/workspaces_spark_connect_test.yaml
vikas-saxena02: this was done to address copilot review but i have reverted it back to align with others

tests/spark_connect_from_workspace.py
vikas-saxena02: It is for cleanup on failure. Copilot flagged that if one of the assertions below fails, the `SparkConnect` resource is left behind — in CI that does not matter since the cluster is torn down, but it bites when running the script against an existing cluster locally, which is how I was developing it.

That said, it does add a level of nesting for something CI never needs. Happy to drop the `try`/`finally` and just delete at the end if you would rather keep it flat — the leak only affects local runs.

@juliusvonkohout whats you take on it?

tests/workspaces_spark_connect_test.sh
vikas-saxena02: Thanks — taking the image bump either way.

The SDK it ships is the base package rather than the `[spark]` extra, so `SparkClient` still cannot import:

```console
$ docker run --rm ghcr.io/kubeflow/kubeflow/notebook-servers/jupyter-scipy:v1.11.0 \
    python -c "import kubeflow, importlib.util; print('kubeflow', kubeflow.__version__); \
      [print(m, importlib.util.find_spec(m) is not None) for m in ('pyspark','kubeflow_spark_api')]; \
      from kubeflow.spark import SparkClient"
kubeflow 0.4.0
pyspark False
kubeflow_spark_api False
...
  File ".../kubeflow/spark/api/spark_client.py", line 20, in <module>
    from pyspark.sql import SparkSession
ModuleNotFoundError: No module named 'pyspark'
```

So the install stays, though the image bump means it only needs to add the Spark dependencies rather than the whole SDK. It also pins the SDK for us at 0.4.0, which addresses the flakiness concern — the version now comes from the image rather than from whatever PyPI resolves to on the day.

On pinning the client: the awkward part is that the Spark Connect client has to match the server version the SDK provisions, and the SDK picks that from its own `DEFAULT_SPARK_VERSION`. The script therefore reads that constant back from the installed SDK and pins `pyspark-connect` to it, rather than hard-coding a number that would silently drift from the server. With the SDK version now fixed by the image, that constant is deterministic too.

Happy to hard-code both if you would rather have it fully explicit — it just means updating two numbers in lockstep whenever the SDK moves but I do not feel strongly about it.

@juliusvonkohout whats your take on it?

vikas-saxena02: @juliusvonkohout @christian-heusel  Made the image bump on my local copy, and it surfaced a bootstrapping problem worth your view on.

The image ships `kubeflow==0.4.0` but not the `spark` extra, so `pyspark` is absent. That means I cannot read `DEFAULT_SPARK_VERSION` back from the SDK before installing the client, because importing `kubeflow.spark` pulls in `SparkClient`, which imports `pyspark`:

```console
+ python -m pip install --quiet kubeflow-spark-api
++ python -c 'from kubeflow.spark.backends.kubernetes import constants; print(constants.DEFAULT_SPARK_VERSION)'
  File ".../kubeflow/spark/api/spark_client.py", line 20, in <module>
    from pyspark.sql import SparkSession
ModuleNotFoundError: No module named 'pyspark'
```

So deriving the client version needs either an unpinned install first and a re-pin afterwards — three pip invocations — or loading the constants module by file path, which reaches two levels into SDK internals.

Both feel like more machinery than the problem deserves, so I would rather just take your original suggestion and pin it:

```bash
SPARK_VERSION="4.0.4"
kubectl -n "${KF_PROFILE}" exec "${WORKSPACE_POD}" -- \
  python -m pip install --quiet "kubeflow-spark-api" "pyspark-connect==${SPARK_VERSION}"
```

The script already asserts the client and server major.minor match once connected, so a drift from the SDK default fails loudly with a clear message rather than as an odd protocol error.

Happy to go that way unless you would prefer one of the other two. Please let me know your thoughts

tests/spark_connect_from_workspace.py
juliusvonkohout: Please keep it simple since this is not a persistent environment and we rather want to reduce the complexity for tests. Fail fast and do not try to handle errors in the tests.

christian-heusel (approved):

tests/workspaces_spark_connect_test.sh
vikas-saxena02: @juliusvonkohout @christian-heusel this review item is still unesolved. can i please get some inputs on my above comment

tests/workspaces_spark_connect_test.sh
```diff
@@ -0,0 +1,46 @@
...
+# so an RBAC regression fails here with a clear message rather than inside Spark.
+kubectl auth can-i create sparkconnects \
+  --as="system:serviceaccount:${KF_PROFILE}:ws-spark-test" \
+  -n "${KF_PROFILE}" | grep -qx yes
+
+# The jupyter-scipy image does not ship the Kubeflow SDK. Installing it here rather
+# than baking a dedicated image keeps this test self-contained. The resolver reports
+# protobuf conflicts against the pre-installed kfp packages; they do not affect Spark.
+kubectl -n "${KF_PROFILE}" exec "${WORKSPACE_POD}" -- \
+  python -m pip install --quiet "kubeflow-spark-api"
```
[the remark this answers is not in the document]
vikas-saxena02: The image does ship the SDK — it is `kubeflow==0.4.0`, and `Name` and `PodTemplateOverride` are both re-exported from `kubeflow.spark`:

```console
$ docker run --rm ghcr.io/kubeflow/kubeflow/notebook-servers/jupyter-scipy:v1.11.0 \
    grep -n "Name\|PodTemplateOverride" /opt/conda/lib/python3.12/site-packages/kubeflow/spark/__init__.py
22:    Name,
24:    PodTemplateOverride,
45:    "Name",
47:    "PodTemplateOverride",
```

You are right that `kubeflow-spark-api` alone was not enough, though — `kubeflow/spark/__init__.py` imports `SparkClient`, which imports `pyspark`, so the version read below failed. Fixed by adding `pyspark-connect` to the same install rather than reinstalling the whole SDK, which keeps the SDK version pinned by the image instead of resolved from PyPI on the day. That was the point of @christian-heusel's image bump.

tests/spark_connect_from_workspace.py
```diff
@@ -0,0 +1,113 @@
...
+        logger.info("Deleting Spark Connect session %s", SESSION_NAME)
+        try:
+            client.delete_session(SESSION_NAME)
+        except RuntimeError:
+            # If the test body already failed, that exception is the useful one and is
+            # re-raised on exit from this block. Only surface a cleanup failure when it
+            # is the only thing that went wrong.
+            if sys.exc_info()[0] is None:
+                raise
+            logger.warning("Could not delete session %s during cleanup", SESSION_NAME)
```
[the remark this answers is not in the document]
vikas-saxena02: unable to commit suggestion from UI.. so pushing it as a separate commit

tests/spark_connect_from_workspace.py
vikas-saxena02: done and dusted!!

tests/workspaces_spark_connect_test.sh
vikas-saxena02: @christian-heusel can you review my code changes for this one and see if you are happy with it.

christian-heusel: I'm happy with it for now, I guess all further improvements should be sorted out via follow ups to the sdk and workspaces repositories 😊

vikas-saxena02: @christian-heusel thanks :)

PR #3592, opened 2026-08-31
tests: reject unknown comparison descriptor fields at load

danish9039:
## What this does

Follow-up to the descriptor harness ([#3577](https://github.com/kubeflow/community-distribution/pull/3577)), closing the five findings of its final Copilot review ([review 5067507448](https://github.com/kubeflow/community-distribution/pull/3577#pullrequestreview-5067507448)). A comparison descriptor decides what is compared, so a misspelled field must fail at load, not be silently replaced by a default.

- **Unknown fields are rejected at three levels**: the descriptor top level, each scenario, and each `ignoredLabels` entry. Previously `helmUsesKustomizeNameHash:` or `excludeKind:` loaded and were ignored.
- **Action values are typed**: `ignorePodTemplateAnnotations` and `compareDataAsYaml` must be non-empty lists of keys. Previously a boolean raised `TypeError` at comparison time and a scalar string was iterated character by character.
- **A declared scenario `values` file must exist in the chart.**
- **`helmOnlyResources` patterns now match with per-segment `*` wildcards**, through the same `resource_matches` used by `knownDifferences`, as `tests/README.md` already documents. Previously the mapping was exact string membership.
- `helmOnlyResources` patterns get the same shape validation (`Kind/name` or `Kind/namespace/name`) as every other pattern family.

## Verification

- `python3 tests/run_helm_kustomize_comparison.py all` — all eleven components, every scenario, green under Helm 4.2.2: every shipping descriptor loads under the stricter rules.
- `python3 tests/comparison_descriptors_test.py` — 24 tests, including new guards for each rejection.
- `python3 tests/helm_kustomize_compare_test.py` (both invocation forms) — includes a new wildcard-matching test.
- `black --check tests/`, `git diff --check` clean.

### DIFF
--- a/tests/run_helm_kustomize_comparison.py
+++ b/tests/run_helm_kustomize_comparison.py
@@ -80,15 +80,33 @@ def _validate_pattern(path, family, pattern):
         )
 
 
+def _reject_unknown_fields(path, context, mapping, allowed):
+    unknown = set(mapping) - allowed
+    if unknown:
+        raise ValueError(
+            f"{path}: unknown {context} fields: {', '.join(sorted(unknown))}"
+        )
+
+
 def _validate_allowances(path, descriptor):
     """Reject malformed declared allowances at load, not at comparison time."""
     for entry in descriptor.get("ignoredLabels") or []:
         _validate_reason(path, "ignoredLabels", entry)
+        _reject_unknown_fields(
+            path, "ignoredLabels", entry, {"reason", "keys", "podTemplates"}
+        )
         keys = entry.get("keys")
-        if not isinstance(keys, list) or not keys:
+        if (
+            not isinstance(keys, list)
+            or not keys
+            or not all(isinstance(key, str) and key for key in keys)
+        ):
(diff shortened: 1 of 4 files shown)


tests/run_helm_kustomize_comparison.py
```diff
@@ -157,6 +203,25 @@ def load_descriptor(path):
...
+        )
+        if "values" in scenario:
+            values = scenario["values"]
+            chart_directory = path.parent.parent.resolve()
+            if not isinstance(values, str) or not values:
+                raise ValueError(
+                    f"{path}: scenario {name!r} values must be a non-empty path in the chart"
+                )
+            values_path = (chart_directory / values).resolve()
+            if not values_path.is_relative_to(chart_directory) or not values_path.is_file():
```
juliusvonkohout:
```suggestion
            if (
                not values_path.is_relative_to(chart_directory)
                or not values_path.is_file()
            ):
```

PR #3613, opened 2026-09-21
helm: support generated charts without custom resource definitions

danish9039:
Components such as Models UI and Hub have no custom resource definitions, but the shared generator currently rejects their empty CRD output. Add an explicit `crds_payload_filename=None` configuration that emits only the resource payload and fails if the source unexpectedly introduces a CRD. Existing components retain the strict default.

Validation: 28 shared-generator tests and 13 freshness tests pass; focused pre-commit checks pass. Tests cover resource-only generation, unexpected definitions and an empty resource payload.


danish9039:
@juliusvonkohout for review , very small change needed for charts

danish9039:
Implemented through the Models UI stack merged in https://github.com/kubeflow/community-distribution/pull/3614 (https://github.com/kubeflow/community-distribution/commit/541f9a6ed4711e3fd2d7171cb598a173680cf077). Closing this duplicate proposal; remaining review corrections will use focused follow-up pull requests.

PR #3615, opened 2026-09-21
helm: add generated Knative Serving chart

danish9039:
Add a co-located Knative Serving chart generated from the distribution's patched Serving/net-istio v1.23.0 baseline. One release establishes retained CRDs, then controllers/webhooks, then the two self-API consumers. It owns a retained component Namespace and stores release history in `kubeflow`; existing Istio gateway ownership stays with the Istio release.

This pull request carries [aggregated-role prerequisite #3609](https://github.com/kubeflow/community-distribution/pull/3609). Component-only review range: `0d7addecb0c034ffdc482c7ec11d5d71187fe422..033bc057e8deab75db428b28737c4b22d161f6b5`. Rebase after that prerequisite merges.

The initial integration run exposed a separate admission-webhook ownership conflict. Two checked chart-side patches now omit only the rules that Knative's webhook controllers replace. The distribution overlay stays unchanged. The comparison requires Helm to omit those exact fields and compares everything else; raw baseline regression covers all 81 objects. Installation proves admission defaulting and rejection with restricted-Pod-Security-compatible server-side dry runs before creating consumers. The lifecycle flow repeats this check after upgrades and rollback. The README documents bootstrap maintenance and pre-fix revision recovery.

Validation: eight chart/installer tests, 31 comparator tests, 33 descriptor tests, partition/freshness checks, parity, release storage size, lint and formatting pass. Full Helm integration includes routed-response/JWT checks, upgrades, rollback and retained-object recovery. All 29 checks passed at `033bc057`, including [the full Helm integration run](https://github.com/kubeflow/community-distribution/actions/runs/35627265263/job/106424576393). Cross-version schema downgrade is not claimed. Migration Jobs remain explicit administrative steps.


danish9039:
@juliusvonkohout for review. This depends on https://github.com/kubeflow/community-distribution/pull/3609.

